PR #2, opened 2017-12-11
Update README.md

Jon-Galloway:


Chalarangelo:
Thanks for this, I didn't notice it. Please make the changes to the respective snippet files (along with their filenames, so that the builder script can pick them up and fix them). We keep snippets in seperate files for easier maintenance.

PR #5, opened 2017-12-11
simple ES6 unique array  values

prodigic:


mbuck:
If you're going ES6, the spread operator with a Set is even easier.
`const unique = [...new Set(arr)];`

PR #7, opened 2017-12-12
Update unique-values-of-array.md

prodigic:


prodigic:
I think the two choices are 
1. `unique` values, i.e. values that only occur once in the list,  
2. `distinct` items, the list with duplicate values remove

Chalarangelo:
The naming was conflicting, so I already dealt with it, but I will change it as your suggestion is better than the naming right now (you can PR it if you like, too).

PR #9, opened 2017-12-12
Add worthy alternative version of `rgbToHex()` function

kutsan:
I think it's worth adding it since once you understand it, it explains to you the whole concept.


slavabez:
That's a really clever one!

zo:
Very nice indeed, and probably faster, but you still need to use padStart at the end to avoid incorrect hex value like

```js
rgbToHex(0,0,0) // => "0"
rgbToHex(0,0,255) // => "ff"
```

kutsan:
@Chalarangelo Well, the problem is if you know how bitwise left-shift operator works, then you already know how this one works. I guess explaining the bitwise operator itself cannot be happen in one line, but I will try what I can do. For the other topic, I'll do it once I get home (I am on mobile now.). Thank you too!

@zo Of course, those snippets are in very simple form of themselves. For the edge cases like you mentioned you can simply extend it. Besides `0`, you can give anything as parameters and result won't be the color you want.

zo:
I disagree that's edge case scenarios, `rgbToHex(0, 127, 255)` is a valid color.
And I agree that you don't have to say anything more than mentioning "bitwise left-shift operator" in the description.

Also I think this version should be the "curated" version, not an alternate version.
```js
const rgbToHex = (r, g, b) =>
  ((r << 16) + (g << 8) + b).toString(16).padStart(6, "0");
```
It's smaller, faster and don't create a "soon to be garbage-collected" Array.

Chalarangelo:
Great points @zo. It's a more efficient implementation, so I think it should be the only version we keep anyways. 

Also, it's true that these are not edge-cases so they should be dealt with be default. Finally, a good explanation is necessary.

kutsan:
@zo Yeah, you are right, it's not edge case. I updated the snippet with `padStart()` function.

@Chalarangelo Care to look at my commit again? I tried to explain but I don't it's okay or not. You can update it or I can delete it. Tell me your opinion.

kutsan:
Good luck! Thanks again.

PR #12, opened 2017-12-12
Add reverseString

fejes713:


fejes713:
@Chalarangelo the good old times :octocat:

PR #13, opened 2017-12-12
Directly return object from arrow function

albertorestifo:
I think directly returning an object from the arrow function improves readability.

### DIFF
--- a/snippets/object-from-key-value-pairs.md
+++ b/snippets/object-from-key-value-pairs.md
@@ -1,8 +1,7 @@
 ### Object from key-value pairs
 
-Use `map()` to create objects for each key-value pair, combine with `Object.assign()`.
+Use `Array.reduce()` to create and combine key-value pairs.
 
 ```js
-var objectFromPairs = arr =>
-  Object.assign(...arr.map( v => ({ [v[0]] : v[1] })));
+const objectFromPairs = arr => arr => arr.reduce((a,b) => { a[b[0]] = b[1]; return a;}, {})
 ```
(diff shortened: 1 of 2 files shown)


darrenscerri:
Can you rebase from master and rebuild the README.md due to #20 ?

snippets/object-from-key-value-pairs.md
```diff
@@ -4,5 +4,5 @@ Use `map()` to create objects for each key-value pair, combine with `Object.assi
 
 ```js
 var objectFromPairs = arr =>
-  Object.assign(...arr.map( v => {return {[v[0]] : v[1]};} ));
+  Object.assign(...arr.map( v => ({ [v[0]] : v[1] })));
```
Chalarangelo: This is indeed shorter, but sacrifices readability for shorter code. I am not entirely sold on the idea. Any other opinions?

darrenscerri: How about: 

`var objectFromPairs = arr => arr.reduce((a,b) => (a[b[0]] = b[1], a), {})`

or 

`var objectFromPairs = arr => arr.reduce((a,b) => { a[b[0]] = b[1]; return a;}, {})`

Chalarangelo: The first variant is quite easy to read. Is it any faster performance-wise than the original one? (just curious) I would definitely prefer the `.reduce()` variant, though in terms of readability as even the original one I had implemented felt wrong to me somehow.

darrenscerri: Although the ES6 version is an interesting implementation, I feel that the `.reduce()` version is more readable. I would assume that the `.reduce()` version would perform marginally better since it doesn't create intermediary arrays and objects.

darrenscerri:
```

Chalarangelo: Right, so we'll stick with `reduce()`.

PR #14, opened 2017-12-12
Created confirm ending of the string and reverse string

meetzaveri:
New JS snippet for confirming ending of the string


Chalarangelo:
String reversal has already been added in #12 and checking for ending can be done with [`String.prototype.endsWith()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith), so there's no need to write a method for that, as it's a pretty easy name to read and remember.

PR #20, opened 2017-12-12
Sort snippets with case-insensitivity

darrenscerri:
Running `scripts/builder.js` on different file systems may result in different ordering of the snippets in `README.md` depending on what order the OS returns the array of files in `fs.readFileSync`. Using a custom case-insensitive comparator will ensure consistency when building on different environments.

### DIFF
--- a/scripts/builder.js
+++ b/scripts/builder.js
@@ -7,7 +7,19 @@ var staticPartsPath = './static-parts';
 var snippets = {}, startPart = '', endPart = '', output = '';
 
 try {
-  for(var snippet of fs.readdirSync(snippetsPath)){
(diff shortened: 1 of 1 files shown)


darrenscerri:
Yes, was building the latest commit and snippets starting with an upper-case letter were hoisted up to the top of the list even though I didn't do any other changes. I'm using a Mac, and HFS is case-insensitive so not sure why this is happening, but apparently the results are being ordered differently than in other environments. Anyway, this should solve the issue, even if other environments return the results in a different order :)

Chalarangelo:
I am so glad that someone tested this on a Mac, because I never could. Thanks a ton for this, it will prevent a lot of problems!

PR #21, opened 2017-12-12
Faster even/odd check with bitwise operator

overflowsith:
No need to use Math.abs() to extend logic to negative numbers and after that check using the modulo (%) operator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,11 +137,10 @@ var escapeRegExp = s =>
 
 ### Even or odd number
 
-Use `Math.abs()` to extend logic to negative numbers, check using the modulo (`%`) operator.
 Return `true` if the number is even, `false` if the number is odd.
 
 ```js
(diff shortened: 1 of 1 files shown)


overflowsith:
Sorry, I will create a new one following the contribution guide :)

README.md
```diff
@@ -137,11 +137,10 @@ var escapeRegExp = s =>
 
 ### Even or odd number
 
-Use `Math.abs()` to extend logic to negative numbers, check using the modulo (`%`) operator.
```
Chalarangelo: 30 seconds of code is primarily a learning resource. Kindly explain how this works in one line so that people understand. I like the idea though. Also, you should apply changes to the individual snippet file, not README.md, as the latter is auto-generated by a script and changes in it will be overwritten (yes, contribution guidelines were not clear, I will update them).

overflowsith: Working on it

PR #22, opened 2017-12-12
add another way to calculate factorial

karamari:
As I commented on [reddit](https://www.reddit.com/r/javascript/comments/7j37iu/i_have_been_collecting_useful_javascript_code/dr4narm/), here's my pull request. I left the original one, was that OK?

English is not my first language so something might be wrong.


karamari:
No problem!

PR #26, opened 2017-12-12
using bitwise operator for checking if the number is even or odd

overflowsith:
I (hopefully) followed the contribution guide and add a simple explanation

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,11 +137,22 @@ var escapeRegExp = s =>
 
 ### Even or odd number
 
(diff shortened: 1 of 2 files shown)


darrenscerri:
Only works for integers: `isEven(2.2) === true` since `2.2 & 1 === 0`.

Using modulus to check if a number is even/odd is quite common practice.

Chalarangelo:
Good point, @darrenscerri. Guess we're sticking with the modulo version then, as this one is error-prone. Thanks for pointing that out! 👍

PR #28, opened 2017-12-12
Correct a typo in "slice".

shlomif:


shlomif:
On Tue, 12 Dec 2017 03:01:17 -0800
Angelos Chalaris <notifications@github.com> wrote:

> Chalarangelo approved this pull request.
> 
> Good catch! Thanks for the heads up!
>

you are welcome and thanks for merging it.
 
> 
> 



-- 
-----------------------------------------------------------------
Shlomi Fish       http://www.shlomifish.org/
List of Portability Libraries - http://shlom.in/port-libs

Only dead fish go with the flow.
    — via Nadav Har’El

Please reply to list if it's a mailing list post - http://shlom.in/reply .

PR #29, opened 2017-12-12
Add getScrollPos of HTMLElement / window

renatodeleao:
Get the current distance scrolled by `window` or `HTMLElement` as an {x,y} object.
I use this often for simple scroll based interactions/triggers
 
[Codepen demo](https://codepen.io/ohsimtabem/pen/MrWpmd)


renatodeleao:
You welcome sir! 👍

PR #30, opened 2017-12-12
Change all snippets from var to const

jorgegonzalez:
Changed all snippets that began with `var` to `const`. Didn't update the README, however, as I'm not sure if you wanted that as well.


Chalarangelo:
Thank you so much for this! README will be rebuilt as soon as I run the script, don't worry about it. 👍

PR #31, opened 2017-12-12
Anagrams recursion refactor (code legibility)

nikfrank:
originally .map was being used as a .forEach

now ternary operators are used to mimic pattern matching style

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,17 @@ Use `map()` to combine the letter with each partial anagram, then `reduce()` to 
 Base cases are for string `length` equal to `2` or `1`.
 
 ```js
-var anagrams = s => {
-  if(s.length <= 2)  return s.length === 2 ? [s, s[1] + s[0]] : [s];
-  return s.split('').reduce( (a,l,i) => {
-    anagrams(s.slice(0,i) + s.slice(i+1)).map( v => a.push(l+v) );
-    return a;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,13 +56,17 @@ Use `map()` to combine the letter with each partial anagram, then `reduce()` to
...
 
 ```js
-var anagrams = s => {
-  if(s.length <= 2)  return s.length === 2 ? [s, s[1] + s[0]] : [s];
-  return s.split('').reduce( (a,l,i) => {
-    anagrams(s.slice(0,i) + s.slice(i+1)).map( v => a.push(l+v) );
-    return a;
-  }, []);
-}
+export const anagramsOf = s => (
```
Chalarangelo: Remove `export`.

README.md
```diff
@@ -56,13 +56,17 @@ Use `map()` to combine the letter with each partial anagram, then `reduce()` to
...
+  (s.length < 2) ? [s] :
```
Chalarangelo: This whole ternary in ternary operator is far less legible than before. I do not disagree that the previous syntax might look complicated, but this is by no means any better.

Chalarangelo:
This snippet has been updated already. If you feel like you can further improve it, please submit a new PR. Thank you for helping out! 👍

PR #33, opened 2017-12-12
docs: use filter replace reduce

WittBulter:
`filter` is more concise than `reduce`, and have higher performance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,11 +95,12 @@ const capitalize = (str, lowerRest = false) =>
 
 ### Count occurrences of a value in array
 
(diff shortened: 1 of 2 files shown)


Chalarangelo:
I had to test both snippets to figure out which on is best performance-wise and [here](https://jsperf.com/count-occurences) you can see the results. `Array.reduce()` is a lot faster, so we're sticking with this. After all, both are easy to read and performance matters quite a lot. Thanks for putting in the time to help, though! 👍

WittBulter:
great! thank you for pointing out my mistake！ 👍  👍

PR #34, opened 2017-12-12
Avoid creating global vars for `f` and `v`.

fictorial:


Chalarangelo:
Huge oversight on my part, thanks for the heads up!

While we're at it, maybe we should update the code to look something like this:

```js
var getUrlParameters = url =>
   url.match(/([^?=&]+)(=([^&]*))?/g).reduce((a,v) => (a[v.split('=')[0]] = v.split('=')[1], a), {});
```

`Array.reduce()` is cleaner and makes more sense for this, I suppose. There was another snippet that was updated in a similar fashion earlier today.

fictorial:
Actually I like the previous version because it might introduce a lot of developers to computed property names.

Chalarangelo:
I agree and 30 seconds of code is mainly a learning resource, but performance-wise (test results [here](https://jsperf.com/test-get-url-parameters)) and in terms of readability `Array.reduce()` seems to be superior. Any other opinions from the community?

skatcat31:
Double splits get me all sorts of bothered. On long enough strings the performance hit could be huge and since there is no accounting for input, avoid duping. Sure the optimizer should pick up it can do it only once, but it might be best to just separate clauses with a semi for clarity and performance, and since it's simple enough it could still just sit on one line(although it looks noisy, which for styling sake could be a deal breaker). The real issue is that if you only use a few time,s the optimizer won't pick it up meaning it's just a performance hit for no reason.

Of course it could have changed since I last checked, so I may just be being paranoid here, but incase that's still the case(Chrome63):

```js
var getUrlParameters = url =>
   url.match(/([^?=&]+)(=([^&]*))?/g).reduce((a,v) => (let [key,value] = v.split('='); a[key] = value; return a), {});
```

PS: I'm having issues adding it to the test cases... If you're willing to do it before I have a chance to try to add it correctly, please feel free to do so

Chalarangelo:
@skatcat31 Good point, the snippet I proposed was slow indeed. But it can be optimized without setting variables, using only `Array.slice()` and some indexing:

```js
var getUrlParameters = url =>
  url.match(/([^?=&]+)(=([^&]*))?/g).reduce(
    (a,v) => (a[v.slice(0,v.indexOf('='))] = v.slice(v.indexOf('=')), a), {}
  );
```

[Performance results](https://jsperf.com/get-url-parameters-2)

skatcat31:
Still can't add in my code to the test case... damned JSPerf... Never could figure out that site.

Only problem I can see with your latest snippet is that it will call indexof twice, causing a loop of O(n) * 2 at worse case(not found) so it could probably be sped up by using an intermediate value. The comma operator was a nifty touch though

PR #37, opened 2017-12-12
Added async function chain

RobertAKARobin:
I included a "long form" as well. I think providing an explanation in this way may be very helpful to beginners, for whom this is a great resource. GitHub's `<details>` functionality is a little-known secret.

### DIFF
--- a/snippets/chain-async-functions.md
+++ b/snippets/chain-async-functions.md
@@ -3,13 +3,12 @@
 Loop through an array of functions containing asynchronous events, calling `next` when each asynchronous event has completed.
 
 ```js
-const chainAsync = fns => {
-	let curr = 0; const next = () => fns[curr++](next); next()
-}
+const chainAsync = fns => { let curr = 0; const next = () => fns[curr++](next); next(); }
+/* 
 chainAsync([
-	next => { console.log('This happens at 0 seconds'); setTimeout(next, 1000) },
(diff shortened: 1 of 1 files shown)


snippets/chain-async-functions.md
```diff
@@ -0,0 +1,40 @@
+### Chain asynchronous functions
+
+Loop through an array of functions containing asynchronous events, calling `next` when each asynchronous event has completed.
+
+```js
+const chainAsync = fns => {
+	let curr = 0; const next = () => fns[curr++](next); next()
+}
```
Chalarangelo: Provide an example in a comment, below the declaration, please.

snippets/chain-async-functions.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+<details>
```
Chalarangelo: Indeed `<details>` is a pretty cool way to explain, but it doesn't play on any MS browser, so this is probably a bit of a problem. Maybe linking to an explanation or article would be preferrable.

snippets/chain-async-functions.md
RobertAKARobin: @Chalarangelo Done.

snippets/chain-async-functions.md
RobertAKARobin: I wasn't aware of the MS limitation. Shouldn't be surprised. Thanks.

PR #38, opened 2017-12-12
Add chunk array

kingdavidmartins:
Creates an array of elements split into groups the length of size. 
If array can't be split evenly, the final chunk will be the remaining elements.

### DIFF
--- a/snippets/chunk-array.md
+++ b/snippets/chunk-array.md
@@ -1,13 +1,12 @@
 ### Chunk Array
 
-Creates an array of elements split into groups the length of size. 
-If array can't be split evenly, the final chunk will be the remaining elements.
+Use `Array.apply()` to create a new array, that fits the number of chunks that will be produced.
+Use `Array.map()` to map each element of the new array to a chunk the length of `size`.
+If the original array can't be split evenly, the final chunk will contain the remaining elements.
 
 ```js
 const chunk = (arr, size) => 
-  Array
-    .apply(null, {length: Math.ceil(arr.length/size) })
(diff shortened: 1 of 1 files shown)


snippets/chunk-array.md
```diff
@@ -0,0 +1,14 @@
+### Chunk Array
+
+Creates an array of elements split into groups the length of size. 
+If array can't be split evenly, the final chunk will be the remaining elements.
```
Chalarangelo: The description should describe the *how* of the implementation (i.e. how the code does what it does), not the *what* (i.e. what you aim to accomplish with this function). Please update your snippet.

snippets/chunk-array.md
```diff
@@ -0,0 +1,14 @@
...
+
+```js
+const chunk = (arr, size) => 
+  Array
+    .apply(null, {length: Math.ceil(arr.length/size) })
+    .map((value, index) => arr.slice(index*size, index*size+size) )
+
+// const myArray = [2, 2, 2, 2, 2, 2, 3, 2, 3, 2, 3, 2, 2];
+// chunk(myArray, 3) -> [ [ 2, 2, 2 ], [ 2, 2, 2 ], [ 3, 2, 3 ], [ 2, 3, 2 ], [ 2 ] ]
```
Chalarangelo: The example can be reduced to one line and shortened (fewer elements, chunks of `size = 2` should be fine).

kingdavidmartins: Thanks 😄! Sorry for the late reply, went out with the family for the holidays. I will be sure to follow the recommended format with my next pull request 👍

PR #40, opened 2017-12-12
minor refactor to the fibonacci array generator

elderhsouza:
The fibonacci sequence being zero indexed as well as the arrays, I think we could just initialize an array with the passed length, fill it with zeroes and use the indexes for the first positions.


elderhsouza:
Sure! No problem, I got confused about the style guide, but then I saw the open issue about it.
Great project!

PR #42, opened 2017-12-12
get minimum value from an array

elderhsouza:
I think it shows a good use of the spread operator in variadic functions

### DIFF
--- a/snippets/get-min-value-from-array.md
+++ b/snippets/get-min-value-from-array.md
@@ -4,5 +4,5 @@ Passing an array, it executes the Math.min method using the spread operator to f
 
 ```js
 const getMinValue = arr => Math.min(...arr);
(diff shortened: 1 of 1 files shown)


snippets/get-min-value-from-array.md
```diff
@@ -0,0 +1,8 @@
+### Get min value from array
+
+Passing an array, it executes the Math.min method using the spread operator to fill its variadic arguments
+
+```js
+const getMinValue = arr => Math.min(...arr);
+// getMinValue([10, 1, 5) -> 1
```
darrenscerri: missin square bracket

elderhsouza:
@Chalarangelo @darrenscerri 
Thanks for the review notes.
If this is good we should definitely add a version to get the max value as well.

PR #44, opened 2017-12-12
Refactor measure-time-taken-by-function from spread syntax to callback function

cengels:
I thought the spread syntax was kinda odd, so I opened this pull request to adopt the more common callback style. Discussion welcome.

One potential benefit of this is that the callback can be stored in a separate variable, allowing you to pass the variable to the function timeTaken() directly. You can also pass parameterless functions directly without wrapping them in another arrow function.

### DIFF
--- a/snippets/measure-time-taken-by-function.md
+++ b/snippets/measure-time-taken-by-function.md
@@ -4,10 +4,10 @@ Use `performance.now()` to get start and end time for the function, `console.log
 First argument is the function name, subsequent arguments are passed to the function.
 
 ```js
-const timeTaken = (func,...args) => {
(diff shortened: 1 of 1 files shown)


atomiks:
- Function expressions need a semicolon at the end of the block
- I think `ms` should be added to the end of the time taken, and maybe optionally use `.toFixed(2)` to round it to 2 decimal places?

Chalarangelo:
Is there any benefit to one or the other? I mean using the callback is the most common way, but is there any value to keeping the other method too? Callbacks can cover a wider use-case I suppose. Anyone against switching to the callback version? 

I'll merge in a couple of hours if nobody is against, sorry for the delay and thanks a lot for contributing to the project!

PR #45, opened 2017-12-12
Added getType function

macsmac:
Snippet to get native type of function (lowered-case constructor name)

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,15 @@ const arrayMin = arr => Math.min(...arr);
 // arrayMin([10, 1, 5]) -> 1
 ```
 
+### Get native type of value
+
+Returns lower-cased constructor name of value, "undefined" or "null" if value is undefined or null
+
+```js
+const getType = v => v === undefined ? "undefined" : v === null ? "null" : v.constructor.name.toLowerCase();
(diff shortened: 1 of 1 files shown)


darrenscerri:
Finding it hard to find a use-case for this function. In what cases would this be useful?

macsmac:
@darrenscerri

using this on my game server

```js
switch(getType(postres)) {
case "entity": /* ... */; break;
case "broadcast": /* ... */; break;
case "dataset": /* ... */;
}
```

skatcat31:
I use an extremely similar function in a production environment as well since typeof is unpredictable in certain cases

darrenscerri:
Interesting, that's a reasonable use-case.

macsmac:
`instanceof` vs `getType`

```js
'use strict';

const getType = v => v === undefined ? "undefined" : v === null ? "null" : v.constructor.name.toLowerCase();

class A { constructor() { this.a = 1; } }
class B extends A { constructor() { super(); this.b = 2; } }

const b = new B();

console.log(b instanceof B); // true
console.log(b instanceof A); // also true

console.log(getType(b) === "b"); // true
console.log(getType(b) === "a"); // obviously false
```

PR #47, opened 2017-12-12
Make "Anagrams of string" fully functional

jussisaurio:
It is good practice to keep map/filter/reduce side-effect free. This PR removes the impure `.push` inside `.map` and instead returns a new array concatenated from the accumulator and the recursive `anagrams` call.


jussisaurio:
https://repl.it/repls/LavenderImpoliteEastrussiancoursinghounds

PR #51, opened 2017-12-12
refactor curry to account for variadic functions

elderhsouza:
altough certainly more verbose, this version accounts for variadic functions such as Math.min that can't represent it's arity with the length property properly. for such cases you can pass the optional argument arity (as in the edited example).
what do you think?

### DIFF
--- a/snippets/curry.md
+++ b/snippets/curry.md
@@ -3,6 +3,7 @@
 Use recursion.
 If the number of provided arguments (`args`) is sufficient, call the passed function `f`.
 Otherwise return a curried function `f` that expects the rest of the arguments.
+If you want to curry a function that accepts a variable number of arguments (variadic function) like Math.min for example, you can optionally pass the number of arguments to the second parameter arity.
 
 ```js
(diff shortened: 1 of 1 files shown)


Chalarangelo:
Good thinking! I like this, but the description should be updated slightly to explain what it does more accurately than it did before. This is really robust, great job!

elderhsouza:
Thank you! I will edit the description to explain the optional parameter
then.

Em ter, 12 de dez de 2017 20:39, Angelos Chalaris <notifications@github.com>
escreveu:

snippets/curry.md
```diff
@@ -5,8 +5,15 @@ If the number of provided arguments (`args`) is sufficient, call the passed func
 Otherwise return a curried function `f` that expects the rest of the arguments.
```
Chalarangelo: Still waiting on the updated description, apart from that the function is solid. Thanks for contributing yet again! 👍

elderhsouza:
I just comitted an edited version.

PR #53, opened 2017-12-12
Create median-of-array-of-numbers.md

atomiks:


Chalarangelo:
Thanks a lot for the contribution!

I was trying my hard at shortening the code and ended up with this ugly thing:

```js
const median = arr => 
  arr.sort((a,b) => a - b).slice(
    arr.length%2?Math.floor(arr.length / 2):Math.floor(arr.length / 2)-1,
    arr.length%2?Math.floor(arr.length / 2)+1:Math.floor(arr.length / 2)+1)
  .reduce((a,v,_,r) => a+v/r.length,0);
``

I'll probably just keep your version pretty much as-is and shorten it a little bit, as mine is mostly unreadable. Review, changes and merge in the morning (about 8.5 hours from now). Cheers!

atomiks:
Yeah I couldn't think of a way to shorten it. Sometimes the extra verboseness is superior to a shorter, obscure version. (And is also better for performance)

Chalarangelo:
I'm naturally suspsicious of bitwise operators and this is definitely one case where we shouldn't. `Math.floor()` is clean and readable. I will just make some minor changes to the snippet in the morning to make it like 4 lines long or something without sacrificing anything (basically moving both variable declarations to one line, using `let` and shortening the ternary operator to one line, too). Apart from that it should be good to go. 👍

atomiks:
I'm a fan of the "readable" style which spans more lines, but if the shortness is your goal here then go for it 👍 

However, why is there a need for `let` if nothing is reassigned?

Chalarangelo:
Good point about using `let`. I assumed `const` would apply to the global scope, but that is definitely not true. I see no reason why not use `const` then, plus I should remember to write about this in the styleguide (which I still haven't gotten around to writing).

atomiks:
>I assumed const would apply to the global scope

Even `var` would still be scoped to the function 😄 

As the for the style guide, I recommend readability & conciseness over shortness personally, but that's just me. (Especially if you want developers to understand the code in 30 seconds or less). 

Whatever style guide you come up with, consistency should be the focus

PR #64, opened 2017-12-13
Add promisify snippet

### LINKED ISSUE #57: Promisify snippet
This snippet might fit into the list?

```javascript
const promisify = func =>
  (...args) =>
    new Promise((resolve, reject) =>
      func(...args, (err, result) =>
        err
        ? reject(err)
        : resolve(result))
    )
```

DanielRamosAcosta:
This PR adds the snipped talked in #57. Resolve #57.

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,7 +419,7 @@ const powerset = arr =>
 
 ### Promisify
 
-Creates a promise version of the given callback-style function. In Node 8+, you
+Creates a promisified version of the given callback-style function. In Node 8+, you
 can use [`util.promisify`](https://nodejs.org/api/util.html#util_util_promisify_original)
 
 ```js
--- a/README.md
+++ b/README.md
@@ -432,7 +432,7 @@ const promisify = func =>
         : resolve(result))
     )
 // const stat = promisify(fs.stat)
-// When called, stat returns a promise
+// stat('foo.txt') -> Promise resolves if `foo.txt` exists, otherwise rejects
 ```
 
 ### Random integer in range
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@
 * [Object from key value pairs](#object-from-key-value-pairs)
 * [Pipe](#pipe)
 * [Powerset](#powerset)
+* [Promisify](#promisify)
 * [Random integer in range](#random-integer-in-range)
 * [Random number in range](#random-number-in-range)
 * [Randomize order of array](#randomize-order-of-array)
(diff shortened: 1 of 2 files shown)


snippets/promisify.md
```diff
@@ -0,0 +1,15 @@
+### Promisify
+
+Creates a promisified version of the given callback-style function. In Node 8+, you
```
Chalarangelo: Explain *how* this works, please instead of *what* it does.

snippets/promisify.md
```diff
@@ -0,0 +1,15 @@
...
+
+```js
+const promisify = func =>
+  (...args) =>
+    new Promise((resolve, reject) =>
+      func(...args, (err, result) =>
+        err ? reject(err) : resolve(result));
+    )
+// const stat = promisify(fs.stat)
+// stat('foo.txt') -> Promise resolves if `foo.txt` exists, otherwise rejects
```
Chalarangelo: Can we add an example that could run in the browser without Node? I'd prefer that, so that people can copy paste and test right away. If not, it's ok.

README.md
```diff
@@ -416,6 +417,24 @@ const powerset = arr =>
 // powerset([1,2]) -> [[], [1], [2], [2,1]]
 ```
```
Chalarangelo: You don't have to make any changes to the README or rebuild, I'll handle that.

snippets/promisify.md
DanielRamosAcosta: Humm I've been thinking a callback-style operation in the browser, but can't think of any. Most of them already returns a promise. Do you know any :thinking: ?

Chalarangelo: A timeout-based one (like a function with a timeout that returns a value) maybe?

darrenscerri: `const delay = (d, cb) => setTimeout(cb, d)`?

DanielRamosAcosta: `setTimeout` has the callback in the first argument of the function, instead of the last... :disappointed:

DanielRamosAcosta: I think setTimeout is a bit ugly for this use case, what do you think @Chalarangelo ?

```js
promisify((d, cb) => setTimeout(cb, d))
```

Chalarangelo: I don't hate it to be honest, but I might be in the minority. @darrenscerri what do you think?

README.md
DanielRamosAcosta: Sorry, how do I revert these changes?

snippets/promisify.md
darrenscerri: I think the idea is to promisify any callback-style function, which do not necessarily need to be native APIs. I guess that's fine and drives the point across.

Chalarangelo: Alright, let's stick with that example then, at least for now. If anybody else has any better ideas later down the line, that's why we have PRs! 😉

snippets/promisify.md
```diff
@@ -8,9 +8,7 @@ const promisify = func =>
   (...args) =>
     new Promise((resolve, reject) =>
       func(...args, (err, result) =>
-        err
-        ? reject(err)
-        : resolve(result))
+        err ? reject(err) : resolve(result));
```
DanielRamosAcosta: I had to remove this semicolon and add it at [L12](https://github.com/Chalarangelo/30-seconds-of-code/pull/64/commits/f6c181e0f4e03d523070448dae937455f0a82134#diff-a86c09b8ca7cbea118d447f7e3cf705dR12).

README.md
Chalarangelo: It's alright, I'll take it from here, don't worry about them.

DanielRamosAcosta:
You welcome @Chalarangelo ! :tada:

PR #65, opened 2017-12-13
Create percentile.md

atomiks:
https://www.easycalculation.com/statistics/percentile-rank.php

Feel free to try and one-linerify it if possible, lol

### DIFF
--- a/snippets/percentile.md
+++ b/snippets/percentile.md
@@ -1,0 +1,18 @@
+### Percentile
+
+Calculate how many numbers are below the value and how many are the same value and
+apply the percentile formula.
+
+```js
+const percentile = (arr, val) => {
+  let below = 0, same = 0;
+  
+  for (const number of arr) {
+    if (number < val) below++;
+    if (number === val) same++;
+  }
+  
+  return 100 * (below + (0.5 * same)) / arr.length;
+};
+// percentile([1,2,3,4,5,6,7,8,9,10], 6) -> 55
+ ```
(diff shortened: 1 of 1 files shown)


Chalarangelo:
Not the best in terms of readability, but it's a start for shortening it.

```js
const percentile = (arr, val) => {
  const [b,s] = arr.reduce((acc,v) => [acc[0], acc[1]] = [acc[0] + (v < val ? 1 : 0), acc[1] + (v === val ? 1 : 0)], [0,0]);
  return 100 * (b + 0.5 * s) / arr.length;
}
```

Chalarangelo:
I am dumb, `Array.filter()` is the way to go:

```js
const percentile = (arr, val) => {
  const [below, same] = [arr.filter(v => v < val).length, arr.filter(v === val).length];
  return 100 * (below + 0.5 * same) / arr.length;
}
```

I could even add the filtered lengths right into the final return formula, so that it's a one-liner. Opinions?

atomiks:
Yup, that's it!:

```js
const percentile = (arr, val) =>
  100 * (arr.filter(v => v < val).length + 0.5 * arr.filter(v => v === val).length) / arr.length;
```

Chalarangelo:
I like it better, as it showcases a pretty good technique and is shorter, so I can comprehend what it does more easily.  Update your snippet file and I'll merge the filter-based version. Running a few tests shows a hit in performance but only for relatively large arrays, so I think it'll be ok. It's not a computationally intense operation anyways.

Chalarangelo:
Uhm, this works, too and is shorter. 

```js
const percentile = (arr, val) => 
  100 * arr.reduce((acc,v) => acc + (v < val ? 1 : 0) + (v === val ? 0.5 : 0), 0) / arr.length;
```

atomiks:
Damn, that's good. I'll use that

PR #68, opened 2017-12-13
Create standard-deviation.md

atomiks:
http://www.calculator.net/standard-deviation-calculator.html

As a one-liner it's really long, feel free to optimize the formatting here (or shorten it further somehow).

### DIFF
--- a/snippets/standard-deviation.md
+++ b/snippets/standard-deviation.md
@@ -13,4 +13,5 @@ const standardDeviation = (arr, val) =>
        .reduce((acc, val) => acc + val, 0) 
     / arr.length
   );
+// standardDeviation([10,2,38,23,38,23,21]) -> 12.298996142875
 ```
--- a/snippets/standard-deviation.md
+++ b/snippets/standard-deviation.md
@@ -1,0 +1,16 @@
+### Standard deviation
+
+Use `Array.reduce()` to calculate the mean of the values, the variance of the values, and the sum of the variance
+of the values to determine the standard deviation of an array of numbers.
+
+NOTE: This is **population standard deviation**. Use `/ (arr.length - 1)` at the end to 
+calculate **sample standard deviation**.
+
+```js
+const standardDeviation = (arr, val) =>
+  Math.sqrt(
+    arr.reduce((acc, val) => acc.concat(Math.pow(val - arr.reduce((acc, val) => acc + val, 0) / arr.length, 2)), [])
+       .reduce((acc, val) => acc + val, 0) 
+    / arr.length
+  );
+```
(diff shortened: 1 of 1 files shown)


Chalarangelo:
I like this. I'll see if i can figure out a way to shorten it a bit, but it doesn't look bad as-is. Thanks for the contribution!

atomiks:
Noticed there was an extra parameter for no reason... removed it!

atomiks:
Hmm, since there are two types of standard deviation (`population` and `sample`), maybe there can be a flag as a second parameter, like so?

```js
const standardDeviation = (arr, useSample) =>
  Math.sqrt(
    arr.reduce((acc, val) => acc.concat(Math.pow(val - arr.reduce((acc, val) => acc + val, 0) / arr.length, 2)), [])
       .reduce((acc, val) => acc + val, 0) 
    / (arr.length - useSample ? 1 : 0)
  );
```

Chalarangelo:
I'd like that, maybe we can set the `useSample` to default to one of the two? (your choice which one)

atomiks:
I'm not sure which one is more common/useful 🤔 , not a statistician hah! Maybe I'll look to see.

atomiks:
http://smallbusiness.chron.com/difference-between-sample-population-standard-deviation-22639.html

https://stats.stackexchange.com/questions/269/what-is-the-difference-between-a-population-and-a-sample

It looks like they both have their use cases, but I think `sample` might be more common?

>The population is the set of entities under study. For example, the mean height of men. This is a hypothetical population because it includes all men that have lived, are alive and will live in the future. I like this example because it drives home the point that we, as analysts, choose the population that we wish to study. Typically it is impossible to survey/measure the entire population because not all members are observable 

>**Instead, we could take a subset of this population called a sample and use this sample to draw inferences about the population under study, given some conditions. Thus we could measure the mean height of men in a sample of the population which we call a statistic and use this to draw inferences about the parameter of interest in the population.** It is an inference because there will be some uncertainty and inaccuracy involved in drawing conclusions about the population based upon a sample. This should be obvious - we have fewer members in our sample than our population therefore we have lost some information.

The `N - 1` is meant to be a corrective measure for samples, which are probably more common, so you can switch it to `usePopulation ? 0 : 1` instead. No need for a default value.

atomiks:
Had to de-one-linerify it because the mean was constantly being recalculated, which is bad for perf obviously 😄 . Updated it to cache it in the const

Chalarangelo:
Hey @atomiks, sorry this has taken so long. Add a default parameter to the second argument and we're good to go. I'll see if we can shorten the description a little bit, as this feels a bit too long to me.

atomiks:
> Add a default parameter to the second argument and we're good to go

It's falsy by default though, so there's no need to?

Chalarangelo:
Better be explicit about it @atomiks. 😉

PR #71, opened 2017-12-13
Get Ordinal Suffix of Number

Emgo-Dev:
JS Code Snippet to get the ordinal suffix of a given number (int or string).
Returns the provided value wtih concatenated ordinal.
I made it as small as I thought I could to make it easier to read and simply use includes to find pattern matches.


Emgo-Dev:
@Kyokou Thanks for catching that! It evaded my attention.

Issue #74, opened 2017-12-13
URL params


jz-nbme:
doesn't appear to produce the results you intended:
https://jsfiddle.net/8tdu8q5p/

Pl4gue:
I have created a new snippet that actually works.

Chalarangelo:
The previous snippet's function was missing a `+1` somewhere. My bad. It should be fine now.

PR #77, opened 2017-12-13
 Create URL-parameters-2.md

Pl4gue:
A new snippet for getting all parameters ("key":"value") in an URL due to the old snippet not working.
Ref: #74 
Here is the working proof:
https://jsfiddle.net/pl4gue/0g5vfjfz/5/


Chalarangelo:
Please submit improvements and changes made to the original script. There should be no duplicates, unless they cover different use-cases. The current snippet is buggy, true, so we should figure out how to fix it.

Issue #78, opened 2017-12-13
Snippet tests


kingdavidmartins:
I wanted to know if there's a specific reason for not using TDD. I believe a light weight testing library like tape can provide a lot of value when trying to test code snippets. Especially snippets that are unresponsive or break when they come across edge cases

I believe having tests automated can also help the community test each snippet more rigorously and consistently before and after each PR

Chalarangelo:
Please join us on [gitter](https://gitter.im/30-seconds-of-code/Lobby) to discuss this further. I'll be posting updates to this issue as we go. 😉

Chalarangelo:
As discussed yesterday, we are currently looking for an online tool where we can post our tests for individual snippets and share them (think jsPerf but for tests). If anyone knows of such a thing, kindly point us towards that direction.

Automated testing, splitting snippets etc. is not going to be added until we figure out a way to have it implemented without raising the barrier of entry for new contributors.

skatcat31:
Actually that might be a really cool system to pull in new contributors! What about a learning collection of .md with a bunch of tests that people can code against for fun to satisfy the tests and learn how testing libraries work? Like `30 Seconds to Tests` or something. Then poeple could come here to learn helpful snippets, and how tests work.

Now that is a large and hard project, but it also gives us an excuse to test older snippets and not new ones just yet :wink:

mycaule:
There are also dynamic documentation generators.
In Scala we have: https://github.com/tpolecat/tut
I don't know if there is similar in Javascript.

kingdavidmartins:
Hey All :wave: :smile: 

So earlier last week I had an Idea. Thinking 
> "wouldn't it be nice to able to create some lib/package that could make adding/integrating a TDD to any project super simple & hastle free. Making life easier for both contributors/collaborators" 

And after spending sometime thinking about it today. I decided to spend a little time brain storming on what could be done to keep the barrier to entry low all while being able to add/integrate a TDD evironment

I ended up comming up with an idea of a lib/package that works based on specified flags. I believe by adding flags the time taken to Integrate/Add TDD with existing/new projects can be reduced down by multiple factors and/or a factor of 10    

### *I'll ellaborate through the following example*

> ### compact

Removes falsey values from an array.

Use `Array.filter()` to filter out falsey values (`false`, `null`, `0`, `""`, `undefined`, and `NaN`).

```js
/* PF: */ const compact = arr => arr.filter(Boolean);
[...]

PR #79, opened 2017-12-13
Testing With Tape & Making Snippets Modular

kingdavidmartins:
As I described in issue #78 
> I wanted to know if there's a specific reason for not using TDD. I believe a light weight testing library like tape can provide a lot of value when trying to test code snippets. Especially snippets that are unresponsive or break when they come across edge cases

> I believe having tests automated can also help the community test each snippet more rigorously and consistently before and after each PR

I've taken the initiative by giving an example of implementing tape into the project

I've currently only updated one method which is RGB-to-hexidecimal.

After checking out this PR 

All you need to do is run the following

```
npm test
```

Let me know what you think

 If you like to move the project in a TDD direction with tape just say the word and I could do the same with all the code snippets currently in the projects directory.


Chalarangelo:
I like the idea of using TDD, but this is too much. Contributors don't always have time to write tests and deal with all the edge-cases. This will make it very difficult for first-time contributors and take a long time for each snippet to be added. Also, he file structure will soon end up an unmaintainable mess with multiple files that need to be changed individually after any changes are made to each snippet. 

If we are to implement TDD, we need to setup some kind of script inside our builder that will read the snippets and test them for a few cases. Cases should be stored elsewhere and they should not have to be provided by contributors, as they will only be used to evaluate if the snippets work correctly. I expect we need a maintanance team for this sort of thing, I definitely cannot handle adding test-cases for every snippet that is being added and I have no clue how to deal with the testing in general (not yet anyways).

Chalarangelo:
As discussed on [gitter](https://gitter.im/30-seconds-of-code/Lobby), I'm putting this on hold for a little bit until we figure out what to do about the testing. Thanks for the help!

Chalarangelo:
Refer to #78 for the conversation on this topic. We will figure it out together! 😉

PR #80, opened 2017-12-13
Simplify the Even or odd number snippet

adrianklimek:
In this case using `Math.abs()` is unnecessary since negative even number modulo 2 returns -0.
```
-0 === 0 // true
```


Chalarangelo:
I remember that the `Math.abs()` was there to deal with some edge-case, but I can't remember why. I'll test it further and get back to you!

PR #82, opened 2017-12-13
Use Array.from instead of slice to capitalize a string

prodigic:
ES6 Array.from is a clearer approach, imo. ;)

### DIFF
--- a/snippets/capitalize-first-letter.md
+++ b/snippets/capitalize-first-letter.md
@@ -1,10 +1,9 @@
 ### Capitalize first letter
 
-Use `slice(0,1)` and `toUpperCase()` to capitalize first letter, `slice(1)` to get the rest of the string.
+Use `Array.from()` and `toUpperCase()` to capitalize first letter.
+
 Omit the `lowerRest` parameter to keep the rest of the string intact, or set it to `true` to convert to lower case.
 
 ```js
-const capitalize = (str, lowerRest = false) =>
-  str.slice(0, 1).toUpperCase() + (lowerRest? str.slice(1).toLowerCase() : str.slice(1));
-// capitalize('myName', true) -> 'Myname'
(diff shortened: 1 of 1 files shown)


snippets/capitalize-first-letter.md
```diff
@@ -1,10 +1,9 @@
...
+Use `Array.from()` and `toUpperCase()` to capitalize first letter.
+
 Omit the `lowerRest` parameter to keep the rest of the string intact, or set it to `true` to convert to lower case.
 
 ```js
-const capitalize = (str, lowerRest = false) =>
-  str.slice(0, 1).toUpperCase() + (lowerRest? str.slice(1).toLowerCase() : str.slice(1));
-// capitalize('myName', true) -> 'Myname'
-```
+const capitalize = (word, forceLowerCase = false) =>  Array.from(forceLowerCase ? word.toLowerCase() : word, (l,i) => i === 0 ? l.toUpperCase() : l).join("")
```
Chalarangelo: Why change `lowerRest` to `forceLowerCase`? I think it was fine as it was.

Chalarangelo:
I want to run a performance test to see if this has any drawbacks. Otherwise, I'm not against it, although the previous one might have looked cleaner. Opinions and/or jsperf checks appreciated.

snippets/capitalize-first-letter.md
prodigic: yeah, just felt it was more explicit to the action it performs, not that important!

Chalarangelo: I'll run jsPerf in the morning and see if there's any major performance hit. Otherwise, this is solid. I still prefer `lowerRest`, but I might be too opinionated on that, so leave it any way you prefer it. Thanks for the help!

Chalarangelo:
`Array.from()` is **significantly** slower (about 6 times) - [performance test results](https://jsperf.com/capitalize-first-letter-82). Let's stick with the original due to that. Thanks for putting in the time to try this, though!

PR #83, opened 2017-12-13
Added snippet for square root

arjunmahishi:
Its a simple snippet to calculate the square root of a number


arjunmahishi:
You can test it here
https://jsfiddle.net/arjunmahishi/bxmurq6b/1/#&togetherjs=vaKNE1irgp

Chalarangelo:
There is a native `Math.sqrt()` function.

PR #85, opened 2017-12-13
group by

elderhsouza:
Tried to reproduce the groupBy behaviour from the lodash lib.
https://lodash.com/docs/4.17.4#groupBy

### DIFF
--- a/snippets/group-by
+++ b/snippets/group-by
@@ -1,16 +1,14 @@
 ### Group by
 
-Passing an array of values, a function or a property name thats going to be run against each value in the array,
-returns an object where the keys are the mapped results and the values is an array of the original values that generated the same results.
+Use `Array.map()` to map the values of an array to a function or property name.
+Use `Array.reduce()` to create an object, where the keys are produced from the mapped results.
 
 ```js
-const groupBy = (values, fn) => {
-  return (typeof fn === 'function' ? values.map(fn) : values.map((val) => val[fn]))
+const groupBy = (arr, func) => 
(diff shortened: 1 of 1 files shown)


snippets/group-by
```diff
@@ -0,0 +1,16 @@
+### Group by
+
+Passing an array of values, a function or a property name thats going to be run against each value in the array,
+returns an object where the keys are the mapped results and the values is an array of the original values that generated the same results.
+
+```js
+const groupBy = (values, fn) => {
+  return (typeof fn === 'function' ? values.map(fn) : values.map((val) => val[fn]))
+    .reduce((acc, val, i) => {
+      acc[val] = acc[val] === undefined ? [values[i]] : acc[val].concat(values[i]);
```
darrenscerri: How about `acc[val] = (acc[val] || []).concat(values[i]);`?

snippets/group-by
```diff
@@ -0,0 +1,16 @@
+### Group by
+
+Passing an array of values, a function or a property name thats going to be run against each value in the array,
+returns an object where the keys are the mapped results and the values is an array of the original values that generated the same results.
+
+```js
+const groupBy = (values, fn) => {
+  return (typeof fn === 'function' ? values.map(fn) : values.map((val) => val[fn]))
```
darrenscerri: No need to enclose `(val)` in braces. Follows syntax convention.

elderhsouza:
Yes. I'll edit to implement your suggestions. Thanks!

PR #86, opened 2017-12-13
Add 'sleep' function

sergei-zelinsky:
Function blocks an execution thread for `ms` milliseconds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@
 * [Scroll to top](#scroll-to-top)
 * [Shuffle array values](#shuffle-array-values)
 * [Similarity between arrays](#similarity-between-arrays)
+* [Sleep](#sleep)
 * [Sort characters in string (alphabetical)](#sort-characters-in-string-alphabetical)
 * [Sum of array of numbers](#sum-of-array-of-numbers)
 * [Swap values of two variables](#swap-values-of-two-variables)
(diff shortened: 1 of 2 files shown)


darrenscerri:
This is extremely bad practice. The main thread will block, which would mean that no other work can be done (page becomes completely unresponsive in browser, for example). Do you have a practical use-case where this is needed for which there is no other alternative solution?

sergei-zelinsky:
Hi, @darrenscerri. Thank you for your response.
Yes, I know about thread's blocking:
> Function blocks an execution thread for `ms` milliseconds.

and I fully agree with you about execution in a browser environment.

I think it may be useful for writing console utils in NodeJS when you need to block the main thread for some period of time and after that continue an execution.

darrenscerri:
A `setTimeout(fn, ms)` would be better in all cases I can think of. Do you have a particular use-case where blocking the main thread is the only solution?

Chalarangelo:
I agree with @darrenscerri. This seems like a bad idea. We also have #84 which does something very similar in a better fashion.

Thanks for contributing, anyways! 👍

sergei-zelinsky:
I tried to imagine some use cases, but nothing has gone into my mind, which is not impossible to be rewritten in an async way. Maybe it's useless in the real world.

Thank you, guys, for showing me the right way 👍 

I am closing this.

PR #87, opened 2017-12-13
add pick code snippet

kingdavidmartins:
similar to lodash pick method 

https://lodash.com/docs/4.17.4#pick

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@
 * [Median of array of numbers](#median-of-array-of-numbers)
 * [Object from key value pairs](#object-from-key-value-pairs)
 * [Percentile](#percentile)
+* [Pick](#pick)
 * [Pipe](#pipe)
 * [Powerset](#powerset)
 * [Promisify](#promisify)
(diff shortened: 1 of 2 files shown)


darrenscerri:
This is a very useful method. However, this is a simpler solution:

```js
const pick = (obj, arr) => arr.reduce((acc, curr) => (acc[curr] = obj[curr], acc), {});
```

kingdavidmartins:
Nice!! I definitely agree the solution is much more elegant is simpler.

@darrenscerri  I was thinking what if the arr passed with keys the users wants doesn't exist.  Rather than just reducing over something that doesn't exist and returning undefined. I can use filter to filter it out.

darrenscerri:
Good point. Something like this guards that:

```js
const pick = (obj, arr) => arr.reduce((acc, curr) => (curr in obj && (acc[curr] = obj[curr]), acc), {});
```

kingdavidmartins:
Nice! Agreed 💯

PR #88, opened 2017-12-13
Remove duplicate shuffle array snippets. Simplify implementation

darrenscerri:


Chalarangelo:
Thanks for contributing! This issue has been brought up before, I think. I only kept the order randomizer as it is biased and produces different results than `shuffle`. Not certain if we should keep it, though. Any other opinions?

darrenscerri:
"Shuffle array values" is an incorrect implementation. The sort comparator arguments are the actual values and not indices, so `r[a]` and r[b]` will be `undefined` unless the keys existence coincidentally.

Chalarangelo:
Check [Will it Shuffle?](https://bost.ocks.org/mike/shuffle/compare.html) to see my justification.

The `randomize-order-of-array.md` snippet is under `sort (random comparator)` and out `shuffle-array-values.md` is `sort (random order)`. The first one is biased, the second one isn't. Also, it seems that the `shuffle` method works perfectly fine. Am I missing something?

darrenscerri:
`shuffle([4,5,6])` will always return `[4,5,6]`.

Chalarangelo:
You're right @darrenscerri. That is really bad. Ok, I'm gonna merge your PR then.

darrenscerri:
There is still some bias with the merged solution, but is fine for most use cases.

PR #92, opened 2017-12-13
refactor count-occurrences | easier & simpler

kingdavidmartins:
Simplify implementation, easier to understand, and shorter


Chalarangelo:
This was brought up before. Based on [testing](https://jsperf.com/count-occurrences-92), `Array.reduce()` is faster than `Array.filter().length)`, therefore we went with `Array.reduce()`.

kingdavidmartins:
Ahhh makes sense kay

Issue #95, opened 2017-12-14
URL parameters:  The regular expression question mark is superfluous


katongzhong1:
I tested, the correct string is ([^?=&]+)(=([^&]*)), no ([^?=&]+)(=([^&]*))?
 Am i wrong?

katongzhong1:
Never mind! It has changed the way just now

katongzhong1:
Sorry, I has tried the new way, the same Problem is founded. The regular expression question mark is superfluous

skatcat31:
With current code:

```js
const getUrlParameters = url =>
  url.match(/([^?=&]+)(=([^&]*))?/g).reduce(
    (a, v) => (a[v.slice(0, v.indexOf('='))] = v.slice(v.indexOf('=') + 1), a), {}
  );
undefined
getUrlParameters('http://url.com/page?name=Adam&surname=Smith')
var result = {"http://url.com/pag": "http://url.com/page", "name": "Adam", "surname": "Smith"}
```

Suggested change:
```js
const getUrlParameters = url =>
  url.match(/([^?=&]+)(=([^&]*))/g).reduce(
    (a, v) => (a[v.slice(0, v.indexOf('='))] = v.slice(v.indexOf('=') + 1), a), {}
  );

```

Chalarangelo:
Resolved in #96

PR #96, opened 2017-12-14
Update URL Paramters snipper

skatcat31:
Remove the Question mark as per https://github.com/Chalarangelo/30-seconds-of-code/issues/95


skatcat31:
Will update ot the individual .md file

PR #97, opened 2017-12-14
the chunk Array:  use Array.from instead of Array.apply

jackPanyj:
use Array.from instead of Array.apply


Chalarangelo:
Performance-wise, this is faster ([results](https://jsperf.com/chunk-array-97)). Thanks for the contribution! I am merging right away!

PR #98, opened 2017-12-14
add compact.md & ran npm run build-list

kingdavidmartins:
similar to lodash compact method

Use `.filter()` to filter falsey values. For example false, null, 0, "", undefined, and NaN are falsey.

`compact([0, 1, false, 2, '', 3, 'a', 'e'*23, NaN, 's', 34]) -> [ 1, 2, 3, 'a', 's', 34 ]`

https://lodash.com/docs/4.17.4#compact

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ const chunk = (arr, size) =>
 Use `.filter()` to filter falsey values. For example false, null, 0, "", undefined, and NaN are falsey.
 
 ```js
-const compact = (arr) => arr.filter( v => [0, null, false, '', undefined].indexOf(v) === -1 && v);
+const compact = (arr) => arr.filter(v => v);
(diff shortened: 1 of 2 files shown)


darrenscerri:
Compact is a useful snippet. This implementation is better:

```js
const arr => arr.filter(v => v);
```

Basically, it's keeping all values which are not falsy (truthy).

Chalarangelo:
I was about to merge, but @darrenscerri's code seems a lot shorter and cleaner. Are we certain this does not break under specific corner cases?

kingdavidmartins:
Huh I thought I knew a few that would break it, but after testing it again seems @darrenscerri does work against all cases. Guess I need some sleep lol @darrenscerri Thanks!!

Chalarangelo:
Thanks for implementing this! I'll merge and, if we somehow manage to break it, we can change the code later down the line.

Issue #41, opened 2017-12-12
Validate number: use only isFinite?


arnaudjuracek:
> ### Validate number
> Use !isNaN in combination with parseFloat() to check if the argument is a number. Use isFinite() to check if the number is finite.
```js
const validateNumber = n => !isNaN(parseFloat(n)) && isFinite(n);
// validateNumber('10') -> true
```

---
According to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite), `isFinite()` returns "**`false`** if the argument is positive or negative `Infinity` or `NaN`; otherwise, **`true`**". 

```js
isFinite(Infinity);  // false
isFinite(NaN);       // false
isFinite(-Infinity); // false

isFinite(0);         // true
isFinite(2e64);      // true
isFinite(910);       // true

isFinite(null);      // true, would've been false with the 
                     // more robust Number.isFinite(null)

isFinite('0');       // true, would've been false with the 
                     // more robust Number.isFinite("0")
```

So technically, the `!isNan()` isn't required and you could shorten your `validateNumber`:
```js
const validateNumber = n => isFinite(n);
// validateNumber('10') -> true
```

Chalarangelo:
I think I based the snippet on [this SO answer](https://stackoverflow.com/a/1830844/1650200). It might be outdated or there might be something I don't understand. Does anyone know of any cases where `Number.isFinite()` alone would fail?

fesebuv:
🤔  what about? 
```
const validateNumber = n => typeof n === 'number';
```

atomiks:
@fesebuv I believe the function is meant to allow a number string as well.

Also OP, what about booleans?

```javascript
const validateNumber = n => isFinite(n)
validateNumber(true) // true
```

elderhsouza:
@atomiks @fesebuv @Chalarangelo 

I think this would do:
```js
const validateNumber = n => isFinite(parseFloat(n))
```

Chalarangelo:
Yes, the function is meant to allow for strings containing numerical data, such as `'10'` for example. @elderhsouza I think this should probably work as intended. Can anyone find any test cases that break this? Otherwise, we should get a PR going and change the current solution to this.

atomiks:
@elderhsouza 

```javascript
validateNumber('20px') // true
```

Chalarangelo:
[...]

PR #101, opened 2017-12-14
Fix: validateNumber: check cooercian

skatcat31:
now passes all 16 test cases:
```js
[
validateNumber('10') // true
,!validateNumber(false) // false
,!validateNumber(true) // false
,validateNumber(10) // true
,validateNumber(0.10) // true
,validateNumber(010) // true
,validateNumber(0xA) // true
,!validateNumber(NaN) // false
,!validateNumber(null) // false
,!validateNumber(undefined) // false
,!validateNumber('10%') // false, % is a construct
,!validateNumber('10px') // false
,!validateNumber(Infinity) // false
,validateNumber(Math.PI) // true, but is such an extreme edge case...
,validateNumber(0) // true, will probably wreck things
,validatenumber(1e10) // true
]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -729,7 +729,7 @@ Use `!isNaN` in combination with `parseFloat()` to check if the argument is a nu
 Use `isFinite()` to check if the number is finite.
 
 ```js
-const validateNumber = n => !isNaN(parseFloat(n)) && isFinite(n) && Number(n) == n;
+const validateNumber = n => !isNaN(parseFloat(n)) && isFinite(n);
 // validateNumber('10') -> true
 ```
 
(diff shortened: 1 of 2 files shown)


skatcat31:
https://github.com/Chalarangelo/30-seconds-of-code/issues/41

README.md
```diff
@@ -729,7 +729,7 @@ Use `!isNaN` in combination with `parseFloat()` to check if the argument is a nu
 Use `isFinite()` to check if the number is finite.
```
Chalarangelo: We should update the description to make sure people understand what it is that the snippet does. Also make changes to the `validate-number.md` file, not README. Apart from that, we are solid! ;)

skatcat31:
I'll update my fork to a different branch and update the individual MD for the PR

Chalarangelo:
Thanks @skatcat31!

skatcat31:
Updated to change the individual .md

PR #113, opened 2017-12-14
Added American to English date

Hammie217:


Chalarangelo:
Is this really necessary? I mean I know I have needed it, but is it as common as I think it is? 🤣

Hammie217:
True, I have to use it a fair amount but that might just be because i'm english

Chalarangelo:
Community, give us your opinions! I, for one, am not against this in all honesty. We can generalize the name to make it apply both ways (I mean if you reverse an English date you get an American one and vise versa), so we can call it something like `switchDayMonth` or something.

Hammie217:
Sounds good!

LeandroFranciscato:
Humm... I'm from Brazil, here we have a lot of this problem. Btw, if you will call it for `switchDayMonth`, there is a more elegant way.

`const switchDayMonth = (date) => date.split('/').reverse().join('/');`
`// switchDayMonth("31/12/2017") -> "2017/12/31"`
`// switchDayMonth("2017/12/31") -> "31/12/2017"`

snippets/American-to-English-Date.md
```diff
@@ -0,0 +1,7 @@
+### American to english date
+Take an American date and split the string re-arranging into english format.
+```js
+var americanDate = "5/10/2004";
```
skatcat31: These are globally scoped and not a function. It also is not very adaptable.

Using the date object and a function you can also get some error checking should something go wrong(return undefined or adopt other methods such as allowing to bubble, up to choice)

Possible replacement:

```js
const toEnglishDate = (time) => {try{return new Date(time).toISOString().split('T')[0].replace(/-/g, '/')}catch(e){return}}
```

Chalarangelo: I second this, more of a general-case, should be more useful this way. @Hammie217 you can update your PR if you agree, so we can merge this.

skatcat31: For a version that the developer must handle the errors of themselves(fails on invalid date, not a Maybe):

```js
const toEnglishDate = (time) => new Date(time).toISOString().split('T')[0].replace(/-/g, '/')
```

Chalarangelo:
The PR has been sitting here for a while, I'll merge and deal with the suggested changes myself!

PR #117, opened 2017-12-14
Create sqaure_it.md

meetzaveri:
I don't exactly know the use of this snippet but contributors can recommend on this approach of `.reduce()`.


meetzaveri:
@Chalarangelo If this PR is suitable, then further I'll make it according to Contributing guidelines and much smaller. Waiting for your response!

Chalarangelo:
If your aim is to get key-value pairs with the key being the original value and tha value being the value after applying square, then you could do this:
```js
const squareArr = arr => Object.assign(...arr.reduce((acc,v) => acc.concat({[v]: v * v}), []));
```

It is a lot shorter and cleaner. I cannot find any super useful use-cases for this, but maybe we could add a generalization where we create an object that maps values, using a function (I think we have one of those already, but I am too tired to look around for it). Something like:

```js
const mapArray = (arr, func) => Object.assign(...arr.reduce((acc,v) => acc.concat({[v]: func(v)}), []));
```

And then you would do the whole square-mapping like so: `mapArray([1,2,3,4,5,6,7], x => x * x)`

meetzaveri:
@Chalarangelo The first snippet is better, Okay it has to be more abstracted view on this point. It's your decision man.

Chalarangelo:
Could use some community opinions on this.

skatcat31:
A more abstract version, instead of square it, would be a zipper with the original and a map:

```js
const mapToZip = (arr, fn) => [arr, arr.map(fn)] // create a zip list of original values, and it's mapped version
```

with the now ziplist you can then do `zipObject` to make an object with key: original, value: mapped

```js
const zipObj = zipList => zipList[0].reduce((a,v,i) => (a[v] = zipList[1][i], a)
```

then using a composer you can make a squareIt function:

```js
const squareIt = arr => zipObject(zipWith(arr, a => a*a))
```

so this abstract version could actually start showing people composition, and how libraries can be formed from several other helper functions!

Or you could inline and closure things correctly for a really cool function monster that would need to abuse the very core of javascript with function operators but would very neatly show some tricks

skatcat31:
A more obtuse example of my monster would be using an array of users in a closure to make an object of users by their email address or ids, but at that point it's getting to such wizardy I think I've saturated what ES6 could possibly do in a single line of abstract art?

skatcat31:
```js
const squareIt = arr => (a => (a = [arr, arr.map(a => a*a)], a[0].reduce( (acc,val,ind) => (acc[val] = a[1][ind], acc), {}) )) ( )
```

there... a monster of a snippet with `squareIt`'s initial purpose

```js
squareIt([1,2,3])
```

meetzaveri:
@Chalarangelo  I like the way @skatcat31 says as that recent example of using accumulator and returning direct sqaured values only.

Chalarangelo:
I really like @skatcat31's final iteration of `mapObject`, I'd love to have that added, but I can't really understand what it says. So, to finally merge this PR, let's do the following:
- @meetzaveri Update the existing code to be the last iteration of `mapObject` and use the example for the square as its example.
- @skatcat31 Kindly explain how that thing works, so we can describe it in the snippet.
- I will merge after you guys are done with that. This will end up being a really awesome snippet!

meetzaveri:
Now snippet description needed from @skatcat31

skatcat31:
Since I don't have write permissions to @meetzaveri repo I'll post the description here unless he'd like me to PR to his.

```js
const mapObject = (arr, fn) => (a => (a = [arr, arr.map(fn)], a[0].reduce( (acc,val,ind) => (acc[val] = a[1][ind], acc), {}) )) ( )
```

Hoo boy here we go...

# Description
Using an anonymous inner function scope we declare an undefined memory space using closures for storing a return value. We then use a new Array to store the array with a map of the function over it's data set and a comma operator to return a second step without needing to move from one context to another thank to closures and order of operations.

Notice the need to declare an order of operations(parenthesis) around the anonymous inner function so we can call it immediately.

Chalarangelo:
I will update and merge tomorrow, as I'm done for the day. Thank you guys! 👍

ForsakenHarmony:
Why not 
```js
const mapObject = (arr, fn) => arr.reduce((acc, val) => ({...acc, [val]: fn(val)}),{});`
```

skatcat31:
Wouldn't changing you snippet to
```js
const mapObject = ( arr, fn ) => arr.reduce( ( acc, val ) => ( acc[ val ]= fn( val ), acc ), { } );
```
avoid having to spread the object every time and instead just accumulate into the object?
____
Your snippet is def. smaller as mine was more an opportunity to show how `mapZipList` and `zipObj` can be used to construct an object out of composable library elements, and eventually shown that such could also be compressed into an anonymous closure. Both will do the same. Mine was merely a for fun example of how composition can be compressed and was not optimized on purpose.

ForsakenHarmony:
Well, composition is nice and all, but writing it out doesn't really work that well if there's a version that's way easier to read

Issue #72, opened 2017-12-13
Curry


broadsw0rd:
What about this kind of curry?

```js
const curry = (fn, ...args) =>
  fn.length <= args.length
    ? fn(...args)
    : curry.bind(null, fn, ...args)
```

elderhsouza:
Can you show an example of how that would work?
I believe this don't account for currying of variadic functions like Math.min for example.

broadsw0rd:
@elderhsouza yes, you right, it was just an example, we have to define arity

```js
const curry = (fn, arity = fn.length, ...args) =>
  arity <= args.length
    ? fn(...args)
    : curry.bind(null, fn, arity, ...args)

// curry(Math.pow)(2)(10) -> 1024
// curry(Math.min, 3)(10)(50)(2) -> 2
```

linrui1994:
@elderhsouza  the curry may not work 
```js
const curry = (f, arity = f.length, next) =>
  (next = prevArgs =>
    nextArg => {
      const args = [ ...prevArgs, nextArg ];
      return args.length >= arity ? f(...args) : next(args);
    }
  )([]);
// curry(Math.min, 3)(10, 200)(2)  -> not work
```
I think should like this
```js
const curry = (f, arity = f.length, next) =>
  (next = prevArgs =>
    (...nextArg) => {
      const args = [ ...prevArgs, ...nextArg ];
      return args.length >= arity ? f(...args) : next(args);
    }
  )([]);
```

elderhsouza:
@linrui1994 I like this one, haven't used currying like this one yet, but I can see how it can be useful, solid implementation.

@broadsw0rd That's really elegant and terse, I tested all versions with the same use cases and it seems to work for all of it.

@Chalarangelo Great discussion here, wanna jump in?

Chalarangelo:
If it works better than the previous one or covers more use-cases, PR it. I like what I see. 😉

Chalarangelo:
Anybody wanna open a PR for this? I'd be glad to merge.

elderhsouza:
@Chalarangelo  Done! 👍

Chalarangelo:
Resolved in #124. Closing...

PR #124, opened 2017-12-14
minor refactor to curry

elderhsouza:


Chalarangelo:
Thanks! Is this the one discussed in #72 ?

elderhsouza:
@Chalarangelo Yes, that's the one.

PR #125, opened 2017-12-14
adding function to clean json objects, deeply.

iamsoorena:
this function removes unwanted keys from json objects deeply.


iamsoorena:
the problems about snippet are resolved. and frankly I don't think it's possible to mimic this functionality with Array.map() since we can't use it on a json. maybe I didn't get what you meant.

PR #131, opened 2017-12-14
Added snippet to convert number to array of digits

arjunmahishi:
Converts a number to an array of digits.

### DIFF
--- a/snippets/number-to-array-of-digits.md
+++ b/snippets/number-to-array-of-digits.md
@@ -1,4 +1,4 @@
-### Number to array to digits
+### Number to array of digits
 
 use `parseInt()` to get only the integer part of the quotient.
 use `Array.reverse()` to return the array in the same order as the number. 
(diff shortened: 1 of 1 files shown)


snippets/number-to-array-of-digits.md
```diff
@@ -0,0 +1,17 @@
...
+```js
+const num2array = (n) =>{
+	let arr = [];
+	while (n>0) { 
+		arr.push(n%10); 
+		n = parseInt(n/10); 
+	}
+	return arr.reverse();
+}
+
```
Chalarangelo: You can use `(''+n).split('').map(parseInt)` to achieve exactly the same thing. Probably more performant, too.

arjunmahishi: done

Chalarangelo:
For whatever odd reason this works
```js
const digitize = n => (''+n).split('').map(i => parseInt(i));
```
but this doesn't:
```js
const digitize = n => (''+n).split('').map(parseInt);
```
I'm merging the first  one, then!

PR #132, opened 2017-12-14
method name

b117020:
code to get name of current method

### DIFF
--- a/snippets/current method name.txt
+++ b/snippets/current method name.txt
@@ -1,0 +1,3 @@
+###code to view name of current method
+
+String methodName = Thread.currentThread().getStackTrace()[1].getMethodName();
--- a/snippets/current method name.txt
+++ b/snippets/current method name.txt
@@ -1,3 +1,0 @@
-###code to view name of current method
-
-String methodName = Thread.currentThread().getStackTrace()[1].getMethodName();
--- a/snippets/log-method-name.md
+++ b/snippets/log-method-name.md
@@ -1,0 +1,8 @@
+### Log method name
+
+Use `console.debug()` and the `name` property of the passed method to log the method's name to the `debug` channel of the console.
+
+```js
+const methodName = fn => (console.debug(fn.name), fn);
+// methodName(methodName) -> methodName (logged in debug channel of console)
+```


skatcat31:
I'm having trouble seeing a use case for this since in most production environments an uglify process is used to condense packages into much smaller and easier to transit forms that are faster. Those destroy the name of the method. It's also a variable that would need to be copy and pasted into every method. It'd be better if it were a function that accepted functions and got their derived name instead.

[MDN states that Function.prototype.name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name) would meet the exact same needs, and can be written as a function easily that isn't reliant on scope:

```js
const methodName = fn => new String(fn.name)
```

Considering that in reality the field is read only and non enumerable the wrapping in new String shouldn't be needed ,but it's always good to sanitize things you don't want to change for everyone else.

NOTE: this still gets ruined by uglify processes, so production use is extremely limited and is often best left to debuggers to inspect your stack

snippets/current method name.txt
```diff
@@ -0,0 +1,3 @@
+###code to view name of current method
+
+String methodName = Thread.currentThread().getStackTrace()[1].getMethodName();
```
skatcat31: I'm having trouble seeing a use case for this since in most production environments an uglify process is used to condense packages into much smaller and easier to transit forms that are faster. Those destroy the name of the method. It's also a variable that would need to be copy and pasted into every method. It'd be better if it were a function that accepted functions and got their derived name instead.

# Possible Change

MDN states that Function.prototype.name would meet the exact same needs, and can be written as a function easily that isn't reliant on scope:
```js
const methodName = fn => new String(fn.name)
```
Considering that in reality the field is read only and non enumerable the wrapping in new String shouldn't be needed ,but it's always good to sanitize things you don't want to change for everyone else ESPECIALLY if it comes from a prototype. So a safe copy is a good idea.

NOTE: this still gets ruined by uglify processes, so production use is extremely limited and is often best left to debuggers to inspect your stack

Chalarangelo: Hi @b117020 thanks for contributing! 

I agree with @skatcat31 in that we can only use it as a debugging utility. Maybe we should stick the output to a proper `console` call, instead of returning a `string` to make sure that developers only use it for debugging. 

I also agree we should use the native `Function.prototype.name` property.

skatcat31: Possible change with logging instead of return, making it an identity function

```js
const methodName = fn => (console.debug(fn.name), fn)
```

PR #140, opened 2017-12-14
Create compose-functions.md

atomiks:


snippets/compose-functions.md
```diff
@@ -0,0 +1,15 @@
+### Compose functions
+
+Use the `...rest` operator to gather all function arguments into an array. Return a function which takes
+a single argument and uses `Array.reduceRight()` to return the result of applying each function.
+
+```js
+const compose = (...fns) => n => fns.reduceRight((acc, fn) => fn(acc), n);
```
skatcat31: This is unary. Composition function are expected to be variadic and then unary, however it looks like this one is purely unary

skatcat31: https://github.com/Chalarangelo/30-seconds-of-code/pull/141/files should show an easy fix for this. It's a little hard to understand, but it works like a charm!

Chalarangelo:
Excuse me for asking but what is the difference between composing and what we have in `pipe` right now? If they are the same, maybe we should rename `pipe` to `compose` as it sounds a lot better to me.

atomiks:
This snippet was taken from here btw:

https://medium.com/javascript-scene/reduce-composing-software-fe22f0c39a1d

Chalarangelo:
Labeling `on hold` because I cannot really see the difference between this and `pipe`. Maybe we should rename `pipe` though as it reads like composition to me. I'm confused...

atomiks:
It seems like they're reverses of each other... lol

Chalarangelo:
@atomiks Ahh, ok, that makes some sense. Let me run a couple of tests and I'll get back on this in a few minutes!

Chalarangelo:
Alright it does work and you are right about it being a reverse of `pipe`. What about @skatcat31's comment? You think we can make it work with variadic arguments?

atomiks:
I have no idea what "variadic" means 😆 

Feel free to do that stuff to it...

Chalarangelo:
It means that it can accept a variable amount of arguments. I can't figure out how to make the change myself, but I want to merge the PR to get the compose script in right away! @skatcat31 could you PR the changes later?

atomiks:
Right... so not just one argument? Then the rest operator can be used?

Chalarangelo:
@atomiks Yes, it can be used to do that. Do you want to try your hand at it? I can wait till the merge, no problem.

atomiks:
I believe that it's the exact same as the new `pipe()`, just using `.reduceRight` instead. Or... the `f` and `g` get swtiched, but I tried the former and it got the same result (17).

Chalarangelo:
So, the only difference is using `reduceRight()` instead? Hmmm, we could possibly generalize the `pipe` function to accept an argument so that we can conditionally `reduce` or `reduceRight` I think.

justin0022:
I've written a variadic version of compose. I'll be happy to submit a PR once this is merged!

Chalarangelo:
Great @jussisaurio but I'll probably merge tomorrow. Sit tight and wait for this PR to be merged! (I'm taking a long break because I'm basically burnt out from working on the project round the clock for half a wekk now, whoops!)

PR #141, opened 2017-12-14
Rewrote pipe such that the first function may have any arity.

justin0022:
Now pipe's first function can accept an arbitrary number of arguments; every subsequent function must be unary. The arguments to reduce also use 'f' and 'g', which makes pipe more intuitive to read as it closely resembles its mathematical background.

### DIFF
--- a/snippets/pipe.md
+++ b/snippets/pipe.md
@@ -1,8 +1,18 @@
 ### Pipe
 
-Use `Array.reduce()` to pass value through functions.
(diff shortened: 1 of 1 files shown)


skatcat31:
Love it. Varidic pipe function composition. Can be reused with same result assuming the functions don't end up being weird closures... In ES6. Well done!

prodigic:
damn that's nice!

PR #143, opened 2017-12-15
Create hexadecimal-to-RGB.md

Pl4gue:
Snippet for converting hexadecimal colorcodes to RGB colorcodes. Also makes use of ECMAScript, working proof here: [JSFiddle](https://jsfiddle.net/pl4gue/ozpLxqsk/)


kingdavidmartins:
Hey @Pl4gue 👋 

I wanted to know if there was a reason you didn't go with  
```javascript
const hexToRgb = hex => hex.slice(1).match(/.{2}/g).map(x => parseInt(x, 16))
```
compared to what was submitted I believe this is more elegant and simpler

kingdavidmartins:
@Pl4gue Also rather than return an array -> [#, #, #]. 
You could return your result in a rgb format -> rgb(#, #, #)

```javascript
const hexToRgb = hex => `rgb(${hex.slice(1).match(/.{2}/g).map(x => parseInt(x, 16)).join()})`
```

skatcat31:
I feel like this could be a problem if people don't understand it's a hex color code string, not a hexidecimal number...

```js
const validHex = 0xFF // 255, but it's not a hex color code
```

Pl4gue:
@skatcat31Thank you for the feedback, I'll rename it :+1:

Pl4gue:
@kingdavidmartins Thank you for the feedback, I'll apply your suggestions :+1:

PR #144, opened 2017-12-15
Add takeRight

kingdavidmartins:
similar to lodash takeRight method

### takeRight

Use `Array.slice()` to create a slice of the array with `n` elements taken from the end.

```js
const takeRight = (arr, n = 1) => arr.slice(arr.length - n, arr.length);
// takeRight([1, 2, 3], 2) -> [ 2, 3 ]
// takeRight([1, 2, 3]) -> [3]
// takeRight([1, 2, 3], 0) -> []
```
https://lodash.com/docs/4.17.4#takeRight

### DIFF
--- a/snippets/takeRight.md
+++ b/snippets/takeRight.md
@@ -1,0 +1,10 @@
+### takeRight
+
+Use `Array.slice()` to create a slice of the array with `n` elements taken from the end.
+
+```js
+const takeRight = (arr, n = 1) => arr.slice(arr.length - n, arr.length);
+// takeRight([1, 2, 3], 2) -> [ 2, 3 ]
+// takeRight([1, 2, 3]) -> [3]
+// takeRight([1, 2, 3]) -> []
+```
--- a/tag_database
+++ b/tag_database
@@ -78,6 +78,7 @@ sum-of-array-of-numbers:array
 swap-values-of-two-variables:utility
 tail-of-list:array
 take:array
+takeRight:array
 truncate-a-string:string
 unique-values-of-array:array
 URL-parameters:utility


snippets/takeRight.md
```diff
@@ -0,0 +1,10 @@
+### takeRight
+
+Use `Array.slice()` to create a slice of the array with `n` elements taken from the end.
+
+```js
+const takeRight = (arr, n = 1) => arr.slice(arr.length - n, arr.length);
```
skatcat31: You can save some typing
```js
cosnt takeRight = (arr, n = 1) => arr.slice(0-n)
```

kingdavidmartins: Hey @skatcat31 👋 

I used arr.length because takeRight works by slicing from the right ~ ( end) to the left

Using array.length assures that everything works as expected

Example 

*Lodash*

```js
// _.takeRight([1, 2, 3], 2) -> [ 2, 3 ]
// _.takeRight([1, 2, 3]) -> [3]
// _.takeRight([1, 2, 3], 0) -> []
```

'My Proposed Snippet*

```js
const takeRight = (arr, n = 1) => arr.slice(arr.length - n, arr.length);
// takeRight([1, 2, 3], 2) -> [ 2, 3 ]
// takeRight([1, 2, 3]) -> [3]
// takeRight([1, 2, 3], 0) -> []
```

Your Proposed Snippet*

```js
const takeRight = (arr, n = 1) => arr.slice(0-n)
// takeRight([1, 2, 3], 2) -> [ 2, 3 ]
// takeRight([1, 2, 3]) -> [3]
// takeRight([1, 2, 3], 0) -> [ 1, 2, 3 ] FAIL TEST : EXPECTED -> []
```

skatcat31: Good test! ternary could fix that, but then you get into a "why would you use this over the other" argument and at that point I don't have a good answer and never could. while more terse, you're right mine fails a pretty critical test case that could only be fixed by doing really dumb stuff instead of just simple math. Good catch

kingdavidmartins: @skatcat31 haha 😆  yup yup, Yea No Problem 😄 it's a pleasure & it's always good to have someone else review code. Might have caught something I could have missed and vise versa

PR #152, opened 2017-12-15
Create convert-json-date.md

sekmenhuseyin:
I hope you like this

### DIFF
--- a/snippets/convert-json-date.md
+++ b/snippets/convert-json-date.md
@@ -1,0 +1,14 @@
+### Convert Json Date
+ 
+Convert dates coming from ajax as json to readable format
+Example: converts "/Date(1489525200000)/" to "15/2/2017"
+ 
+ ```js
+cconst JsonToDate = arr =>
+    {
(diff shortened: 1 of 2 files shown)


Chalarangelo:
I'll need a bit to test this, but it looks promising. Thanks for contributing!

Emgo-Dev:
The getMonth() method returns current month as index of 11, so u need to +1 to it in your return right? Or am I missing something? Is it also common to store dates like this in JSON?

Chalarangelo:
As far as my experience suggests, I don't think anyone stores months zero-indexed. Any opinions?

Emgo-Dev:
This function seems more useful for its' parsing of a JSON string. In practice should be passed to a formatter function.

Emgo-Dev:
@Chalarangelo  I just did a test in the browser...is that no reliable?

Chalarangelo:
I don't suppose this will work any differently on Node @Emgo-Dev, so it probably stores months starting with January at zero.

PR #165, opened 2017-12-15
Fix broken link by changing the title

yashghelani:


Chalarangelo:
Good catch! Thanks for contributing!

PR #166, opened 2017-12-15
Added compose such that the rightmost function may have any arity.

justin0022:
Compose's last (rightmost) function can accept an arbitrary number of arguments; every subsequent function must be unary. The arguments to `Array.reduce()` also use 'f' and 'g', which makes pipe more intuitive to read as it closely resembles its mathematical background.

Also added spread operator into explanation of pipe.

### DIFF
--- a/snippets/array-zip.md
+++ b/snippets/array-zip.md
@@ -6,7 +6,7 @@ If lengths of the argument-arrays vary, `undefined` is used where no value could
 
 ```js
 const zip = (...arrays) => {
-  const maxLength = Math.max.apply(null, arrays.map(a => a.length));
+  const maxLength = Math.max(...arrays.map(x => x.length));
   return Array.from({length: maxLength}).map((_, i) => {
    return Array.from({length: arrays.length}, (_, k) => arrays[k][i]);
   })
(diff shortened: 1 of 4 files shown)


skatcat31:
Branches! A PR can come from any branch to be merged into the remote source. I could have one copy of the repo, with 12 branches, none of them my master, and a PR for each, and git wouldn't care. After everything gets merged in, I bring my repo up to date with the remote and I can continue with even more PRs.

So if you make a branch in your repo, make the changes to that branch, and then change the PR to point to that branch you'll be golden.

justin0022:
Ah, okay. So in this case, should I cancel this PR, go back to my forked repo, undo the commits, and then commit each individual change into its own branch, then submit 3 PRs?

Thanks for the info, I really appreciate it!

skatcat31:
You should be able to achieve this with a mixed reset and pushing the individual changes to other branches, or the method you described. Whichever is easiest for you will work.

justin0022:
Thanks, I'll try to get this fixed sometime in the weekend.

Chalarangelo:
I agree with @skatcat31 but there's no standard procedure just yet, so I'll merge to get these changes in. Plus even if we needed a revert, I couldn't even find my way through the branch tree right now with everything that's happening!

PR #177, opened 2017-12-16
Fix `Write json to file` anchor

demsking:


iamsoorena:
another solution for #165

PR #186, opened 2017-12-16
John Wolfe - edited filter-out-non-unique-values-in-an-array to be mo…

wolfejw86:
…re performant.

benchmark: 
john wolfe filter x 137,953 ops/sec ±0.65% (88 runs sampled)
current filter x 100,221 ops/sec ±1.02% (86 runs sampled)
Fastest is john wolfe filter


Chalarangelo:
This gives different results from the original snippet. Thanks for taking the time to contribute, though!

Issue #195, opened 2017-12-16
Array#remove is slow and description is not very clear.


kirilloid:
The `splice` and `concat` in a loop introduce O(N^2) solution.
Also, the current implementation has tricky side-effects. I'll address both things in a PR.

Chalarangelo:
I'm closing the issue, as we can talk about thins under #196. Thanks for taking the time to contribute!

PR #196, opened 2017-12-16
improve array.remove performance and description

kirilloid:
Addressing #195
Please, look at the description, which I think now reflects real function behaviour better.
Maybe, we need to change the behaviour instead?

### DIFF
--- a/snippets/array-remove.md
+++ b/snippets/array-remove.md
@@ -1,13 +1,22 @@
 ### Array remove
 
-Use `Array.filter()` to find array elements that return truthy values and `Array.reduce()` to remove elements using `Array.splice()`.
-The `func` is invoked with three arguments (`value, index, array`).
+Finds elements that satisfy a predicate and returns them, at the same time modifying an original array.
+In other words, it works like [R#partition](http://ramdajs.com/docs/#partition), except that it mutates the original array and returns another partition instead of returning two.
 
 ```js
-const remove = (arr, func) =>
-  Array.isArray(arr) ? arr.filter(func).reduce((acc, val) => {
-    arr.splice(arr.indexOf(val), 1); return acc.concat(val);
-    }, [])
-  : [];
-// remove([1, 2, 3, 4], n => n % 2 == 0) -> [2, 4]
+/**
+ * @param {E[]} arr
+ * @param {(E, number, E[]) => boolean}
+ * @returns mutated array
+ */
+const remove = (arr, func) => {
+  if (!Array.isArray(arr)) return [];
+  const matching = arr.filter(func);
+  const nonMatching = arr.filter((e, i) => !func(e, i, arr));
+  arr.splice(0, arr.length - 1, nonMatching);
+  return matching;
+}
+// const a = [1, 2, 3, 4];
+// remove(a, n => n % 2 == 0) -> [2, 4]
+// a -> [1, 3]
 ```


snippets/array-remove.md
```diff
@@ -1,13 +1,22 @@
 ### Array remove
 
-Use `Array.filter()` to find array elements that return truthy values and `Array.reduce()` to remove elements using `Array.splice()`.
-The `func` is invoked with three arguments (`value, index, array`).
+Finds elements that satisfy a predicate and returns them, at the same time modifying an original array.
+In other words, it works like [R#partition](http://ramdajs.com/docs/#partition), except that it mutates the original array and returns another partition instead of returning two.
```
Chalarangelo: The description should explain *how* the snippet works, not what it does.

snippets/array-remove.md
```diff
@@ -1,13 +1,22 @@
...
 ```js
-const remove = (arr, func) =>
-  Array.isArray(arr) ? arr.filter(func).reduce((acc, val) => {
-    arr.splice(arr.indexOf(val), 1); return acc.concat(val);
-    }, [])
-  : [];
-// remove([1, 2, 3, 4], n => n % 2 == 0) -> [2, 4]
+/**
+ * @param {E[]} arr
+ * @param {(E, number, E[]) => boolean}
```
Chalarangelo: Refer to the styleguide (CONTRIBUTING.md) for this. We do not write function signatures in snippets.

snippets/array-remove.md
```diff
@@ -1,13 +1,22 @@
...
+ * @returns mutated array
+ */
+const remove = (arr, func) => {
+  if (!Array.isArray(arr)) return [];
+  const matching = arr.filter(func);
+  const nonMatching = arr.filter((e, i) => !func(e, i, arr));
+  arr.splice(0, arr.length - 1, nonMatching);
+  return matching;
+}
```
Chalarangelo: Refer to the styleguide (CONTRIBUTING.md) for this.
- You should always prefer the ternary operator instead of `if else` statements.
- Avoid variables and try to chain calls as much as possible.

snippets/array-remove.md
kirilloid: But then `remove` might be a little misleading. It should be something like `withdraw`, I think.

snippets/array-remove.md
kirilloid: The whole expression will look then like minified and then de-minified code: heavy use of comma operator, etc. Is it still OK?

snippets/array-remove.md
Chalarangelo: I absolutely agree with you that snippet names alone are not enough to explain what they do. I am already looking into added descriptions for what snippets do retroactively. For now, follow the existing guidelines and we should be able to work out the other details later down the line.

snippets/array-remove.md
Chalarangelo: It is ok, the linter will take care of formatting to make it reasonably readable. Lots of snippets are written like that. If you can't make it a one-liner, try and shorten it as much as possible.

Chalarangelo:
I'm closing this PR as it has been dead for a couple of days and there have already been a bunch of changes made to the relevant file. Feel free to make a new one if you still think there are problems that need to be addressed.

PR #200, opened 2017-12-16
Update palindrome

iamsoorena:
use spread operator and change regex implementation to more conventional `/\s+/g` to match white-space.

### DIFF
--- a/snippets/check-for-palindrome.md
+++ b/snippets/check-for-palindrome.md
@@ -1,7 +1,7 @@
 ### Check for palindrome
 
 Convert string `toLowerCase()` and use `replace()` to remove non-alphanumeric characters from it.
-Then, `split('')` into individual characters, `reverse()`, `join('')` and compare to the original, unreversed string, after converting it `tolowerCase()`.
+Then spread(`[...str]`) into individual characters array, `reverse()`, `join('')` and compare to the original, unreversed string.
 
 ```js
 const palindrome = str => {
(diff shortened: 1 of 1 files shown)


iamsoorena:
I'll do it asap

wojtekmaj:
I'm not so sure about being "definitely" faster.

https://jsperf.com/30sec-palindrome

On Chrome, I get consistently lower performance. On Edge, spread operator is 4 (!) times slower.

obraz

Interestingly, spread operator *is* faster on Firefox 57.

iamsoorena:
@Chalarangelo what is your opinion about @wojtekmaj comment? I think he is right even if spread operators seem more like native code.

Chalarangelo:
@wojtekmaj Thanks for pointing that out, I am primarily a Firefox user, so tests were giving better results with spread operator for me.

atomiks:
I prefer using the nifty `timeTaken` function:

```js
// split
timeTaken(() => palindrome1('racecar'.repeat(1000000)))
// Chrome: timeTaken: 347.054931640625ms
// Firefox: timeTaken: 288.46ms

// ...spread
timeTaken(() => palindrome2('racecar'.repeat(1000000)))
// Chrome: timeTaken: 762.1259765625ms
// Firefox: timeTaken: 696.11ms
```

So using `split` is over 2x faster in both.

Chalarangelo:
@atomiks the spread operator is pretty slow then, although jsPerf indicated otherwise in most tests I have run. Let's stick with `split()` then as it is faster.

PR #204, opened 2017-12-17
Update Hexcode to RGB function with bitwise version

kutsan:
Hey me again. Wrote this this time; tell me what do you think.

We could have wrote this in one line, if we wouldn't have had to use `#` prefix for hexcode.

Speaking of which, I wrote also this version with alpha values. Should I create a new snippet for that (maybe `hexcode-to-RGBA.md`?) or should I add next to other one or should I update the current function with if statements to check whether or not value contains alpha segment?

```js
const hexToRgba = hex => {
  const h = parseInt(hex.slice(1), 16)
  return `rgba(${h >> 24}, ${(h & 0x00ff0000) >> 16}, ${(h & 0x0000ff00) >> 8}, ${(h & 0x000000ff)})`
}
// hexToRgba('#27ae60ff') -> 'rgba(39, 174, 96, 255)'
```

### DIFF
--- a/snippets/hexcode-to-RGB.md
+++ b/snippets/hexcode-to-RGB.md
@@ -1,8 +1,11 @@
 ### Hexcode to RGB
 
-Use `Array.slice()`, `Array.map()` and `match()` to convert a hexadecimal colorcode (prefixed with `#`) to a string with the RGB values.
+Use bitwise right-shift operator and mask bits with `&` (and) operator to convert a hexadecimal color code (prefixed with `#`) to a string with the RGB values.
 
 ```js
-const hexToRgb = hex => `rgb(${hex.slice(1).match(/.{2}/g).map(x => parseInt(x, 16)).join()})`
-// hexToRgb('#27ae60') -> 'rgb(39,174,96)'
+const hexToRgb = hex => {
+  const h = parseInt(hex.slice(1), 16);
+  return `rgb(${h >> 16}, ${(h & 0x00ff00) >> 8}, ${h & 0x0000ff})`;
+}
+// hexToRgb('#27ae60') -> 'rgb(39, 174, 96)'
 ```


Chalarangelo:
As far as RGBA is concerned, I think we should extend the existing snippet with a second argument (`alpha`) that defaults to `false` so that we return RGB usually, except if explicitly asked to return RGBA. What do you think about that?

kutsan:
I think that's too verbose to work with. We can do better like this:

```js
let hexToRgb = (hex) => {
	const h = parseInt(hex[0] === '#' ? hex.slice(1) : hex, 16)
	const withAlpha = (hex[0] === '#' ? hex.length === 9 : hex.length === 8)

	return {
		red: h >> (withAlpha ? 24 : 16),
		green: (h & (withAlpha ? 0x00ff0000 : 0x00ff00)) >> (withAlpha ? 16 : 8),
		blue: (h & (withAlpha ? 0x0000ff00 : 0x0000ff)) >> (withAlpha ? 8 : 0),
		...(withAlpha ? { alpha: h & 0x000000ff } : {})
	}
}
// hexToRgb('#57ce4a') -> { red: 87, green: 206, blue: 74 }
// hexToRgb('#57ce4a27') -> { red: 87, green: 206, blue: 74, alpha: 39 }
// hexToRgb('57ce4a') -> { red: 87, green: 206, blue: 74 }
// hexToRgb('57ce4a27') -> { red: 87, green: 206, blue: 74, alpha: 39 }
```

So, if you give hexcode with alpha then you get alpha back. [transform-object-rest-spread](https://babeljs.io/docs/plugins/transform-object-rest-spread/) is used so that plugin is needed. It works fine even without given `#` prefix.

And I think we shouldn't include `rgba(R, G, B)` string format like this. It's not so API friendly across your code. Returning a object is much more elegant.

kutsan:
By the way, this is the test I did.

https://jsperf.com/hex-to-rgb-proposal

Chalarangelo:
Hey, this one slipped through the cracks somehow, sorry about that. I prefer the object, too, but in all honesty the `rgb/rgba` string is better, because I know of a ton of uses cases involving CSS where you need a string exactly like that. Converting from an object wouldn't be the best thing in those cases, especially if you have to do it for every single value you convert. So, I say we stick with the string implementation and maybe add a similar method that returns an object (so we have both in different snippets), in case there are other use-cases for that. Yes, it's a bit of a duplication, but both serve their purpose. 

To reduce complexity and provided that there are browsers that do not support the `rgba` syntax yet, it would be preferrable to have a `hexToRGBA` method and keep the existing one as well, making sure we cover as many use-cases as possible. After all, we want the snippets to be useful to a wide range of people.

Thanks for all the time you spent on this and sorry for the delay, I can miss a few conversations every now and then!

kutsan:
No problem. I'm glad you get through.

Well, you are right `rgb/rgba` strings are indeed needed. So, what do you think about like this instead of creating a new snippet for `rgba`?

```js
hexToRGB('#27ae60ff') // 'rgba(39, 174, 96, 255)'
hexToRGB('#27ae60')   // 'rgb(39, 174, 96)'
hexToRGB('#fff')      // 'rgb(255, 255, 255)'

// without #
hexToRGB('27ae60ff') // 'rgba(39, 174, 96, 255)'
hexToRGB('27ae60')   // 'rgb(39, 174, 96)'
hexToRGB('fff')      // 'rgb(255, 255, 255)'
```

If there is an alpha value, then string should be in its `rgba` form.

I don't think we should create an other function that returns an object. Those are will be very similar. Although if you want to, you can tell me.

You are welcome. This project is a great idea, congrats for your stars and community effort.

Other than that, it's night time for me. I'll make your requests happen in tomorrow. :3

Chalarangelo:
Alright, including alpha when the input provides alpha seems like the best option. Good thinking!

kutsan:
Which one do you prefer? The only difference is in the return statement. Should I keep it as one long line or readable multiple lines? It works as I promised by the way. I tried to make it as short as possible, so it turns out like this.

```js
const hexToRGB = hex => {
	let alpha = false, h = hex.slice(hex.startsWith('#') ? 1 : 0)
	if (h.length === 3) h = [...h].map(x => x + x).join('')
	else if (h.length === 8) alpha = true
	h = parseInt(h, 16)
	return `rgb${alpha ? 'a' : ''}(${h >>> (alpha ? 24 : 16)}, ${(h & (alpha ? 0x00ff0000 : 0x00ff00)) >>> (alpha ? 16 : 8)}, ${(h & (alpha ? 0x0000ff00 : 0x0000ff)) >>> (alpha ? 8 : 0)}${(alpha ? ', ' + (h & 0x000000ff) : '')})`;
}
```

```js
const hexToRGB = hex => {
	let alpha = false, h = hex.slice(hex.startsWith('#') ? 1 : 0)
	if (h.length === 3) h = [...h].map(x => x + x).join('')
	else if (h.length === 8) alpha = true
	h = parseInt(h, 16)
	return 'rgb' + (alpha ? 'a' : '') + '('
		+ (h >>> (alpha ? 24 : 16)) + ', '
		+ ((h & (alpha ? 0x00ff0000 : 0x00ff00)) >>> (alpha ? 16 : 8)) + ', '
		+ ((h & (alpha ? 0x0000ff00 : 0x0000ff)) >>> (alpha ? 8 : 0))
		+ (alpha ? `, ${(h & 0x000000ff)}` : '') + ')';
}
```

Chalarangelo:
Although snippets should be short, the return statement of the first one is nearly impossible to read in one line, so I definitely prefer the second one.

kutsan:
Yeah I know. I don't really want to write that, that long but it can't be helped. At the end of the day, it provides a couple of functionality.

I created a PR https://github.com/Chalarangelo/30-seconds-of-code/pull/260.

PR #205, opened 2017-12-17
coalesce snippet

prodigic:
coalesce function. should probably be in the utility category


prodigic:
split into 2.

prodigic:
i've tidied up the factory function to use a function, makes it much cleaner IMO. 
so you can do stuff like...

```js
const firstTruthyVal = coalesceFactory(Boolean);
```

PR #209, opened 2017-12-17
Refactor union to not use Set and accept any number of arguments

AvraamMavridis:
`Set` cannot be transpiled to es5

This PR makes the union accept any number of arguments, and can be used in enviroments where there is no `Set`

### DIFF
--- a/snippets/array-union.md
+++ b/snippets/array-union.md
@@ -1,8 +1,18 @@
 ### Array union
 
-Create a `Set` with all values of `a` and `b` and convert to an array.
+Create a new array with unique values
 
 ```js
-const union = (a, b) => Array.from(new Set([...a, ...b]));
-// union([1,2,3], [4,3,2]) -> [1,2,3,4]
+const union = function() {
+  const arr = [...arguments];
+  
+  return arr.length ? arr
+    .reduce((val, sum) => sum = [...val, ...sum])
+    .filter((val, i, arr) => arr.indexOf(val) === i) : [];
+}
+
+
+console.log(union([], [1,2,3,4], [1000], [3,4,5])); // [1,2,3,4,1000,5]
+console.log(union(10, 10, 100)); // [10, 100]
+console.log(union()); // []
 ```


AvraamMavridis:
> Babel transpiles Set just fine

I don't think that this is the case, you can check on the [babel repl](https://babeljs.io/repl/#?babili=true&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAEAq0G8C-0D0q4CcwDsIAcBLEAUwBMAoCgNzE2gBdoBeaHEgd2gGUSGAKAJQBuNBgByAexyNseIqUo060CAE8AtgCYWPTQCNJIfgHIAZpMkmR0MdCkyGcgsXIUgA&debug=false&circleciRepo=&evaluate=true&lineWrap=true&presets=es2015%2Ces2016%2Ces2017%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3%2Ces2015-loose%2Cbabili&prettier=true&targets=&version=6.26.0)

e.g.

```
class T {} // Transpiled

var t = new Set(); // Non transpiled

var sym2 = Symbol('foo');  // Non transpiled
```

I think Set (and Map, WeakMap, WeakSet) is a feature the cannot be transpiled. To be able to iterate on `Set` it implements the `Symbol.iterator`, and `Symbols` are not transpilable either. 

Another thing (if that is important) is that the current implementation accepts only two arguments, so if you want a union of 3 arrays you have to do `union([1,2], union([3,4], [5,6]))` whereas with the provided PR you could do `union([1,2], [3,4], [5,6])`. A bit more clean IMO :)

skatcat31:
Considering that each JS engine in major use is capable of Set and most ES6 uses I'm not certain refactoring it to arrays is the right approach, especially since ES6 is considered "finished" and even supported to a level of species in all major browsers(and NodeJS).
____
Also if you want to do a union of many you could also do something using a variadic `unionAll` function that isn't too hard since you already have a `union`2 function

```js
const unionAll = (...sets) => sets.reduce(union);
```

Chalarangelo:
`Set` is transpilable s far as I can tell (testing the transpiled code in an older browser seemed to work fine) and a variadic function for `union` is not a necessity as of now. It might be a worthwhile addition in the future, but it should be a different PR later down the line.

Issue #206, opened 2017-12-17
Naming


fejes713:
This is potentially a serious problem.
While talking with @Chalarangelo, we agreed that naming should be **consistent**. The main idea behind this issue is to fix naming across the **whole** project. Let me explain the idea:

- Most of the functions have the different name than their .md file has. For example `initial()` is located inside the `initial-of-list.md`. We need to use the **same name** for **both** function name and snippet.md name.

- Most of the functions have names that don't precisely describe what they do and that should be changed. In my opinion, loadash has good naming for their functions. You can exactly tell what each function does just by looking at the name of that function. 

Before changing the naming of snippet files, **change the function names first** to the name that **describes more precisely** what it does. Later on, someone will PR all snippet names that need change.

Chalarangelo:
Very well explained, @fejes713. I also think we should extend descriptions a bit to explain in one line what the snippets do, then how they work (which we already have), so that it's easier to find everything in the long run.

As far as function names go, I need to start looking at the snippets when I'm done with PRs and list the ones that seem a bit off. I'll get back on this soon.

PR #218, opened 2017-12-17
Fix naming

fejes713:
I have started fixing the naming issue. First of all, I have changed all snippet file names to their corresponding names of functions. I've changed all the files that didn't match their name with the name of the function they're storing. 

**There is still lots of work to be done.** While going through the names of function, most of them could be improved. I didn't want to touch them since we first need to discuss our ideas in #206 

I've kept everything **consistent**. Used _camelCase_ style in code and _kebab-case_ in snippet file names.


Chalarangelo:
I generally don't write in full caps, but **THIS IS AMAZING**. Thanks for the time and effort to do this. We now only need to update descriptions and we should be good to go! I'll get started on the descriptions right away!

PR #220, opened 2017-12-17
Array Sum

grantmx:
Use `Array.reduce()` to sum an array of integers.


Chalarangelo:
Thanks for the effort, but this already exists under [arraySum](https://github.com/Chalarangelo/30-seconds-of-code#arraysum).

PR #225, opened 2017-12-17
Fix hexToRGB() issue with 3-digit colorcodes

### LINKED ISSUE #216: hexToRgb method optimization
**When the color value is 3 digits, incorrect results**
```js
const hexToRgb = hex => {
  const h = parseInt(hex.slice(1), 16);
  return `rgb(${h >> 16}, ${(h & 0x00ff00) >> 8}, ${h & 0x0000ff})`;
}
// hexToRgb('#27ae60') -> 'rgb(39, 174, 96)'
// hexToRgb('#acd') -> 'rgb(0, 10, 205)'
```
`#acd` == `#aaccdd`:
*wrong*: `rgb(0, 10, 205)`
*correct*: `rgb(170, 208, 221)`

---

```js
const getRgb = hex => {
    let _hex = hex.slice(1)
    let h
    if(_hex.length === 3) {
        for (let i of _hex) {
            h += i.repeat(2)
        }
        h = parseInt(h.slice(9), 16)
    } else {
        h = parseInt(_hex, 16)
    }
    return `rgb(${h >> 16}, ${(h & 0x00ff00) >> 8}, ${h & 0x0000ff})`
}
// getRgb('#acd') -> 'rgb(170, 208, 221)'
```

Pl4gue:
Fixes #216

### DIFF
--- a/snippets/hexToRGB.md
+++ b/snippets/hexToRGB.md
@@ -2,12 +2,16 @@
 
 Converts a colorcode to a `rgb()` string.
 
-Use bitwise right-shift operator and mask bits with `&` (and) operator to convert a hexadecimal color code (prefixed with `#`) to a string with the RGB values.
+Use bitwise right-shift operator and mask bits with `&` (and) operator to convert a hexadecimal color code (prefixed with `#`) to a string with the RGB values. In case it's a 3-digit-colorcode, do the same with the 6-digit-colorcode extended by the extendHex() function (ref. `extendHex` snippet)
 
 ```js
-const hexToRGB = hex => {
-  const h = parseInt(hex.slice(1), 16);
-  return `rgb(${h >> 16}, ${(h & 0x00ff00) >> 8}, ${h & 0x0000ff})`;
-}
+const hexToRgb = hex => {
+  const extendHex = shortHex => 
+    '#' + shortHex.slice(shortHex.startsWith('#') ? 1 : 0).split('').map(x => x+x).join('');
+  return hex.slice(1).length==3 ? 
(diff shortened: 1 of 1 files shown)


kutsan:
You could have use a variable instead of calling the same functions over and over again. Although it will make it one more line longer but still will be short in character-wise.

Chalarangelo:
@kutsan You can sen a PR if you hae an improvement ready! 😉

kutsan:
Can we talk about this one for `extendHex()`?

```js
const extendHex = hex => {
	const h = parseInt(hex.startsWith('#') ? hex.slice(1) : hex, 16)
	return (((((h & 0xf00) + ((h & 0xf00) >> 4)) >> 4) << 16)
		+ (((h & 0x0f0) + ((h & 0xf0) >> 4)) << 8)
		+ ((h & 0x00f) + ((h & 0x00f) << 4)))
		.toString(16)
		.padStart(6, '0')
}
// extendHex('#acd') -> aaccdd
// extendHex('acd') -> aaccdd
```

It's fast but I am not sure if it's long? We could make one line though 😝.

Chalarangelo:
If explained properly, this snippet is reasonably easy to understand as it's mostly bit shifting, so I see no reason why not use it as-is. If it's fast, even better.

Generally speaking, if the code is really complicated but fast, we prefer something that strikes a balance between readability and speed. In this case, due to the nature of the snippet and the operations in it, it doesn't seem too hard to grasp the core idea, bit shifting, and then there's not much else to really understand apart from that. Therefore, I think it's a good fit.

kutsan:
Even tough I think current `extendHex()` function is more proper for this repository. I know we are shifting just bit left and right but not I know not every people would be happy to see my version over current one which is more understandable. So, that's why I won't make a PR about `extendHex()` function. But there is that, you can copy if you want to see it.

I'll make this function more reliable once get your thoughts about https://github.com/Chalarangelo/30-seconds-of-code/pull/204.

PR #229, opened 2017-12-17
Fix all test cases

fejes713:
We have changed all the names of functions and their corresponding snippet files but forgot to change test cases. 

Example:
`truncateString()` is defined and then in test case we use `truncate()`. The same mistake happened in every 2nd file


fejes713:
Also for the naming. I've seen we've moved to camelCase.md naming for our snippets. That's totally fine with me. However `RGBToHex` isn't camelCase. Are snippets that include `RGB`,`JSON`,`UUID` exceptions?

Chalarangelo:
Thanks a ton for fixing the test cases! For now, we are going to keep `RGB`, `UUID`, `URL` and `JSON` capitalzed (JSON is capitalized in the native JS methods as well), but I would like to peer review this decision to figure out what should classify as an exception to the camelcasing rules for consistency.

fejes713:
No problem! Glad to help.

PR #231, opened 2017-12-17
reverseString: Replace spread operator with split() for compatibility

wojtekmaj:


iamsoorena:
@wojtekmaj  I think It needs to be decided once and for all
`spread VS. split('')`

Chalarangelo:
@wojtekmaj, @iamsoorena Honestly, we have changed half the snippets that use either back and forth like 10 times. I need a definitive answer, so we can add it to the contribution guidelines. I am primarily a Firefox user and spread is definitely faster on Firefox. It's also easier to read, however if split is significantly faster, we should opt to use that. If the difference is like very minor, we should only use the optimal one of the two if we expect huge inputs and operations are costly (for example anagrams). Can one of you test this consistently and give us some results for all browsers? I would be glad to finalize the guideline on this.

skatcat31:
I think we can all still agree that the rest operator is still fine to use in most(99.99%) cases though

snippets/reverseString.md
```diff
@@ -2,10 +2,10 @@
 
 Reverses a string.
 
-Use array destructuring and `Array.reverse()` to reverse the order of the characters in the string.
+Use `split('')` and `Array.reverse()` to reverse the order of the characters in the string.
 Combine characters to get a string using `join('')`.
 
 ```js
-const reverseString = str => [...str].reverse().join('');
+const reverseString = str => str.split('').reverse().join('');
```
skatcat31: Maybe we should just avoid `split` vs `spread` and make it obvious what we want to do with `Array.from`
```js
Array.from(str).reverse.join('')
```

skatcat31:
Right maybe I'm not used to github just yet, but for future reference how do you just do a comment on code and not a review? Did I miss a button somewhere?

Chalarangelo:
@skatcat31 there's a button that says `Add single comment` instead of `Start a review`. But it's fine either way, don't worry about it.

Chalarangelo:
From the test results presented in #200, `split()` is faster, so we should stick with that as much as possible, especially if we anticipate the input and processing of a function to be significant.

Issue #100, opened 2017-12-14
Elect Lodash Methods To Demystify ~ Log


kingdavidmartins:
Hey @elderhsouza Thanks again for the awesome log suggestion 👋  & Hey All 😄 

## Library

**Category**

- [x] link to snippet

**Legend**
- [x] link to snippet ~ Means no PR has been ref or Submitted
- [x] link to snippet | PR ref ~ Means PR is opened and ref
- [x] <del> link to snippet | PR ref </del> ~ Means PR was closed and not merged
- [x] link to snippet | `file-name.md` ~ Means PR has been merged & snippet has been implemented 

**Note** *If a snippet does not have a PR attached. That means a PR wasn't created or referenced*

Here's a list of the following snippets so far with their respective PR. Remember if there isn't a PR ref attached to a snippet it means no one has gotten around to it yet 💔 Which also means there's work to be done & contributions to be made 💻 💯  

Don't be shy and give it a try. Fork, Clone, Contribute, & Submit a PR 🎊 

Also shout out to @Chalarangelo for starting such an awesome project 🎉 

### *REMEMBER*
Follow the guidelines in CONTRIBUTING.md
No copy-pasting. Let's just try to have a fun and educative experience

## Lodash

**Function**

- [x]  https://lodash.com/docs/4.17.4#throttle | `throttle.md`
- [x]  https://lodash.com/docs/4.17.4#debounce | `debounce.md`
- [x]  https://lodash.com/docs/4.17.4#rearg | `rearg.md`
- [x]  https://lodash.com/docs/4.17.4#overArgs | `overArgs.md`
- [x]  https://lodash.com/docs/4.17.4#rest | `collectInto.md`
- [x]  https://lodash.com/docs/4.17.4#partial | `partial.md`
- [x]  https://lodash.com/docs/4.17.4#partialRight | `partialRight.md`
- [x]  https://lodash.com/docs/4.17.4#delay | `delay.md`
- [x]  https://lodash.com/docs/4.17.4#bindKey | `bindKey.md`
- [x]  https://lodash.com/docs/4.17.4#bind | `bind.md`
- [x]  https://lodash.com/docs/4.17.4#ary | `ary.md`
- [x]  https://lodash.com/docs/4.17.4#unary | `unary.md`
- [x]  https://lodash.com/docs/4.17.4#negate | `negate.md`
- [x]  https://lodash.com/docs/4.17.4#once | #455
[...]

PR #241, opened 2017-12-18
Array - Drop right

taltmann42:
#100 - https://lodash.com/docs/4.17.4#dropRight


skatcat31:
https://github.com/Chalarangelo/30-seconds-of-code#takeright

skatcat31:
Sorry, misread code. Seems good. I'm just getting confused because of the name.

taltmann42:
Well, I was just using the same name lodash does, as it's part of #100 - and I checked and the lodash function also doesn't mutate the original array. And the short explanation gives away that it returns a new array.

However, it surely is confusing when having snippets start with `drop` and once being pure and once not. Maybe there should be some flags in array function headlines determining if they mutate the original array or not?

### Array dropRight [pure]

and

### Array dropRight [mutating]

PR #244, opened 2017-12-18
add pullAll

kingdavidmartins:
similar to lodash pullAll method

### pullAll

This method is like pull except that it accepts an array of values to filter out before Mutating and pulling all the values from the original array.

Use `Array.filter()` and `Array.includes()` to pull out the values that are not needed.
Use `Array.length = 0` to mutate the passed in array by resetting it's length to zero and `Array.push()` to re-populate it with only the pulled values.

```js
const pullAll = (arr, pullArr) => {
  let pulled = arr.filter((v, i) => !pullArr.includes(v));
  arr.length = 0;
  pulled.forEach(v => arr.push(v));
}

// let myArray = ['a', 'b', 'c', 'a', 'b', 'c'];
// pullAll(myArray, ['a', 'c']);
// console.log(myArray) -> [ 'b', 'b' ]
```

https://lodash.com/docs/4.17.4#pullAll

### DIFF
--- a/snippets/pullAll.md
+++ b/snippets/pullAll.md
@@ -1,0 +1,18 @@
+### pullAll
+
+This method is like pull except that it accepts an array of values to filter out before Mutating and pulling all the values from the original array.
+
+Use `Array.filter()` and `Array.includes()` to pull out the values that are not needed.
+Use `Array.length = 0` to mutate the passed in array by resetting it's length to zero and `Array.push()` to re-populate it with only the pulled values.
+
+```js
+const pullAll = (arr, pullArr) => {
+  let pulled = arr.filter((v, i) => !pullArr.includes(v));
+  arr.length = 0;
+  pulled.forEach(v => arr.push(v));
+}
+
+// let myArray = ['a', 'b', 'c', 'a', 'b', 'c'];
+// pullAll(myArray, ['a', 'c']);
+// console.log(myArray) -> [ 'b', 'b' ]
+```
--- a/snippets/pullAll.md
+++ b/snippets/pullAll.md
@@ -1,6 +1,6 @@
 ### pullAll
 
-This method is like pull except that it accepts an array of values to filter out before Mutating and pulling all the values from the original array.
+Mutates the original array to filter out the values specified (accepts an array of values).
 
 Use `Array.filter()` and `Array.includes()` to pull out the values that are not needed.
 Use `Array.length = 0` to mutate the passed in array by resetting it's length to zero and `Array.push()` to re-populate it with only the pulled values.
@@ -8,10 +8,8 @@ Use `Array.length = 0` to mutate the passed in array by resetting it's length to
 ```js
 const pullAll = (arr, pullArr) => {
   let pulled = arr.filter((v, i) => !pullArr.includes(v));
-  arr.length = 0;
-  pulled.forEach(v => arr.push(v));
+  arr.length = 0; pulled.forEach(v => arr.push(v));
 }
-
 // let myArray = ['a', 'b', 'c', 'a', 'b', 'c'];
 // pullAll(myArray, ['a', 'c']);
 // console.log(myArray) -> [ 'b', 'b' ]
--- a/tag_database
+++ b/tag_database
@@ -71,6 +71,7 @@ pipe:function
 powerset:math
 promisify:function
 pull:array
+pullAll:array
 randomIntegerInRange:math
 randomNumberInRange:math
 readFileLines:node


skatcat31:
What's a use case for this? As a mutating function that uses filter to modify the original array, why would I not just use filter?

kingdavidmartins:
Hey @skatcat31 👋  

Thanks for the feedback.
> What's a use case for this versus pull?
Couldn't I also achieve the same with the following code?
```js
const pullAll = (arr, pullArr) => pull(a,...pullArr)
```

So yea you def can. But the reason I think using a implemented function from 30 seconds of code in the example is a potential pitfall is that since the project isn't necessarily a lib or a module like lodash & etc. A user can't just do something like

```js
const sec = require('30-sec-of-code');
const pullAll = (arr, pullArr) => sec.pull(a,...pullArr);
``` 

If that was the case then I would understand implementing the function in that manner. However my understanding about 30 seconds of code is for users to be able to want to achieve/implement a particular thing and finding out exactly how it is done in 30 secs or less or fairly quickly

I think the only thing that is fair to assume is that the users will and should be able to understand/learn Native JavaScript ES^ methods like 
```js
.filter();
```
or
```js 
.from(); 
```
which already has sufficient enough of documentation out there.

by using the `pull` method we assume the user already knows about the method and already has it implemented in their environment. However if they don't it will def take them sometime to reference pull and see where and why we are using it in the said implementation or if they copy and paste the said snippet in their environment it would just break.

> For that matter I'm not certain pull should be in here either... but that's just personal dislike of mutating functions since I work in a large multi developer environment at work.

Sure there's always going to dev's in the developer community who prefer `Object Oriented` vs `Functional` or `Mutating State` vs `Immutable State` but I think at that point that is more personal preference. Lodash implement features based on demand/request and believe It def brings value depending on the use case. 

Personally I try not to mutate state when I can or when I am working on certain projects. But I've also found instances where mutating state and being able to keep track of what was mutated using lodash `pullAt` in conjunction with `pull` very helpful in those instances

skatcat31:
What I meant is why is this being submitted when pull is so extremely similar but variadic instead, and I'm curious to why you're implementing this function that is almost identical to pull right next to it.

kingdavidmartins:
> What I meant is why is this being submitted when pull is so extremely similar

@skatcat31 I honestly agree, it def is extremely similar and doesn't really take much more effort to implement when comparing it to pull. All you need to do is destructuring `...`

However lodash implemented it in their project and as such implementing it in the current project was just knocking 1 more method/snippet of the Lodash Demystified Backlog #100 

That being said I think it really comes down to if we as a community believes if this falls under the <strong>*Natively Implemented*</strong> category  or not.

Personally even though I agree that it might seem pretty intuitive to take the snippet/method `pull()` and implement the snippet/method `pullAll()`

I def think that what's intuitive is subjective and for complete beginners it may not be so

iamsoorena:
I agree that [this](https://github.com/Chalarangelo/30-seconds-of-code/blob/master/README.md#pull) is very similiar to `pull*` functions you PRed.
and I don't think we want to make a searchable index for programmers to find the way to implement functions of lodash or whatever.
I think(or at least I hopethat) we want to show programmers a beautiful peace of code that is unserstandable in 30 seconds or less.

kingdavidmartins:
Hey @iamsoorena 👋 

Thanks for the feedback

> I don't think we want to make a searchable index for programmers to find the way to implement functions of lodash or whatever.

Yea, honestly I'm not saying that is the goal or case. Was just using lodash as an example. The truth is a lot of organizations/companies/developers use lodash snippets/methods in production/projects. However lodash day by day source code is getting more convoluted. I 'm just saying that a lot of those same users can gain a lot of value from the project if it's possible to get an over view of the code they have been using i.e lodash.

> I agree that this is very similiar to pull* functions you PRed.

To be honest it gets tricky. What is considered naive & intuitive and not?

For Example to date 30 Seconds of Code host the following code

<strong>*arrayMax.md*</strong>
```js
const arrayMax = arr => Math.max(...arr);
// arrayMax([10, 1, 5]) -> 10
```

<strong>*arrayMin.md*</strong>
```js
const arrayMin = arr => Math.min(...arr);
// arrayMin([10, 1, 5]) -> 1
```

In these two examples as you can see just 3 characters `max` & `min` separate one from the other. Yet added in.

Also there's a couple of methods that are variations/extensions of itself that are implemented in 30 sec of code that aren't too far from the other. What makes this code snippet different from the others?

iamsoorena:
I think and I hope that we will somehow merge these functions (`min` and `max`) too.

kingdavidmartins:
I think what it also comes down to is what we <strong>*define as the level of assumed proficiency*</strong>

Because merging `min` & `max` honestly isn't something that's hard. It's just that we can end up trading "simplicity" for "complexity" which can potentially increase the barrier to entry when it comes to contributing/understanding.

Chalarangelo:
Sorry, I merged without reading the conversation. My bad!
We should remove `pullAll` and update `pull` to work with any amount of values *or* an array, so we have a more general-use-case function without cluttering the list.

kingdavidmartins:
@Chalarangelo  No worries I will get on it.

PR #248, opened 2017-12-19
Add pullAtValue

kingdavidmartins:
similar to lodash pullAt method with a slight variation explained in #245 

### pullAtValue

This method is like pull except that it accepts an array of values to filter out before Mutating and pulling all the values from the original array. Which then returns an array of removed elements.

Use `Array.filter()` and `Array.includes()` to pull out the values that are not needed.
Use `Array.length = 0` to mutate the passed in array by resetting it's length to zero and `Array.push()` to re-populate it with only the pulled values.
Use `Array.push()` to keep track of pulled values 

```js
const pullAtValue = (arr, pullArr) => {
  let removed = [];
  let pushToRemove = arr.forEach((v, i) => pullArr.includes(v) ? removed.push(v) : v)
  let mutateTo = arr.filter((v, i) => !pullArr.includes(v))
  arr.length = 0;
  mutateTo.forEach(v => arr.push(v));
  return removed;
}

let myArray = ['a', 'b', 'c', 'd'];
let pulled = pullAtValue(myArray, ['b', 'd']);

// let myArray = ['a', 'b', 'c', 'd'];
// let pulled = pullAtValue(myArray, ['b', 'd']);

// console.log(myArray); -> [ 'a', 'c' ]
// console.log(pulled); -> [ 'b', 'd' ]
```

### DIFF
--- a/snippets/pullAtValue.md
+++ b/snippets/pullAtValue.md
@@ -1,0 +1,27 @@
+### pullAtValue
+
+This method is like pull except that it accepts an array of values to filter out before Mutating and pulling all the values from the original array. Which then returns an array of removed elements.
+
+Use `Array.filter()` and `Array.includes()` to pull out the values that are not needed.
+Use `Array.length = 0` to mutate the passed in array by resetting it's length to zero and `Array.push()` to re-populate it with only the pulled values.
+Use `Array.push()` to keep track of pulled values 
+
+```js
+const pullAtValue = (arr, pullArr) => {
+  let removed = [];
+  let pushToRemove = arr.forEach((v, i) => pullArr.includes(v) ? removed.push(v) : v)
+  let mutateTo = arr.filter((v, i) => !pullArr.includes(v))
+  arr.length = 0;
+  mutateTo.forEach(v => arr.push(v));
+  return removed;
+}
+
+let myArray = ['a', 'b', 'c', 'd'];
+let pulled = pullAtValue(myArray, ['b', 'd']);
+
+// let myArray = ['a', 'b', 'c', 'd'];
+// let pulled = pullAtValue(myArray, ['b', 'd']);
+
+// console.log(myArray); -> [ 'a', 'c' ]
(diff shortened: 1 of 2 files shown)


kingdavidmartins:
`pull` or `pullAll` just pulls the passed in array to your specifications. Which just mutates the array

```js
const pull = (arr, ...args) => {
  let pulled = arr.filter((v, i) => !args.includes(v));
  arr.length = 0; pulled.forEach(v => arr.push(v));
};
// let myArray = ['a', 'b', 'c', 'a', 'b', 'c'];
// pull(myArray, 'a', 'c');
// console.log(myArray) -> [ 'b', 'b' ]
```

Where as `pullAtValue` acts as `pull` with the difference being that `pullAtValue` returns the values that was removed from the original array

```js
const pullAtValue = (arr, pullArr) => {
  let removed = [];
  let pushToRemove = arr.forEach((v, i) => pullArr.includes(v) ? removed.push(v) : v)
  let mutateTo = arr.filter((v, i) => !pullArr.includes(v))
  arr.length = 0;
  mutateTo.forEach(v => arr.push(v));
  return removed;
}

let myArray = ['a', 'b', 'c', 'd'];
let pulled = pullAtValue(myArray, ['b', 'd']);

// let myArray = ['a', 'b', 'c', 'd'];
// let pulled = pullAtValue(myArray, ['b', 'd']);

// console.log(myArray); -> [ 'a', 'c' ]
// console.log(pulled); -> [ 'b', 'd' ]
```

#### Use Case

An example of it's usefulness is when implementing a game like hangman 
If the empty spaces are populated with letter variables and lets say for an example the user inputs 
`'o'` and the expected word is `30 seconds of code` we can use `pullAtValue(word, ['o', 's'])` to not only mutate the array/ui to show the guessed letters, but also use the return arr as a way to keep track of what was removed which can also be used to show the user their history. Rather then using a method like difference in tangent with pull

*hope this answers your question*

Chalarangelo:
Ah, I see, makes sense. Thanks for the clarification!

PR #263, opened 2017-12-19
Snippet to convert string to array of words

arjunmahishi:
## Description

Added new snippet

## Types of changes

## Checklist:

### DIFF
--- a/snippets/stringToArrayOfWords.md
+++ b/snippets/stringToArrayOfWords.md
@@ -2,7 +2,7 @@
 
 Converts a given string into an array of words
 
-First create an array of dirt that you want to remove from the string. Then replace each dirt with an empty string. Use `String.split(' ')` to conver the string to an array with the space as a delimiter.
+First create an array of dirt that you want to remove from the string. Then replace each dirt with an empty string. Use `String.split(/[\W_]+/)` to convert the string to an array with the use of regex.
 
 ```js
 const stringToArrayOfWords = str => {
--- a/snippets/stringToArrayOfWords.md
+++ b/snippets/stringToArrayOfWords.md
@@ -2,7 +2,7 @@
 
 Converts a given string into an array of words
 
-Use `String.split(/[\W_]+/)` to convert the string to an array with the use of regex.
+Use `String.split` to convert the string to an array.
 
 ```js
 const stringToArrayOfWords = (str, pattern = /[\W_]+/) => str.split(pattern).filter(Boolean)
--- a/snippets/stringToArrayOfWords.md
+++ b/snippets/stringToArrayOfWords.md
@@ -5,7 +5,7 @@ Converts a given string into an array of words
 First create an array of dirt that you want to remove from the string. Then replace each dirt with an empty string. Use `String.split(' ')` to conver the string to an array with the space as a delimiter.
 
 ```js
-const stringToArray = str =>{
+const stringToArrayOfWords = str =>{
 	[".", ",", "?", "!", "& ", "(", ")", "[", "]"].map(d => str = str.replace(d, ""));
 	return str.split(" ");
 }
--- a/snippets/stringToArrayOfWords.md
+++ b/snippets/stringToArrayOfWords.md
@@ -2,7 +2,7 @@
 
 Converts a given string into an array of words
 
-Use `String.split` to convert the string to an array.
+Use String.split with a supplied pattern(defaults to non alpha as a regex) to convert to an Array of Strings and then filter to remove any empty strings.
 
 ```js
 const stringToArrayOfWords = (str, pattern = /[\W_]+/) => str.split(pattern).filter(Boolean)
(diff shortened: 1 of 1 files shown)


arjunmahishi:
Picked up from #100

atomiks:
Hmm this seems to be tricky because of all the different possible cases. For example, it fails here:

`stringToArrayOfWords(' I love Java-Script, & coffee !!!! ')`

This can be done more efficiently (although it leaves empty chars at the start and end for the above string - can someone better at regex fix that?):

```js
const stringToArrayOfWords = (str, pattern) => str.split(pattern || /[\W_]+/);
```

For non-English characters they can use a custom pattern.

arjunmahishi:
@atomiks  Updated the function based on your suggestion. Check it out.

snippets/stringToArrayOfWords.md
```diff
@@ -0,0 +1,15 @@
+### stringToArrayOfWords
+
+Converts a given string into an array of words
+
+First create an array of dirt that you want to remove from the string. Then replace each dirt with an empty string. Use `String.split(/[\W_]+/)` to convert the string to an array with the use of regex.
+
+```js
+const stringToArrayOfWords = str => {
+	[".", ",", "?", "!", "& ", "(", ")", "[", "]"].map(d => str = str.split(d).join(""));
```
skatcat31: This can be done as a reduce
```js
const stringToArrayOfWords = str => [".", ",", "?", "!", "& ", "(", ")", "[", "]"].reduce( ( s, d ) => s.split( d ).join(""), str).split(/[\W_]+/)
```

skatcat31:
@atomiks why didn't you use a default?
```js
const stringToArrayOfWords = (str, pattern = //[\W_]+//) => str.split(pattern);
```

snippets/stringToArrayOfWords.md
skatcat31: Also why not just combine the array pattern into a regex?

arjunmahishi: Regex is a pain in some parts of the body :P

arjunmahishi: Writing all the code in one line might not be clear to most of the beginners in javaScript (including me). Certainly cannot be understood in 30 sec :P

skatcat31: I made a comment that fixes @atomiks approach for a much easier to understand and faster one liner than the loop version.

By comparison on performance, the loop version is orders of magnitude slower that the non alphabet regex of `/[\W_]/`

arjunmahishi: Working well! and easy to understand too! :)

snippets/stringToArrayOfWords.md
```diff
@@ -0,0 +1,12 @@
+### stringToArrayOfWords
+
+Converts a given string into an array of words
+
+Use `String.split(/[\W_]+/)` to convert the string to an array with the use of regex.
+
```
skatcat31: You should add to the description about why you're filtering the result of the split

snippets/stringToArrayOfWords.md
```diff
@@ -0,0 +1,12 @@
+### stringToArrayOfWords
+
+Converts a given string into an array of words
+
+Use `String.split(/[\W_]+/)` to convert the string to an array with the use of regex.
```
skatcat31: The pattern is no able to be user supplied, so `String.split` would suffice here

skatcat31:
Description still needs a bit more since it's missing the intermediate, and the filter explination
> Use `String.split` with a supplied pattern(defaults to non alpha as a regex) to convert to an Array of Strings and then filter to remove any empty Strings

atomiks:
I wanted to ONLY use split if possible, to avoid going through the entire array with `filter`. I'm sure you can do it with some regex, I'm not that good with it.

atomiks:
Also note that regex is splitting hyphenated words... you'd probably need this:

```js
const stringToArrayOfWords = (str, pattern = /[^a-zA-Z-]+/) => str.split(pattern).filter(Boolean)
```

atomiks:
@arjunmahishi we need to fix the regex so it doesn't use lone hyphens though 😓 

Can someone else good at regex fix it?

arjunmahishi:
True.. But we should probably leave that regex to be used as a custom one. No need to put it as default. What do you say?

Chalarangelo:
I like the current implementation and it works well enough for most test-cases I tried. It should be fine and providing a default value for the regex makes the snippet a lot more robust and allows people to tokenize strings any way they want. I'll merge as-is and if someone wants to make any improvements, please feel free to do so.

arjunmahishi:
@kingdavidmartins This was picked up from #100

kingdavidmartins:
@arjunmahishi Thanks!! Do you remember the exact name of the method? 
### *Example*
https://lodash.com/docs/4.17.4#methodName

arjunmahishi:
@kingdavidmartins it's https://lodash.com/docs/4.17.4#words

kingdavidmartins:
@arjunmahishi Thanks Loads!!! Keep up the amazing work 💯 ⚡️

PR #264, opened 2017-12-19
Snippet for checking armstrong number

arjunmahishi:
## Description

Added new snippet

## Types of changes

## Checklist:

### DIFF
--- a/snippets/isArmstrongNumber.md
+++ b/snippets/isArmstrongNumber.md
@@ -5,8 +5,8 @@ Checks if the given number is an armstrong number or not.
 Convert the given number into array of digits. Use `Math.pow()` to get the appropriate power for each digit and sum them up. If the sum is equal to the number itself, return `true` otherwise `false`.
 
 ```js
-const isArmstrongNumber = digits => ( arr => arr.reduce( ( a, d ) => a + Math.pow( parseInt( d ), arr.length ), 0 ) == digits ? true : false )( ( digits+'' ).split( '' ) )
-
+const isArmstrongNumber = digits => 
+  ( arr => arr.reduce( ( a, d ) => a + Math.pow( parseInt( d ), arr.length ), 0 ) == digits ? true : false )( ( digits+'' ).split( '' ) );
 // isArmstrongNumber(1634) -> true
 // isArmstrongNumber(371) -> true
 // isArmstrongNumber(56) -> false
--- a/snippets/isArmstrongNumber.md
+++ b/snippets/isArmstrongNumber.md
@@ -1,0 +1,17 @@
+### isArmstrongNumber
+
+Checks if the given number is an armstrong number or not.
+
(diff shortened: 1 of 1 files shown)


arjunmahishi:
Done :)

snippets/isArmstrongNumber.md
```diff
@@ -0,0 +1,17 @@
+### isArmstrongNumber
+
+Checks if the given number is an armstrong number or not.
+
+Convert the given number into array of digits. Use `Math.pow()` to get the appropriate power for each digit and sum them up. If the sum is equal to the number itself, return `true` otherwise `false`.
+
+```js
+const isArmstrongNumber = digits => {
+	let total = 0, arr = (digits+"").split("");
+	arr.map(d => total += Math.pow(parseInt(d), arr.length))
```
skatcat31: This could be a reduce instead, saving the deceleration of total, and making the comparison
```js
return arr.reduce( ( a, d ) => a + Math.pow( parseInt( d ), arr.length ) ) == digits ? true : false
```

skatcat31: Of course then you could also use the internal closure trick to declare the arr too, but that gets pretty obtuse for most people. While fun, it often leads to a lot of parenthesis

skatcat31:
for fun and future reference, but like I said the amount of parenthesis and the order of passing makes it really confusing
```js
const isArmstrongNumber = digits => ( arr => arr.reduce( ( a, d ) => a + Math.pow( parseInt( d ), arr.length ), 0 ) == digits ? true : false )( ( digits+'' ).split( '' ) )
```

snippets/isArmstrongNumber.md
arjunmahishi: need to do `parseInt(a)` before adding...

arjunmahishi:
@skatcat31 check it out now...

snippets/isArmstrongNumber.md
```diff
@@ -0,0 +1,16 @@
+### isArmstrongNumber
+
+Checks if the given number is an armstrong number or not.
+
+Convert the given number into array of digits. Use `Math.pow()` to get the appropriate power for each digit and sum them up. If the sum is equal to the number itself, return `true` otherwise `false`.
+
+```js
+const isArmstrongNumber = num => {
+	let arr = (num+"").split("");
+	return arr.reduce((a, d) => parseInt(a) + Math.pow(parseInt( d ), arr.length), 0) == num ? true : false
```
skatcat31: why are you doing `parseInt(a)`? a is always guaranteed to be an integer

skatcat31:
Other than doing a parse where not needed, it looks great!

arjunmahishi:
done :)

skatcat31:
toss up between the two line version and the closure version, but a great add nonetheless!

PR #280, opened 2017-12-20
snippet to implement toString

arjunmahishi:
## Description

Added new snippet

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/toString.md
+++ b/snippets/toString.md
@@ -1,0 +1,13 @@
+### toString
+
+Converts the given `value` to a String
+
+Works exactly like the native `.toString()` but it also cover two exceptional cases (-0 and null)
+
+```js
+const toString = value => value == -0 ? "-0" : (value+"").replace("null", "");
+
+//toString(null) -> ''
+//toString(-0) -> '-0'
+//toString([1, 2, 3]) -> '1,2,3' 
+```


arjunmahishi:
Not sure if this is a legit PR because `toString()` is natively available. But it covers two exceptional cases (-0 and null) as seen [here](https://lodash.com/docs/4.17.4#toString)

arjunmahishi:
@skatcat31 I don't get it. How is this related to `toString()` ?

skatcat31:
```js
toString(0) // '-0'
toString(false) // '-0'
```

arjunmahishi:
Ohh right!! didn't realize that! it's midnight here :P

arjunmahishi:
How do I fix this?

skatcat31:
think of the edge cases surrounding -0 specifically in terms of math.

```js
if ( x === 0 && something specific to -0 ) '-0'
```

iamsoorena:
I have an opinion, :D I'm not a fan of this approach. since we can almost do this with every other native method and what's the outcome?

Chalarangelo:
@iamsoorena I agree, extending native methods should only be done when they handle common edge-cases or it helps somewhere (i.e. a really long method name or a property that is hard to remeber and used a lot or a lesser known property or method that we want to expose new developers to). My vote, too, is to not merge this snippet and mark it as natively implemented in #100.

arjunmahishi:
cool..

skatcat31:
@arjunmahishi What I was trying to guide you to was.

```js
1/0 == Infinity
1/null == Infinity
1/-0 == -Infinity
```

arjunmahishi:
@skatcat31 I will reopen the PR in the morning. Thanks :smile:

skatcat31:
@arjunmahishi I wouldn't reopen the PR as it was decided it might not be the right place here since toString is native and the use cases for this are fairly limited(I've never ONCE ran into a -0), but we still appreciate the PR even if it did get turned down.

Chalarangelo:
@arjunmahishi since this only deals with very specific corner cases and we're kind of not very much in love with the idea so far, better focus your energy on some other snippet, no need to waste your time on this one. Plus, if we decide this should be added, you've already implemented like 80% of it.

PR #298, opened 2017-12-21
Improve isPrime to be O(sqrt(n))

bobby569:
Improve isPrime to be O(sqrt(n)). Some tailing space is removed by IDE

### DIFF
--- a/snippets/isPrime.md
+++ b/snippets/isPrime.md
@@ -6,10 +6,9 @@ Returns `false` if the provided number has positive divisors other than 1 and it
 
 ```js
 const isPrime = num => {
-	for (var i = 2; i * i <= num; i++) if (num % i == 0) return false;
-	return num >= 2;
+  for (var i = 2; i * i <= num; i++) if (num % i == 0) return false;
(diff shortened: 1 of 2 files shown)


skatcat31:
Do we have a rule about tabs versus spaces and how much should be in our kerning?

Chalarangelo:
@skatcat31 we like spaces (2 per level of indentation), I think this comes with (semi)standard, but I am not entirely sure about that. All snippets that I know of are 2-space indented.

snippets/isPrime.md
```diff
@@ -6,9 +6,9 @@ Returns `false` if the provided number has positive divisors other than 1 and it
 
 ```js
 const isPrime = num => {
-  for (var i = 2; i < num; i++) if (num % i == 0) return false;
-  return num >= 2;
-}
+	for (var i = 2; i * i <= num; i++) if (num % i == 0) return false;
+	return num >= 2;
```
skatcat31: Please update form tabs to sapces(2)

Issue #299, opened 2017-12-21
[BUG]everyNth doesn't return what I'd consider the "Nth" values.


Y-Less:
### Expected Snippet Behavior

everyNth([1,2,3,4,5,6], 2) -> [ 2, 4, 6 ]

### Current Snippet Behavior

everyNth([1,2,3,4,5,6], 2) -> [ 1, 3, 5 ]

### Possible Solution

`i % nth === nth - 1`

Chalarangelo:
The behavior is either poorly explained or not perfectly coded, but this definitely needs a fix.

PR #301, opened 2017-12-21
[FIX: #299] Make everyNth return the "Nth" value.

Y-Less:
## Description

**Resolves** #299 - everyNth doesn't actually return the Nth values.


skatcat31:
image

PR #309, opened 2017-12-22
[FEATURE] add call snippet

skatcat31:
## Description
Adds the call snippet which is extremely useful in composition

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/call.md
+++ b/snippets/call.md
@@ -1,0 +1,14 @@
+### Call
+
+Given a key and a set of arguments call them when given a context. Primarily useful in composition
+
+Use a closure to call a stored key with stored arguments
+
+```js
+const call = ( key, ...args ) => context => context[ key ]( ...args )
+/*
+Promise.resolve( [ 1, 2, 3 ] ).then( call('map', x => 2 * x ) ).then( console.log ) //[ 2, 4, 6 ]
+const map = call.bind(null, 'map')
+Promise.resolve( [ 1, 2, 3 ] ).then( map( x => 2 * x ) ).then( console.log ) //[ 2, 4, 6 ]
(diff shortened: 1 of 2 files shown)


iamsoorena:
sorry but what is the difference with simply doing:
```js
Promise.resolve([1, 2, 3]).then(e => e.map(x => 2 * x)).then(console.log)
```
I mean this function makes debugging our program hard. since functionality is achievable in native JS.
Is there any special use case that you can point me to? 
@skatcat31

skatcat31:
Primary benefits are pretty high level in design and not so noticeable in the code, but in the memory, page size, and compilation optimization level of Javascript they make a huge difference. Plus now I don't always have to type `e => e.map` and risk typos every time instead of just `map`. Saving typos, decreasing source size, increasing the ability of the JS engine to optimize it, and knowing it will ONLY fail if you don't give it the right context is pretty good right?
___
example of developer level JS optimizations:
```js
Fetch( getFriendsPictures ).then( fetchToJson ).then( map( friendsToPictures ) ).then( storeFriendsPictures )
// later
Fetch( getPost( postID ) ).then( fetchToJson ).then( map( friendsLikesWithPictures ) ).then(displayPost)
```
You don't need to know what those functions look like, you know what they do. You know if these failed, you would look at only that individual implementation. You also avoid the common problem of mistyping `e => e.map(...)` which is hard to find, and linters won't warn you about since e is an object in that context, and the linter has no idea what it looks like. They will warn you if you mistype `map` though since it hasn't been defined

iamsoorena:
@skatcat31 your explanation was awesome. thank you.

PR #315, opened 2017-12-22
[FEATURE] reduced Filter

myapos:
Filter an array of objects based on condition and return array with reduced objects.

## Description

This function utility is using ES6 Array.reduce() in order to filter an array with objects based on condition.
If condition is met, then Array.reduce is used again to construct object with the predefined properties. Further more, in order to dynamically evaluate condition based in object properties, item must be passed through window to Function constructor. Finally, when new object is constructed then it is saved in the output array. 

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/snippets/reducedFilter.md
+++ b/snippets/reducedFilter.md
@@ -2,21 +2,22 @@
 
 Filter an array of objects based on condition and return array with reduced objects.
 
-// ----------------------------------- Input -----------------------------------//
-// 1. Data: the data to be filtered (array of objects) -------------------------//
-// 2. Condition: will be used for filtering (string) ---------------------------//
-// 3. outputProps: an array of properties that will be used to contruct --------//
-// new array of objects --------------------------------------------------------//
+#### Input
 
-// ----------------------------------- Output ----------------------------------//
-// Filtered array with new objects. Properties of new objects are a subset of --// 
-// properties of original objects ----------------------------------------------//
+i. Data: the data to be filtered (array of objects) 
+ii. Condition: will be used for filtering (string) 
+iii. outputProps: an array of properties that will be used to contruct new array of objects 
 
-// ----------------------------------- Info ------------------------------------//
-// Used ES6 reduce -------------------------------------------------------------//
-// Dummy data for testing ------------------------------------------------------//
-// Generated with http://www.mockaroo.com/ -------------------------------------//
+#### Output
 
+Filtered array with new objects. Properties of new objects are a subset of
+properties of original objects 
+
+#### Info
+
+Used ES6 reduce 
+Dummy data for testing 
+Generated with http://www.mockaroo.com/ 
 
 ```js
 const reducedFilter = (data, condition, outputProps) => 
--- a/snippets/reducedFilter.md
+++ b/snippets/reducedFilter.md
@@ -1,0 +1,37 @@
+### reducedFilter
+
+Filter an array of objects based on condition and return array with reduced objects.
+
+// ----------------------------------- Input -----------------------------------//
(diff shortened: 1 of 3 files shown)


snippets/reducedFilter.md
```diff
@@ -0,0 +1,38 @@
...
+        return aggr;
+      }, {});
+      acc.push(parsedObj); 
+    } 
+    return (acc);
+  }, []);
+```
+#### Usage Example available in :
+
+`https://codepen.io/myapos/pen/dJGByW?editors=0112`
```
Chalarangelo: Examples have to be inlined in the code, absolutely no external links.

snippets/reducedFilter.md
```diff
@@ -0,0 +1,38 @@
...
+
+* Filtered array with new objects. Properties of new objects are a subset of
+properties of original objects 
+
+#### Info
+
+Used ES6 reduce 
+Dummy data for testing 
+Generated with http://www.mockaroo.com/ 
+
```
Chalarangelo: Explain more thoroughly the techniques and methods used.

myapos:
Ok I will be back with the necessary updates. Thanks for your review.

iamsoorena:
beside the things my friends said. just FYI @myapos It's a very good practice to have more descriptive commit messages. :)

snippets/reducedFilter.md
myapos: Hello @Chalarangelo. I added more details in the update files. Could you please review it? Thanks.

snippets/reducedFilter.md
myapos: Added examples in comments

myapos:
@Chalarangelo could you please more specific?

atomiks:
What you're looking for is a callback function instead of an eval'd condition.

```js
const data = [{
  id: 1,
  name: 'john',
  age: 24
}, {
  id: 2,
  name: 'mike',
  age: 50
}];

const reducedFilter = (data, keys, callback) => 
  data.reduce((acc, item) => 
    callback(item)
      ? acc.concat(keys.reduce((acc, key) => {
          acc[key] = item[key];
          return acc;
        }, {}))
      : acc,
  []);

reducedFilter(data, ['id', 'name'], item => item.name.startsWith('m'));
// [{ id: 1, name: 'mike' }]
```

skatcat31:
@atomiks callback is semantically incorrect here since it isn't async and makes it a little confusing(still good code though)

Also why concat instead of push?

```js
const reducedFilter = (data, keys, test) =>
  data.reduce(
    (acc, item) =>
      test(item)
        ? acc.push(
            keys.reduce((acc, key) => {
              acc[key] = item[key];
              return acc;
            }, {})
          )
        : acc,
    []
  );
```

atomiks:
`push` returns the index, not the `acc` array

Also I thought callback wasn't necessarily an async term, it's just when a certain event occurs. For example, you'd your function is called by the `.filter/map/.forEach` methods despite them being sync,  making it a "callback"

skatcat31:
callbacks are a generic terms for a function to be called in response to a trigger. Predicates are callbacks used as a trigger and will always be coerced to Booleans.

example of predicate versus callbacks:
```js
const ifElse = (predicate, ifcb, elsecb) => (...a) => predicate(...a) ? ifcb(...a) : elsecb(...a)
[1,2,3].map(ifElse(x => !x%2, x => x * 2, x =>x)) // [1,4,3] 
```

Again purely a flavor thing, but stating test or predicate will make the code much clearer in the long run, especially for beginners since if they see test or predicate they know it should return a truthy or a falsy instead of an any

PR #335, opened 2017-12-23
[ENHANCEMENT] a better difference

lvzhenbang:
## Description

a faster and more concise of difference

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/difference.md
+++ b/snippets/difference.md
@@ -5,6 +5,6 @@ Returns the difference between two arrays.
 Create a `Set` from `b`, then use `Array.filter()` on `a` to only keep values not contained in `b`.
 
 ```js
-const difference = (a, b) => { const s = new Set(b); return a.filter(x => !s.has(x)); };
+const difference2 = (a, b) => a.filter(x => !b.includes(x))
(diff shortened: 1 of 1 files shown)


Chalarangelo:
I'm pretty certain that in a previous discussion we had decided that `Set` was the best and most intuitive way to go about this, due to the fact that it is pure ES6 and showcases a neat native feature. I'm voting to keep the snippet as-is.

Pl4gue:
Oh alright :+1: :smile: ,didn't know about this discussion @Chalarangelo Well, we can keep it as it is now as well.

Chalarangelo:
I also think that we run a jsPerf and `Set` was faster, but I cannot verify this at this time. Can anyone check and give us some solid evidence on that?

skatcat31:
`Set` is faster according to design(constant time lookup as an object map with key === value(not really, but conceptually), whereas includes will always be O(x*n) where `x == traversed objects`.

JSPerf confirms this with `Set` being faster in all cases including ordered array case

PR #342, opened 2017-12-24
Better syntax highlighting (close to GitHub's)

atomiks:
- [x] Website
- [ ] Snippets

## Types of changes

### DIFF
--- a/docs/prism.css
+++ b/docs/prism.css
@@ -1,32 +1,26 @@
-/* PrismJS 1.9.0
-http://prismjs.com/download.html?themes=prism&languages=clike+javascript */
-/**
- * prism.js default theme for JavaScript, CSS and HTML
- * Based on dabblet (http://dabblet.com)
- * @author Lea Verou
- */
-
 code[class*="language-"],
 pre[class*="language-"] {
-	color: black;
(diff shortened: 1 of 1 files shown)


atomiks:
Prism doesn't tokenize the same though... so methods aren't purple, and the params of a function aren't tokenized so they can't be orange.. etc

Chalarangelo:
I will personally review these changes later today and merge them most likely.

Chalarangelo:
Thanks for putting in the time to make these changes, however I will close the PR for two main reasons:
- We should definitely make no changes in the Prism file, but rather apply our own changes on top of these, so that we can update to later versions of Prism without causing any trouble (yes, I know this creates extra CSS, I will use SCSS to make this less of an issue after Christmas, when I'm back from vacation).
- The current coloration is not at all problematic, as it used the same color scheme as many other websites and, apart from that, there are a lot of changes in this PR that overlap and cause trouble with certain styles that were put in place in the CSS to make sure the reading experience is pleasant, based on best UX practices.

There are issues that need to be dealt with in the Prism styling (especially the ugly background on operators), but we should be able to handle these minor issues outside the Prism.css file.

PR #345, opened 2017-12-24
[FEATURE] Generate tests and run tests with AVA

mycaule:
## Description

Generates tests from `snippets/` folder to `tests/` folder. The script can possibly be merged with `lint-script.js`.
The generated files can then be run against [ava](https://github.com/avajs/ava), those files are gitignored. 

During the build, the maintainer or CI system has to run `npm run tester`. 
See https://asciinema.org/a/154143 for the demo.

Only 80% of the tests pass at this point. 
I still have to handle browser functionnalities and multicomments with `/* */`.

Please agree on the workflow and the proposed testing tool before going further.

See this discussion https://github.com/Chalarangelo/30-seconds-of-code/issues/78

General / Things regarding the repository (like CI Integration)

## Types of changes
New feature (non-breaking change which adds functionality)


Chalarangelo:
I will review this later when I am on my laptop and see if this works and how well it can be integrated. Thanks for the contribution!

kingdavidmartins:
@mycaule This is exactly what I was describing. The only difference here is you 

1. You decided to go with `ava` instead of `tape`
2. You are streaming the the contents of the each file and testing them on the fly rather than generating the actual `test/` with the test & It's described contents

Also the reason I didn't go with just using assert.deepEqual is because what If said users want to test their code differently. They should be able to test using things like `assert.doesNotThrow()` or `assert.notOk()` & etc. Your proposed solution imposes deepEqual for test. Having some sort of flag give users the ability have more say over the kind of test they want to run & etc

Also I believe the test files should be generated to make dubbing less cumbersome

skatcat31:
Issues:
 - Breaks as soon as we update the snippet structure
 - Only tests validity of user submitted examples, is not an actual test
 - Breaks on multi line complex examples
 - Fragile as it makes assumptions

Good things:
 - Example of changes code needed to extract snippet to put it into a file for testing purposes( with some adaption )

Chalarangelo:
As @skatcat31 and @kingdavidmartins have already mentioned:
- Snippet testing is currently on hold and will remain so until we restructure snippets (primary reason I am closing this PR).
- There are some problems that stem from this particular implementation, meaning some snippets will not test properly and there is no architecture in place for the actual testing.

Based on these issues, I will close this PR and feel free to help out when we actually decide on a test framework, restructure snippets and decide on the specifics of the implementation we are going to roll with.

Thanks for the contribution and your time nonetheless.

PR #351, opened 2017-12-24
added sdbm-hash algorithm

christianbender:
I added the sdbm-hash algorithm as a function called **sdbm**


snippets/sdbm-hash.md
```diff
@@ -0,0 +1,13 @@
+
+This function implements the **sdbm** hash-algorithm. 
+It uses the charCodeAt-methode and the shift-operator.
```
Chalarangelo: Please update the description to explain the methods and techniques used.

snippets/sdbm-hash.md
```diff
@@ -0,0 +1,13 @@
+
+This function implements the **sdbm** hash-algorithm.
```
Chalarangelo: Please update the description to explain what the algorithm does.

snippets/sdbm-hash.md
```diff
@@ -0,0 +1,13 @@
+
+This function implements the **sdbm** hash-algorithm. 
+It uses the charCodeAt-methode and the shift-operator.
+
+``` js
+function sdbm(s) {
```
Chalarangelo: Please follow the styleguide, as seen in CONTRIBUTING.md and snippet_template.

snippets/sdbm-hash.md
```diff
@@ -0,0 +1,13 @@
...
+    var hashCode = 0;
```
Chalarangelo: Please use `let` or `const` instead of `var`.

snippets/sdbm-hash.md
```diff
@@ -0,0 +1,13 @@
...
+    for (var i = 0; i < s.length; i++) {
+        hashCode = s.charCodeAt(i) + (hashCode << 6) + (hashCode << 16) - hashCode;
+    }
```
Chalarangelo: Prefer using `Array.map` or `Array.reduce` over loops, if possible.

snippets/sdbm-hash.md
```diff
@@ -0,0 +1,13 @@
...
+    return hashCode;
+}
+```
```
Chalarangelo: Please add examples with sample input and expected output in your snippet.

christianbender:
@Chalarangelo  

hello, 

thanks for feedback. 
I will fix that style issues as fast as I can
It is a simple hash algorithm. see here.
http://www.cse.yorku.ca/~oz/hash.html
https://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed

greetings

christianbender:
@iamsoorena  

I don't know. I need to access to the last item I puted in the array. I considered over  the map-function. Has someone got an idea?

kriadmin:
@christianbender This is how you can do it using `Array.reduce()`
[Array.reduce() MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) 
```js
const sdbm = str => {
     let arr = str.split('');
    return arr.reduce( (hashCode,currentVal) =>{
                 return  hashCode = currentVal.charCodeAt(0) + (hashCode << 6) + (hashCode << 16)  - hashCode;
     }
 ,0)
 }

christianbender:
@kriadmin Thanks! Why didn't I come up with this myself?

PR #365, opened 2017-12-27
[FIX] lint-script.js

atomiks:
## Description
So after some debugging I found that using `.temp` made the linter not actually run. By removing it it works. I also cleaned the code style a bit... 

**Lodash[ACTION]** #100 -> https://lodash.com/docs/4.17.4#(method) 

- [ ] Website
- [x] Snippets

## Types of changes

- [ ] I added the prefix [UPDATE: `method.md`] or [ADD: `method.md`]

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -391,7 +391,7 @@
     "builder": {
       "version": "3.2.3",
       "resolved": "https://registry.npmjs.org/builder/-/builder-3.2.3.tgz",
-      "integrity": "sha512-DXkZ27RHCw4uCn08SEx7hhgXbtamZOBKa8FEuHjsooyMfNfZXNn7YYqpfWrBNigwX2sMgKYWY8Jupxp6waofSg==",
+      "integrity": "sha1-M7I5+mLSQygCFk63Ll9JY5sK4RU=",
       "requires": {
         "async": "1.5.2",
         "chalk": "1.1.3",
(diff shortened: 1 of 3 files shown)


atomiks:
Ok so in order for the code to be properly normalized/formatted prettily, we DO need prettier as well.  `semistandard --fix` does help fix some issues but doesn't fully re-format it. 

So here I've added it to run after the `semistandard` command and used `.replace()` with some regex instead of `.slice()` to replace the file with the linted code since it's a bit easier to understand.

atomiks:
No we need prettier because the linter formatting doesn't fix this one (and similar):

```js
const difference = (a, b) => { const s = new Set(b); return a.filter(x => !s.has(x)); };
```

It should break to new lines

Chalarangelo:
No, it shouldn't. This is intentional, as we want snippets that can be displayed in one line to be displayed in one line (this is something we do manually so far).

atomiks:
But that one shouldn't be displayed on a single line because it needs to be separated by a semicolon. It looks cluttered...

Chalarangelo:
The whole styling thing is a flaming pile of garbage at this point, our guidelines aren't well-defined enough about things like this and we have no specific line length or anything to sort this issue. I would prefer if we jsut fixed the file extension problem in this PR and actually open an issue to discuss these guidelines, so we can then set up `prettier` with a set of well-defined rules that we have decided on as a community, wouldn't you agree?

atomiks:
Yes that's a good idea. I'll ake a new PR with just the extension changes.

And we can also lint/prettier-ify the script files too

Chalarangelo:
Yep, send a PR with the extension changes and let's open up a discussion for this to define the styling a little bit better then! Thank you! 👍

PR #370, opened 2017-12-27
[FIX] Hyper speed-boost lint-script to lint examples + use prettier formatting

atomiks:
## Description

Makes it like 60x faster by running semistandard on the entire folder. Lints and prettifies both the function definition and example blocks

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -561,6 +561,7 @@
       "requires": {
         "anymatch": "1.3.2",
         "async-each": "1.0.1",
+        "fsevents": "1.1.3",
         "glob-parent": "2.0.0",
         "inherits": "2.0.3",
         "is-binary-path": "1.0.1",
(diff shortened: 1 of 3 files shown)


skatcat31:
I'm curious why we're reading it one synchronously instead of using chaining?

atomiks:
What do you mean? From what I've read, sync read/write methods are faster than async ones. The total linting takes around 1s-3s on my laptop.

kingdavidmartins:
I agree with @atomiks 

> sync read/write methods are faster than async ones.

Especially for the purpose of this project. Since we will be running it on our machines and not on a web server of sorts. I believe It makes sense to go this route.

[SO](https://stackoverflow.com/questions/13822085/fs-readfilesync-seems-faster-than-fs-readfile-is-it-ok-to-use-for-a-web-app-in)
[Medium](https://medium.com/@adamhooper/node-synchronous-code-runs-faster-than-asynchronous-code-b0553d5cf54e)

skatcat31:
I misread a section of the commit diff. When looking at the full file it becomes a little more clear what's happening( I use a color changing app that had a hiccup apparently, making the removed code look like it was still present ). It looked to me like we were manually linting each file and then sending ti to prettify synchronously instead of async, or as a folder. Now that I see it's over the whole folder, yeah no this is fine XD

Perfect storm:
 - Web colorizer
 - Automatic contrast negotiation on graphics card driver when not in game
 - Blue Light reducer

Surprised it hasn't been an issue until now, but this is the largest stream of commits I've seen in awhile in one MR in this project.

Issue #376, opened 2017-12-28
[Bug] UUID Generator doesn't work in Node.js 8.9.1


KingDGrizzle:
### Expected Snippet Behavior
It generates an UUID

### Current Snippet Behavior
It errors out.

### Possible Solution
Maybe figure out a better UUID generator using `crypto.randomBytes` or whatnot

EDIT: Reading this insanity https://gist.github.com/jed/982883#gistcomment-1615714 I see the fix would be to replace `getRandomBytes(new Uint8Array(1))` with `randomBytes(1)`

Chalarangelo:
I haven't tested the snippet in a little while, but I trust that it won't work on my environment either. I'll see if I or someone else from the team can fix this. If you have the time and come up with a solution (like the one you mentioned), feel free to submit a PR with that fix! 👍

KingDGrizzle:
Of course. I'll look into a patch, and submit a PR.

skatcat31:
This should be a seperate version then, however if you're REALLY going to generate UUIDs in node you should be using the `node-uuid` package

PR #380, opened 2017-12-28
[FIX: #376] Update UUIDGenerator with new examples

KingDGrizzle:
## Description
**Resolves** #376 
Also removes trailing spaces due to VSCode. Woops!

Let me know if I should re-make my PR due to VSCode (and me opening the README file and saving it which caused VSCode to remove all trailing spaces.)

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/UUIDGenerator.md
+++ b/snippets/UUIDGenerator.md
@@ -4,13 +4,26 @@ Generates a UUID.
 
 Use `crypto` API to generate a UUID, compliant with [RFC4122](https://www.ietf.org/rfc/rfc4122.txt) version 4.
 
+Please note that this version will only work in **Node.JS** environments!
+
 ```js
+const crypto = require("crypto");
+
 const UUIDGenerator = () =>
   ([1e7] + -1e3 + -4e3 + -8e3 + -1e11).replace(/[018]/g, c =>
-    (c ^ (crypto.getRandomValues(new Uint8Array(1))[0] & (15 >> (c / 4)))).toString(16)
+    (c ^ (crypto.randomBytes(1)[0] & (15 >> (c / 4)))).toString(16)
   );
 ```
 
+If you'd like a dependency-less UUID Generator that will work in both browsers and Node.JS, you should choose this one instead.
+
 ```js
-UUIDGenerator(); // '7982fcfe-5721-4632-bede-6000885be57d'
+const UUIDGenerator2 = a =>
+	a ? ((a ^ Math.random() * 16) >> a / 4).toString(16) :
+		([1e7] + -1e3 + -4e3 + -8e3 + -1e11).replace(/[018]/g, UUIDGenerator2);
+```
+
+```js
+UUIDGenerator();  // '7982fcfe-5721-4632-bede-6000885be57d'
+UUIDGenerator2(); // '777c85e1-6fdf-47f4-8c65-2842487453bb'
 ```
--- a/README.md
+++ b/README.md
@@ -1232,8 +1232,8 @@ console.log(pulled); // [ 'b', 'd' ]
 
 QuickSort an Array (ascending sort by default).
 
-Use recursion. 
-Use `Array.filter` and spread operator (`...`) to create an array that all elements with values less than the pivot come before the pivot, and all elements with values greater than the pivot come after it. 
+Use recursion.
+Use `Array.filter` and spread operator (`...`) to create an array that all elements with values less than the pivot come before the pivot, and all elements with values greater than the pivot come after it.
 If the parameter `desc` is truthy, return array sorts in descending order.
 
 ```js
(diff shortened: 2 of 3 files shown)


snippets/UUIDGenerator.md
```diff
@@ -4,13 +4,26 @@ Generates a UUID.
...
-    (c ^ (crypto.getRandomValues(new Uint8Array(1))[0] & (15 >> (c / 4)))).toString(16)
+    (c ^ (crypto.randomBytes(1)[0] & (15 >> (c / 4)))).toString(16)
   );
 ```
 
+If you'd like a dependency-less UUID Generator that will work in both browsers and Node.JS, you should choose this one instead.
+
+```js
+const UUIDGenerator2 = a =>
+	a ? ((a ^ Math.random() * 16) >> a / 4).toString(16) :
```
skatcat31: Math.random is not cryptographically safe, and thus not suitable for use in generating UUIDs

snippets/UUIDGenerator.md
```diff
@@ -4,13 +4,26 @@ Generates a UUID.
 
 Use `crypto` API to generate a UUID, compliant with [RFC4122](https://www.ietf.org/rfc/rfc4122.txt) version 4.
 
+Please note that this version will only work in **Node.JS** environments!
```
skatcat31: incorrect. Browsers also exposes a global crypto

snippets/UUIDGenerator.md
```diff
@@ -4,13 +4,26 @@ Generates a UUID.
...
+
 ```js
+const crypto = require("crypto");
```
skatcat31: needed only in nodeJS, browsers expose a crypto object

snippets/UUIDGenerator.md
```diff
@@ -4,13 +4,26 @@ Generates a UUID.
...
+
 const UUIDGenerator = () =>
   ([1e7] + -1e3 + -4e3 + -8e3 + -1e11).replace(/[018]/g, c =>
-    (c ^ (crypto.getRandomValues(new Uint8Array(1))[0] & (15 >> (c / 4)))).toString(16)
+    (c ^ (crypto.randomBytes(1)[0] & (15 >> (c / 4)))).toString(16)
   );
 ```
 
+If you'd like a dependency-less UUID Generator that will work in both browsers and Node.JS, you should choose this one instead.
```
skatcat31: this is not needed as browsers expose a crypto API and nodeJS would only need ot include crypto to get the exact same functionality

snippets/UUIDGenerator.md
Chalarangelo: Therefore, this should be added as part of the description and not part of the code.

snippets/UUIDGenerator.md
Chalarangelo: Yup, `Math.random` is a bad bad idea.

snippets/UUIDGenerator.md
KingDGrizzle: Not what I meant. There are 2 different functions to do the same thing in different platforms. Web has getRandomValues, while Node.JS has randomBytes

Chalarangelo: Maybe we should actually add a Node version of this snippet then, under the **Node** category, this one was for browser, so maybe we can change the tag for this, too and keep them separate. That sounds like a better plan!

KingDGrizzle:
Alrighty. I'll handle the requested changes tomorrow (although I don't know how I can revert the readme changes.)
But, where do I make the Node UUID (and how do I define it), and where do I do the Web one? I don't know the category system yet, so :P

snippets/UUIDGenerator.md
skatcat31: Fair. Had not remembered that. We may want to note the differences instead of having two different examples. I'd suggest a browser version with a note on the changes for a nodeJS version since most nodeJS developers probably just use the node-uuid package since that is tried and true

snippets/UUIDGenerator.md
skatcat31: update: crypto in browser is different, however a note would be better than a secondary code block example for style reasons

Chalarangelo:
@KingDGrizzle Good questions. 

- First off, I'll close this PR and start over in a new branch, so that you can write the snippet as `UUIDGeneratorNode.md` or something like that.
- After you are done writing it, run `npm run tagger` and then open `tag_database`. Find your new snippet and add set the category to `node` (e.g. `UUIDGeneratorNode:node`). 
- Change the category of the other snippet we now have to `browser` in `tag_database` (e.g. `UUIDGenerator:browser`).
- Do not run `builder, `webber` or `linter` as these are currently handled automatically by our Travis CI, so changes will build a couple of minutes after you submit your PR.

PR #385, opened 2017-12-29
[FIX #295] Update lcm.md and gcd.md

### LINKED ISSUE #295: [SUGGESTION] Array, xValues Generalizations
### Description 

I believe we should decide how we want methods that take a array or collection of values to be implemented 

When going through the snippets I realized that methods like `arrayMax()` or `arrayMin()` or any other general method meant for a collection of values only work for 1 specific type. 

<strong>*For example:*</strong>
```js
const arrayMax = arr => Math.max(...arr);
arrayMax([10, 1, 5, 23, 2, 12445, 25])// -> 12445
arrayMax(10, 1, 5, 23, 2, 12445, 25)// -> Breaks & Throws an Err
```

I think rather than creating a seprate method to handle the said exception/problem I think we could just easily handle and update all snippets with a oneliner that could easily fix the stated problem

I think its important for the following reasons: <strong>*consistency*</strong>, <strong>*readability*</strong>, <strong>*performance*</strong>, and <strong>*modernity*</strong> 

 Early on in the project we decided to stray from implementing 
`pull()` for `pullAll()` simply because 1 takes arr and the other take x amount of values and it didn't makes sense to implement 2 different functions that do the same thing when can easily be turned to 1 function.

It also helps reduce clutter  

we ended up decided merging it into 1 function.

Now there's quite a few of snippets initializing like so.  Only implementing one or the other

Just wanted to make sure I knew which is preferred in the future when implementing/updating other array/collection type methods/snippets & the reasoning behind which we do so choose

kriadmin:
## Description
Update lcm.md and gcd.md to accept variadic arguments and arrays.
Also deleted arrayGcd.md and arrayLcm.md

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:


skatcat31:
I'm confused why you have a flattened variadic function?

PR #390, opened 2017-12-29
[FEATURE] Create a script for building a module

atomiks:
## Description
Creates a script that can build a module from the snippets.

Created with the help of @kriadmin 

- Run with `npm run module`
- Places the module files in the `dist/` directory. It's just an object with each property being the snippet function.
- UMD format + ESM
- I don't know where the rollup and babel deps should lie, `deps` or `devDeps` in package.json?

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:


Chalarangelo:
I'd like core team member to state their opinions on this, as this issue has come up multiple times. As far as I am concerned, I believe the following:
- Having a module/package generated from our snippets isn't a terrible idea, however we should mark it as not production-ready and make sure people know that the code is exactly the same as the one seen in the repo/website.
- Extracting the module will be a manual (or scheduled once a month or so) task, not one we will add to the core build process of Travis CI, as that would be problematic in many ways.
- Creating a module could be useful for TDD (and vice versa), so don't dismiss the idea entirely.
- We need to come up with a name and someone needs to keep an eye for the builds as well as release the package every once in a while.

skatcat31:
Since there's so many thoughts I have on this I'm going to drop them here in no particular order. I have more, but here's the forefront of what comes to mind
___
It's a cool idea, but it's really more for just making it something you can just push out to a playground for exploring. If that's the case we may want to do this and include it in the browser of the site so people can open the browser and just call the method(or do a REPL)

We should make more than one if this is the case: one for node capable, and one for browser capable but that would require a way to mark something as node/client but lets admit it a micro collections of things like the adapters, flatten, and some other snippets would be super helpful to node developers who don't want the super powers that come with Ramda or _ if they just want some helpers

Would there be functions we wouldn't want to include?

kriadmin:
@skatcat31 We already are provided a testing playground for users at runkit.com/npm/package-name . We can provide a link to it on the site. 
As for your second thought @atomiks has already made it UMD so that it can be used anywhere from a single package. 
And yes for now he is excluding the node snippets as they require modules so we can't use it in browser. But we may overcome this problem soon. 
Also we are automatically provided a CDN at unpkg.com/package-name . 😃

iamsoorena:
I think we are not pushing someone to use this package. It's not a big deal. but I will use this package. It's lightweight and It has a lot of useful snippets.
about node/client issue, we can separate them with `npm run module client` and `npm run module node`. nothing more than that.
Overall, I like this idea.

fejes713:
> I think we are not pushing someone to use this package. It's not a big deal. but I will use this package. It's lightweight and It has a lot of useful snippets.

Just this on its own is enough, I absolutely agree with @iamsoorena 

> However we should mark it as not production-ready and make sure people know that the code is exactly the same as the one seen in the repo/website.

@Chalarangelo 👍  As discussed a few times already our snippets aren't meant for production really. The do work but they're not as robust for production. 

In general, I think this is awesome idea, 👍  @atomiks

atomiks:
I'm going to create some different types of bundle files. 

- Non-minified versions of UMD and ESM
- Minified versions of UMD and ESM
- ES2017 version (no transpilation)
- ES5 version (also, they would need to include `babel-polyfill` because of snippets' reliant use on modern methods, constructors, etc, that can't be detected easily).

skatcat31:
Should we export an ES5 version?

Chalarangelo:
Moving to an organization would basically destroy our traffic and all the link-building we have done, so I am strongly against this at this time, especially with the project just having launched on ProductHunt and getting extra traffic.

Releasing a package can be relatively harmless, provided the fact all of the code is contained in the README anyways and that we can use semantic versioning to mark as non-production ready (i.e. `0.x.y-beta`) so that it is abundantly clear that nobody should use it in production. Anyways, people will most likely use snippets in production anyways, so this is not a problem we can solve and quite frankly we don't have to deal with the fact that people might not read the 5 lines of description at the top of the README.

However, the whole project focuses on learning/teaching, instead of providing a library. The only reason I am agreeing to release a module/package is that we have been getting requests about that non-stop for the past two weeks, so better do that and let people take their risks using our code like that. There have been at least 2 repos that I know of that have packaged our code into their own npm packags, so I'd rather we could at least maintain the code packages ourselves and be able to keep an eye out on everything properly.

On a side note, it is highly unlikely that we will get the 50mil monthly downloads that lodash has, so I wouldn't worry about people using this in production all that much. 

As far as the playground ideas go, we can use an online tool like [repl.it](https://repl.it/) to embed a playground for each snippet in the website, much like lodash does, making certain that each snippet runs on its own. There's no real need to have a playground with multiple snippets or a whole pacakge in my opinion.

mariastervic:
There are so many npm packages that are not production ready, yet fail to mention that. I think that 30seconds of code could use a package and having a notice, plus use of semver could help make sure people are informed it is not for production. You are not forcing anyone to drop lodash and use 30seconds of code instead, so that will be fine.

kingdavidmartins:
@Chalarangelo Make's sense. 

> Anyways, people will most likely use snippets in production anyways, so this is not a problem we can solve.

Which was & is my main concern, which I believe you've addressed & considered

> On a side note, it is highly unlikely that we will get the 50mil monthly downloads that lodash has, so I wouldn't worry about people using this in production all that much. Plus, even if we do, we can see it coming and figure out what to do.

No matter how improbable I think there's potential. I don't think we'll replace lodash or underscore or any of the similar well established lib's, But I do think we can become an alternative lib or sorts that might be preferred during certain dev cycles/circles. Although it's mere speculation, I'm sure time will tell 😉 

Really excited to see how the projects develops in the years to come 😄 ⚡️

Chalarangelo:
It's time...

PR #417, opened 2017-12-29
Create onUserInputChange.md

atomiks:
## Description

Runs the supplied callback whenever the user changes their input method (mouse or touch).
It's the most robust solution I had come up with for my tooltip library.

Simply using a userAgent or `'ontouchstart' in window` fail in a lot of cases. 

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/onUserInputChange.md
+++ b/snippets/onUserInputChange.md
@@ -1,0 +1,43 @@
+### onUserInputChange
+
+Will run the callback whenever the user changes their input type (either `mouse` or `touch`). This is useful
+if you want to disable certain code depending on if the user is using touch as input or a mouse (including trackpads).
+
+Use two event listeners. Assume `mouse` input initially and bind a `touchstart` event listener to the document. 
+On `touchstart`, the callback is run and supplied with the current input type as an argument. 
+Then, add a `mousemove` event listener to listen for two consecutive `mousemove` events firing within 20ms 
+using `performance.now()` (browsers recently fire them every animation frame). Run the callback and supply the new type
+`mouse` as the argument. This process needs to happen dynamically because of hybrid devices (such as a touchscreen laptop), 
+where the user can switch between either input type at will.
+
+```js
+const onUserInputChange = callback => {
+  let type = 'mouse';
+
+  const mousemoveHandler = (() => {
+    let lastTime = 0;
+    return () => {
+      const now = performance.now();
+      if (now - lastTime < 20) {
+        type = 'mouse';
+        callback(type);
+        document.removeEventListener('mousemove', mousemoveHandler);
+      }
+      lastTime = now;
+    }
+  })();
+
+  document.addEventListener('touchstart', () => {
+    if (type === 'touch') return;
+    type = 'touch';
+    callback(type);
+    document.addEventListener('mousemove', mousemoveHandler)
+  });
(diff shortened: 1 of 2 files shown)


fejes713:
Great snippet 👍 , however:
>CONTRIBUTING.md
> Snippets should be short (usually below 10 lines)

Considering this, this won't be a great great idea to merge. Maybe, we could make separate `.md` where we would keep those longer useful snippets.

atomiks:
I've used the comma operator to make it shorter. Still over 10 lines tho..

fejes713:
@Chalarangelo any opinion about this one? After the changes it's much shorter, I think we could let this one merge into our codebase

skatcat31:
While more than ten lines, it shows some useful management for detecting touches versus clicks. The question is how is that useful?

I think it's helpful if you want to hide certain things based on touch versus type... but I don't know when I'd ever do that. Maybe for disabling tooltips or changing how tool tips work, but often if that's the case I use a tip box and an onfocus handler for the form to update the tip box with a bound message on the element...

I think it's cool, but I'm struggling to find where I'd want to use it and when it'd be a good design decision to do so

@atomiks where would you use this snippet?

atomiks:
@skatcat31 in my tooltip library, there is a `hideOnClick` option by default. This only applies to mouse devices, since it won't ever show the tooltip on touch if they're using that input method.  There's a couple other options too that need to distinguish between the two, but it's relatively rare I guess. Still I think it shows a technique for detecting it

skatcat31:
If you're displaying tooltips on click instead of onFocus I think you're using the tooltips the wrong way. Touch users should also be able to see the tooltip for the form input, and what happens with tab navigation with your tooltip library?

atomiks:
It shows them on `mouseenter focus` (allowing tabbed nav), but clicking the element will hide the tooltip. For touch devices this doesn't apply -- it's just a tap to show it, then a tap elsewhere to hide it.

skatcat31:
... I think no matter what we can all agree that stylized tooltips SUCK... I use the title field since it displays on interaction and hover, and fades out on input( the default browser tooltip ) and if I need a stylized one, I use a tip box that floats and follows with an explanation along with the form, and will get updated by the onfocus event of the inputs will fire an event to update the tips box with a string
```js
const updateTip = (function(node, text){ node.innerHTML = text }).bind(null, document.getElementById('tipBox'))
```
```html
<input type='text' onfocus='updateTip(`Hi mom!`)'>
```
since now whoever does the page can update the tipbox however they want, but there are infinitely many ways to do it... that I find it hard to find cases for this code to justify it without it being extremely specific. Maybe we should do an advanced snippet category on a seperate page with much longer explinations and examples? Maybe each advanced is a different page?

Chalarangelo:
@skatcat31 We might eventually add a category for more advanced snippets, but I think we can all agree this is an interesting one to have, even if slightly longer than the average snippet in the repo. I am merging this just because it's really cool and because I think we can get away with a few snippets being a bit more advanced or longer for the sake of showcasing something that many devs might not be familiar with.

@atomiks I shortened the description a tiny bit, just to make sure people don't get intimidated by its length. I also removed some curly brackets from an `if` statement, as they were not necessary.

PR #420, opened 2017-12-30
fix fibonacci example typo

bobby569:
fix fibonacci example typo in Math section


bobby569:
Found the same issue with fibonacciUntilNum. The expected output should be [0, 1, 1, 2, 3, 5, 8] for fibonacciCountUntilNum(10)

PR #422, opened 2017-12-30
[FEATURE] Elo rating system snippet

atomiks:
## Description

https://en.wikipedia.org/wiki/Elo_rating_system

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/elo.md
+++ b/snippets/elo.md
@@ -5,14 +5,14 @@ of two pre-ratings and returns an array containing two post-ratings.
 The winner's rating is the first element of the array.
 
 Use `Math.pow()` and math operators to compute the expected score (chance of winning) of each opponent
-and compute the change in rating for each. Omit the second argument to use the default K-factor of
+and compute the new rating for each. Omit the second argument to use the default K-factor of
 32, or supply a custom K-factor value.
 
 ```js
 const elo = ([a, b], kFactor = 32) => {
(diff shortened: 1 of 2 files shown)


skatcat31:
This is useful in multiplayer competitive leagues using ELO but ELO is now considered an out of date rating system for video games due to the extreme de-bouncing needed. While still used, it is instead preferred to use a more metric based approach to your games(truskill, skillz, other rating systems that take into account minute metrics to determine ratings).

ELO is still a valid rating system for games that use strategy for complexity. Such games are Chess, Go, or any perfect information symmetrical game. While a cool example, I'm not certain it has much use here since if it isn't one of these few games and it has minute differences in skill like 'headshot' or 'objective data' then it is an extremely inaccurate rating.

atomiks:
Can you explain further about the debouncing? Is that to do with rating inflation or something?

Do you think it's accurate for a 1v1 game of Call of Duty where it's either win or loss? I've researched a bit and came across Glicko 2 and TrueSkill but their implementations are too complex for a snippet (I think).

Chalarangelo:
I'm merging as this has practical uses, even if there are better ranking systems than this one. We can add more snippets like this anyways and someone might need this, so why not?

skatcat31:
Debouncing and ELO rating are required to clamp gains based on differences in the ratings and how often you defeat opponents of the same rating as you, and opponents of a lower rating than you

PR #427, opened 2017-12-30
[ADD: `size.md`] add size snippet

phollome:
## Description

**Lodash[ADD: size.md]** #100 -> https://lodash.com/docs/4.17.4#size 

- [ ] Website
- [x] Snippets

## Types of changes

- [x] I have referenced the `method` to the lodash backlog.

## Checklist:

### DIFF
--- a/snippets/size.md
+++ b/snippets/size.md
@@ -1,0 +1,15 @@
+### size
+
+Get size of arrays, objects or strings.
+
+Get type of value – array, object and string. Use `length` property for arrays. Return `length` or `size` value if available or number of object keys. Split strings into array of characters with `split('')` and return its length.
+
+```js
+const size = value => Array.isArray(value) ? value.length : value && typeof value === 'object' ? value.size || value.length || Object.keys(value).length : typeof value === 'string' ? value.split('').length : 0;
+```
+
+```js
+size([ 1, 2, 3, 4, 5 ]); // 5
+size('size'); // 4
+size({ one: 1, two: 2, three: 3 }); // 3
+```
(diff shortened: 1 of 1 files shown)


atomiks:
Is there a reason you're splitting the string instead of just using `.length`?

```js
const size = value =>
  Array.isArray(value) || typeof value === 'string'
    ? value.length
    : value && typeof value === 'object'
    ? value.size || value.length || Object.keys(value).length
    : 0;
```

lvzhenbang:
I agree with @atomiks. In addition，you should first check to see if  `value` exists。

skatcat31:
... functions get tricky with this but if someone is passing a function to a length check they're doing it for something unrelated to value checking... or are horrible people XD

snippets/size.md
```diff
@@ -0,0 +1,22 @@
...
+Get type of value – array, object and string. Use `length` property for arrays. Return `length` or `size` value if available or number of object keys. Split strings into array of characters with `split('')` and return its length.
+
+```js
+const size = value =>
+  Array.isArray(value)
+    ? value.length
+    : value && typeof value === 'object'
+    ? value.size || value.length || Object.keys(value).length
+    : typeof value === 'string'
+    ? value.split('').length
```
skatcat31: @atomiks mentioned this but didn't start a review. should just use `value.length` to avoid the loop and because both values should be the same.

Problem: Emojis. While valid unicode, don't work quite right with this.

😄 has a length of 2, but a bytesize of 4

atomiks: I honestly don't know how to highlight a section of the code 🤔 😆 . You can use the `byteSize` string's trick, `new Blob([str]).size`.

phollome:
@atomiks Didn't know that it's so easy to get a string size 😆

PR #431, opened 2017-12-31
[FEATURE] isPrimitive utility snippet

atomiks:
## Description

Determines if a value is primitive or not.

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:


snippets/isPrimitive.md
```diff
@@ -0,0 +1,22 @@
+### isPrimitive
+
+Returns a boolean determining if the supplied value is primitive or not.
+
+Use `Array.includes()` on an array of type strings, supplying the type using `typeof`.
+Since `typeof null` evaluates to `'object'`, it needs to be directly compared.
+
+```js
+const isPrimitive = value =>
+  ['string', 'number', 'symbol', 'boolean', 'undefined'].includes(typeof value) || value === null;
```
Pl4gue: According to Mozilla definition, a primitive is "_data that is not an object and has no methods_". Therefore, wouldn't that be a much more elegant solution?
```js 
const isPrimitive = value =>
typeof value != 'object' && typeof value != 'function' || value === null;
```

atomiks: In that case...

```js
const isPrimitive = val => !['object', 'function'].includes(typeof value) || value === null;
```

Chalarangelo: @atomiks I like that, it's short and easier to read, plus we can keep the description the same. Care to update this PR so that we can merge it?

PR #457, opened 2018-01-02
[UPDATE] memoize.md

atomiks:
## Description

Uses a `Map` instead of a plain object to store any value as a key, and also allows access to the cache by setting a `.cache` property on the returned function.

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/memoize.md
+++ b/snippets/memoize.md
@@ -2,13 +2,16 @@
 
 Returns the memoized (cached) function.
 
-Use `Object.create(null)` to create an empty object without `Object.prototype` (so that those properties are not resolved if the input value is something like `'hasOwnProperty'`).
+Create an empty cache using by instantiating a new `Map` object.
 Return a function which takes a single argument to be supplied to the memoized function by first checking if the function's output for that specific input value is already cached, or store and return it if not.
+Allow access to the `cache` by setting it as a property on the returned function.
 
 ```js
 const memoize = fn => {
-  const cache = Object.create(null);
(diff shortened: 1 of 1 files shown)


snippets/memoize.md
```diff
@@ -2,13 +2,16 @@
...
 Return a function which takes a single argument to be supplied to the memoized function by first checking if the function's output for that specific input value is already cached, or store and return it if not.
+Allow access to the `cache` by setting it as a property on the returned function.
 
 ```js
 const memoize = fn => {
-  const cache = Object.create(null);
-  return value => cache[value] || (cache[value] = fn(value));
+  const cache = new Map();
+  const cached = val => cache.has(val) ? cache.get(val) : (cache.set(val, fn(val)) && cache.get(val));
+  cached.cache = cache;
```
skatcat31: Why are we exposing the Cache? This leads to outside interference leading to unpredictable results.. We should keep it a private should we not?

atomiks:
Lodash exposes the `cache` property.


>Note: The cache is exposed as the cache property on the memoized function. Its creation may be customized by replacing the _.memoize.Cache constructor with one whose instances implement the Map method interface of clear, delete, get, has, and set.

atomiks:
It doesn't make sense for memoized functions to be variadic, right? 

Unless you pass it a single object or array (the SAME object/array -- otherwise you can check if the new one has the exact same elements/props as the stored one...)

skatcat31:
Memoized functions need to be unary from what I know

PR #459, opened 2018-01-02
[FEATURE] add infixToPostfix and postfixToInfix

kriadmin:
- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/infixToPostFix.md
+++ b/snippets/infixToPostFix.md
@@ -4,7 +4,7 @@
 ```js
 const infixToPostfix  = expr => {
   let rpn = ''
-  let solve = expr.replace(/\^/g,'**').match(/([0-9]+|[\+\(\)\/\-]|\*+)/g).filter(el => !/\s+/.test(el) && el !== '')
+  let solve = expr.replace(/\^/g,'**')str.replace(/([^\d])\.(\d+)/g, '$10.$2').match(/((\d\.?)+|[\+\(\)\/\-]|\*+)/g).filter(el => !/\s+/.test(el) && el !== '')
   let stack = []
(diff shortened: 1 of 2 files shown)


kingdavidmartins:
Also I believe this is data structure specific & as such I believe adding other categories representing different operations/manipulation of other data structures other than arrays can also be beneficial to the project.

Example:
linkedList:
DoubleLinkedList:
DFS
BFS
Trees
Tries
Etc

Thought's @Chalarangelo

kriadmin:
@kingdavidmartins I didn't add proper description as I am waiting for suggestions to imporve the second snippet(postFix to Infix)

skatcat31:
Do we want to keep allowing more RPN stuff? I know we have the solveRPN snippet, but are we certain we want to go deeper with things like translators? They are large, but interesting. I'm just not certain they fall under '30 seconds to understand'... especially with how esoteric RPN is these days

kingdavidmartins:
Honestly @skatcat31 I agree

> especially with how esoteric RPN is these days

Couldn't have said it better myself. Def takes more than 30 seconds to grasp and as such out of scope with the current project.

Personally I believe we should vote to decide if we want to merge/add the following snippets into the project

I vote it shouldn't for the described reasons above 👎

kriadmin:
Even I think this snippets are too big and belong in 30-hours-of-code. I support @kingdavidmartins 's decision to remove these snippets but I think we should first ask @Chalarangelo .
If he agrees I will close the PR

Chalarangelo:
Ah, hadn't seen the discussion when I reviewed. Yeah, I think these are pretty hard to grasp in 30 seconds or so. Based on the fact that we all agree that they are too long, I'm closing the PR.

PR #463, opened 2018-01-03
[FIX #419] remove max and min shims

### LINKED ISSUE #419: [DESIGN] Flatten and variadic inputs
I've noticed some snippets come in recently that start
```js
const fn = (...arr) => arr = [].concat(...arr), ...
```
which is a combination of the variadic adapter and a piped flatten before the function.

Should we allow these? While a cool trick, it also leads to a lot of code duplication in snippets.
It's also a situation where
```js
const fn = pipe(fn, flatten, collectInto)
```
should do the same thing and allow us to keep snippets smaller, less duplicated, and not have every snippet have a confusing footprint of `(*|Array<*>|...*)` and instead just be one of the three

Now don't get me wrong, the contributions have been great, I'm just curious if we want to duplicate flatten in those snippets or just make them variadic to match with Math.?

skatcat31:
## Description
**Resolves** #419 

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:


skatcat31:
Since we don't really want shim methods we want to remove these, especially since the same functionality can be achieved by
```js
pipe(collectInto(flatten), spreadOver(Math.max))
```

PR #470, opened 2018-01-03
Create factors.md

kriadmin:
- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/factors.md
+++ b/snippets/factors.md
@@ -18,21 +18,18 @@ const factors = (int,prime = false,powers = false) => {
   }
   const isPrime = num => {
   const boundary = Math.floor(Math.sqrt(num));
-  for (var i = 2; i <= boundary; i++) if (num % i == 0) return false;
+  for (var i = 2; i <= boundary; i++) if (num % i === 0) return false;
   return num >= 2 
   }
-  let array = []
-  const posFactor = num => {
-  let arr = []
-  for (let i = 2; i <= num; i++) if (num % i === 0) arr.push(i)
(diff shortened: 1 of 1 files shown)


kriadmin:
@Chalarangelo I am not sure I shorten it further and even If I am able to remember what `pretteir` did to `palindrome` 😆

Chalarangelo:
Good point, I'll see if I can shorten it by making some changes, but don't expect too much yet.

Chalarangelo:
@kriadmin I'd rather we converted this to `primeFactors` instead, so that we can shorten the code and make sure this works for just the most common use-case. As it currently stands, it is pretty slow and very hard to understand in 30 seconds or less.

kriadmin:
@Chalarangelo primeFactors is not good as it combine all the common use cases (I myself have felt the need for it two times this week while solving codewars). I think we should keep all of the existing functionality.

Chalarangelo:
Well maybe remove the `powers` part as it adds a lot of overhead to the method? Snippets should do one thing and one thing only, this does 3 right now (which could be two or three snippets in their own right).

kriadmin:
@Chalarangelo Ok we can remove the powers and prime part if you allow me to make seperate snippets for them 😄 Please 🙏

Chalarangelo:
Just the powers part is not necessary, the rest can stay as it is pretty simple and usable! 😉 

Already updated, but adding the comment for clarity in case someone else stumbles upon this discussion!

PR #479, opened 2018-01-03
Improve factors

kriadmin:
## Description

Added some extra functionality to factors

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/factors.md
+++ b/snippets/factors.md
@@ -39,7 +39,8 @@ const factors = (num, primes = false) => {
 
 ```js
 factors(12); // [2,3,4,6,12]
-factors(12, true); // [2,3]
+factors(12, 1); // [2,3]
+factors(12, -1); //[4,6,12]
 factors(-12); // [2, -2, 3, -3, 4, -4, 6, -6, 12, -12]
-factors(-12, true); // [2,3]
+factors(-12, 1); // [2,3]
 ```
--- a/snippets/factors.md
+++ b/snippets/factors.md
@@ -7,13 +7,13 @@ If `num` is less than `0` returns all the factors of `-int` together with their 
 
 Use `Array.from()`, `Array.map()` and `Array.filter()` to find all the factors of `num`.
 If given `num` is negative, use `Array.reduce()` to add the additive inverses to the array.
-Return all results if `primes` is `false`, else determine and return only the prime factors using `isPrime` and `Array.filter()`.
+Return all results if `primes` is `0`, only prime factors if `primes` is `1` & only the non-prime factors if `primes` is `-1`
 Omit the second argument, `primes`, to return prime and non-prime factors by default.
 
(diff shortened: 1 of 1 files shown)


kriadmin:
@Chalarangelo Even I can't think of any use-case but it might be useful at some places

Chalarangelo:
Well, while it *might* be useful, this makes the snippet longer and harder to read. Unless we can find valid use-cases, I vote against merging these changes as they harm readability.

kriadmin:
@Chalarangelo They add only like 3 lines of code which are easy to understand (easier than adapter functions).

skatcat31:
while an improvement I'm of the opinion that `factors` is not a function we should have. It's long Math Golf which is a way of saying "won't use outside of academia" and teaches little about Javscript. We have a few snippets like this that need to be addressed.

Really when we think of snippets we should begin to start thinking "What can an advanced beginner learn from this snippet" of which this snippet in and of itself teaches nothing but math. Mathematicians often make AWFUL programmers.
___
adapter functions, while a little esoteric, teach the interface and first order language functionality of javascript and become extremely useful in chains of transformations since you can use them to build your own libraries for your project especially with the prevalence of Promises and SPA.

kriadmin:
@skatcat31 I def agree with you that it teaches nothing new about Javascript. But it is definitely a useful snippet and is also easy to understand. We have many other math snippets and one more will not do any harm .

Chalarangelo:
@skatcat31 We have a few too many math snippets and I feel like there are practical uses for them, especially provided how popular JS is nowadays. Yes, it's slow when doing super complex mathematical things, but I have found myself needing things like factorization a few times and I suspect many students that come here while in college might benefit from them. The point here is that while not teaching a lot about JS, they might be the selling point to a student trying to whip up a project for college and that might make them get into JS in the first place. That, and the fact that something like Matlab takes forever to learn, so I would suspect a few people might take a look and think they are cool.

We are getting a lot of more academic snippets nowadays, so we might want to invent a category for them to de-clutter math in the long run. 

@kriadmin As it currently stands, I am not sold on the extra functionality and I don't think we should really consider adding it in, provided that these snippets might end up archived/deleted in the long run. I'm closing the PR and I might reopen when/if we decide on what to do with academia-focused snippets.

skatcat31:
I like the idea of moving Academic snippets out of Math maybe into an Academic category with a Math sub category.
___
As for the improvement aside from my feelings on the snippet:
This sort of improvement actively harms usability by having a hard to understand( albeit programatically faster ) API. I think a switch statement on a Boolean might have been a better idea with the cases being
```js
switch (primes) {
  case true:
    return array.filter(isPrime)
  case false:
    return array.filter(el => !isPrime(el))
  default:
    return array;
}
```

PR #480, opened 2018-01-03
[UPDATE] pluralize.md

atomiks:
## Description

- Allows for autopluralization with currying
- Allows any number. `-1` and `1` are considered "singular". Actually should -1 be plural? I don't really know. It doesn't make sense to pluralize negatives, but I feel like it shouldn't throw an error at all. It also coerces strings etc to a number to make life easier.

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/pluralize.md
+++ b/snippets/pluralize.md
@@ -18,7 +18,7 @@ pluralize(0, 'apple'); // 'apples'
 pluralize(1, 'apple'); // 'apple'
 pluralize(2, 'apple'); // 'apples'
 pluralize(1, 'person'); // 'person'
-pluralize(2, 'person', 'people'); // people
+pluralize(2, 'person', 'people'); // 'people'
 
 const PLURALS = {
   person: 'people',
--- a/snippets/pluralize.md
+++ b/snippets/pluralize.md
@@ -1,23 +1,29 @@
-# pluralize
+### pluralize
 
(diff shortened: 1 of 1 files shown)


Chalarangelo:
Pluralizing negatives is valid imo. I mean you might want to say minus that many things, so yeah, no problem with that.

snippets/pluralize.md
```diff
@@ -1,23 +1,29 @@
-# pluralize
+### pluralize
```
Chalarangelo: This was breaking the website before! 😆

kriadmin:
@Chalarangelo @atomiks I think should exist as an separate snippet 😕 . Also if `num` is 0 it should return a empty string. This is much more difficult to understand than the original one.

Chalarangelo:
@kriadmin I sincerely disagree, the snippet is barely any longer and it serves a very practical purpose.

skatcat31:
Is the auto pluralization needed?

atomiks:
@kriadmin `0` is a must.

For example, if you want to pluralize `0 votes` or `0 comments`, how would you do that? Those are very common in apps

@skatcat31 yes, if you want to import a dictionary of plurals and then just use the function once over and over without needing to think about the 3rd argument. Very useful imo

skatcat31:
Yeah I think I'm just thinking that the closure could somehow be avoided( probably not without greatly harming readability )

Issue #473, opened 2018-01-03
To Binary Tree or Not to Tree


kingdavidmartins:
### Description 

I believe adding other categories representing different operations/manipulation of other data structures other than arrays / objects can also be beneficial to the project.

I think its important for the following reasons: <strong>*exposure*</strong>, <strong>*growth*</strong>, <strong>*performance*</strong>, and <strong>*simplicity*</strong>

An example of this would be `Math.max()` which is native and takes variadic inputs 
But then we had `arrayMax()`
Then we made `max()` take both variadic & arrays. Which I regret & I partially take credit for see #295 
Then #419 happened
Now there's #463 

I honestly believe separating methods by their data structure would be best, and would prevent what was described above from repeating itself. Especially when looking at the following example 

For example let's say we want create a function that can find the maximum number from a particular data structure

#### max
```js
const max = (arr) => Math.max(...arr);
```
```js
max(Array); // Number
```

#### max
```js
const max = (Binary Tree) => code;
```
```js
max(Binary Tree); // Number
```

#### max
```js
const max = (Linked List) => code;
```
```js
max(Linked List); // Number
```

As you can see moving the project in this direction will not only prevent what was described with the following issues #295 | #419 | #463  from repeating them selves. But also keep snippetNames simple when adding to their respective categories.

Thoughts? @atomiks @skatcat31 @fejes713 @Chalarangelo

Chalarangelo:
First off, it's impossible to have multiple snippets with the same name, so we will have to still name them `arrayMax`, `binaryTreeMax` etc.

Secondly, data structure categories should be added but only as soon as we have enough snippets for each (5 sounds like a good starting amount). If someone can get started on coding snippets for each data structure and PRing them, I would be glad to get those categories up and running.

kingdavidmartins:
@Chalarangelo I def agree 

> First off, it's impossible to have multiple snippets with the same name, so we will have to still name them arrayMax, binaryTreeMax etc.

Haha a lot of times I am lazy & omit my full explanation. I just used `max()` as an example snippetName for each Category because I was lazy and copied and pasted 😆 

Didn't intend that we will actually have the same name, especially seeing that we have a package & would be conflicting AF when someone does `tsoc.max()`  🤣

Chalarangelo:
We are not `tsoc` anymore, we now are `_30s` (30 seconds of lodash seems like). 🤣

kingdavidmartins:
[...]

PR #484, opened 2018-01-03
[FEATURE] Binary tree snippets

fejes713:
## Description
Enough for the start. Two of the very basic snippets for working with binary trees. I've added emoji 🎄  and updated the tag docs. The tagger script capitalizes everything down but that could be changed before merging.

Also if someone has ideas on how to name the snippets, all ideas are welcome! 
If someone has time, these are on the list too: `BFS`, `DFS`, `contains`. If not I will do them in the few days. 
**Resolves** #473

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:


atomiks:
Having two functions will break the package builder btw...

snippets/insertIntoBST.md
```diff
@@ -0,0 +1,24 @@
+### insertIntoBST
+
+Inserts a new node into binary tree.
+
+Creates and inserts new node into existing tree With the help of `createBST` function.
+
+```js
+const createBST = value => ({value: value, right: null, left: null});
```
skatcat31: This feels like it's something we said we wanted to avoid in our own contribution guidelines isn't it?

snippets/createBST.md
```diff
@@ -0,0 +1,13 @@
+### createBST
+
+Returns an object which represents a binary search tree.
+
+Creates a new binary tree based on the passed argument which represents the root of a tree.
+
+```js
+const createBST = value => ({value: value, right: null, left: null});
```
skatcat31: Did we want to go with the Node Direct Structure implementation?

Chalarangelo:
Like I said in #473, I feel like we are taking a step in the wrong direction here. What I would prefer is one snippet for the data structure, something like this:

```js
const BST = () => ({
  create: val => ({value: value, right: null, left: null}),
  insert: val => /* logic that handles insertion here */,
  find: val => /*logic that handles search here */
});
```

That way we can quickly showcase methods about the data structure, keep the snippet short and avoid problems with the package, while not assuming prior knowledge.

skatcat31:
I think we should revisit this after #473 decisions regarding how to showcase the type itself

fejes713:
@skatcat31 Totally agree with you. This was just a test to see how it will go, but as @Chalarangelo  said: 
>I feel like we are taking a step in the wrong direction here. 

This is not 30 seconds of CS, this is 30 seconds of **useful** JS snippets. Binary trees and linked lists aren't things that somebody will use in their code.

skatcat31:
@feiwen8772 use in their code often*

PR #496, opened 2018-01-05
If using ES6, use template litorals

lesbaa:
Probably being pernickety.  But if we're using ES6, you could just use a template litoral instead of string concat? Not sure if the litoral version is a bit more difficult to read mind you.

## Description
use a string litoral instead of string concat.

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

- [x] My code follows the code style of this project.  (I think so)
- [x] I have read the **CONTRIBUTING** document. (very briefly)

### DIFF
--- a/snippets/digitize.md
+++ b/snippets/digitize.md
@@ -2,11 +2,11 @@
 
 Converts a number to an array of digits.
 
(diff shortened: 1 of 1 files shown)


atomiks:
The parentheses also become unnecessary with this change, so you can remove them

PR #498, opened 2018-01-05
[FEATURE] Add createEventHub

atomiks:
## Description
A tiny pubsub implementation

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:


snippets/createEventHub.md
```diff
@@ -0,0 +1,36 @@
+### createEventHub
+
+Creates a pubsub event hub with `emit`, `on`, and `off` methods.
```
Chalarangelo: `pubsub` might not be apparent to some readers, is there a better term?

tag_database
```diff
@@ -20,6 +20,7 @@ compose:function
 copyToClipboard:browser
 countOccurrences:array
 countVowels:string
+createEventHub:utility
```
Chalarangelo: Tags should be `browser,utility,advanced`.

tag_database
```diff
@@ -20,7 +20,7 @@ compose:function
 copyToClipboard:browser
 countOccurrences:array
 countVowels:string
-createEventHub:utility
+createEventHub:browser,utility,advanced
```
Chalarangelo: 👍

atomiks:
@skatcat31 in that case we can replace the plain object with a Map?

skatcat31:
@atomiks I feel like it went over your head possibly. What I meant is
```js
hub.emit('message', { hello: 'world'}); // this is valid in this implementation, would require type checking and throwing otherwise
```

atomiks:
😅 yeah I was confused

But, wait.. why should an object as the data throw an error? That allows you to access multiple data points from a single argument

skatcat31:
@atomiks that's not what I meant either. I just meant that you should note in the description/example that the event can be of any type subscribed to the topic.

As for the topic being an object I'm not certain I like that functionality from Map as it goes against the idea of being able to provide new instances of a string to trigger the event. The problem becomes if someone does
```js
emit(String('test'), 'test1');
emit(String('test'), 'test2');
```

skatcat31:
Right so some more clarification:

What I'm saying is that you should show the example
```js
hub.emit('message', {hello:'world'})
```
and note that the event is anything in the description. Sorry my last comment was a bit esoteric

atomiks:
I think events should only be strings, doesn't really make sense to be anything else? I made the example use an object now... not 100% sure what you mean tho

snippets/createEventHub.md
```diff
@@ -0,0 +1,39 @@
...
+  off(event, handler) {
+    const i = (this.hub[event] || []).findIndex(h => h === handler);
+    if (i > -1) this.hub[event].splice(i, 1);
+  }
+});
+```
+
+```js
+const fn = data => console.log(data.hello);
+const hub = createEventHub();
```
skatcat31: I'm guessing I'm not explaining it very well but what I was trying to get at is that the example should be
```js
const fn = data => console.log(data);
const hub = createEventHub();

hub.on('message', fn); // subscribe the handler to listen for 'message' events
hub.emit('message', { hello: 'world' }); // logs the object
hub.emit('message', 'hello world' }); // logs 'hello world'
hub.off('message', fn); // unsubscribe the handler from 'message' events
```

skatcat31: I've seen so many developer not understand that the type to the event handler is whatever is passed to the event. I'm not sure why I've seen it so many times, but I have enough that it should be here

skatcat31:
@Chalarangelo all of my concerns are met

Chalarangelo:
I'll review this again in the morning (mainly cleanup and a final test just in case) and merge. Seems ok to me!

PR #499, opened 2018-01-05
Add Plurals and improve pluralize

kriadmin:
- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:


mariastervic:
Isn't pluralize enough? You can use it with with a string like 
```js 
`2 ${pluralize(2,'day')}`
```
and you would not need anything more than that. Also, the regex makes the snippet hard to read and handles far too many cases that might not be necessary if you are pluralizing in another language, apart from English. The dictionary is enough I think to make sure `pluralize` works for any string and people can specify what they need for their own language/enviroment.

Chalarangelo:
@mariastervic That is a fair point. We are usually only thinking about english, but if I used this on Greek that would cause a whole host of problems for me. Adding a regex that applies only for a specific linguistics group is not the best idea definitely and only harms readability for a useful snippet with a good idea. 

`plurals` is redundant as you can use `pluralize` with a `curry` to actually achieve the same:

```js
const plurals = curry(pluralize)(2);
plurals('day'); // 'days'
```

For these reasons, I am closing as these changes are not really useful and can be done by composing other snippets without harming readability.

PR #500, opened 2018-01-05
Add.binary search

kingdavidmartins:
```js
const binarySearch = (arr, val, start = 0, end = arr.length - 1) => {
  if (start > end) return -1;
  const mid = Math.floor((start + end) / 2);
  const target = arr[mid];
  return (target > val) 
      ? binarySearch(arr, val, start, mid - 1)
      : (target < val) 
        ? binarySearch(arr, val, mid + 1, end)
        : mid
}
```

```js
binarySearch([1, 4, 6, 7, 12, 13, 15, 18, 19, 20, 22, 24], 6); // 2
binarySearch([1, 4, 6, 7, 12, 13, 15, 18, 19, 20, 22, 24], 21); // -1
```


snippets/binarySearch.md
```diff
@@ -0,0 +1,23 @@
+### binarySearch
+
+Use recursion to perform a logarithmic search similar to `.indexOf()` that finds the index of a value within an array. The two differences being
```
Chalarangelo: `Array.indexOf()`

snippets/binarySearch.md
```diff
@@ -0,0 +1,23 @@
+### binarySearch
+
+Use recursion to perform a logarithmic search similar to `.indexOf()` that finds the index of a value within an array. The two differences being 
+1. Operation only works with sorted arrays
```
Chalarangelo: Only arrays sorted in ascending order, if I understand this correctly.

snippets/binarySearch.md
```diff
@@ -0,0 +1,23 @@
...
+2. Offers a major performance boost when compared to a linear search or `.indexOf()`
```
Chalarangelo: You need to add an explanation on how the snippet works. Remember to say at the very top `Uses recursion`. and then explain the rest of the code. It's pretty simple, but it needs to be explained for readers to grasp.

snippets/binarySearch.md
```diff
@@ -0,0 +1,23 @@
...
+
+Use recursion to perform a logarithmic search similar to `.indexOf()` that finds the index of a value within an array. The two differences being 
+1. Operation only works with sorted arrays
+2. Offers a major performance boost when compared to a linear search or `.indexOf()` 
+
+```js
+const binarySearch = (arr, val, start = 0, end = arr.length - 1) => {
+  if (start > end) return -1;
+  const mid = Math.floor((start + end) / 2);
+  const target = arr[mid];
```
Chalarangelo: You might as well not assign `target` and use it right away in the conditions below, makes no difference performance-wise.

snippets/binarySearch.md
kingdavidmartins: Yea, about to do push updated description

snippets/binarySearch.md
kingdavidmartins: Yea either or, can be ascending or sorted

atomiks:
Since we use the comma operator for declarations in the snippets (which I don't really like but for snippets it's ok I guess):

```js
const binarySearch = (arr, val, start = 0, end = arr.length - 1) => {
  if (start > end) return -1;
  const mid = Math.floor((start + end) / 2),
    target = arr[mid];
  return target > val || target < val
    ? binarySearch(arr, val, target > val ? start : mid + 1, target > val ? mid - 1 : end)
    : mid;
};
```

snippets/binarySearch.md
kingdavidmartins: Yea, But trying to have readability on par with 30 seconds or less. Feel like when user reads the updated description and see snippet they can easily & quickly figure out which things represent which

Chalarangelo: @atomiks made a nice suggestion about this, which would not harm readability and shorten the code just about enough. I like it, what do you think about it?

kingdavidmartins:
@atomiks It impacts readability negatively though. Doesn't flow & isn't as simple. At least for 30 seconds. Or am I the only that think this?

atomiks:
The double ternary looks kinda weird as well, for readability I would do `if` + `return` but this repo seems to favor shortness over all else so...

kingdavidmartins:
@Chalarangelo Want to confirm before pushing. Which change is preferred

```js
const binarySearch = (arr, val, start = 0, end = arr.length - 1) => {
  if (start > end) return -1;
  const mid = Math.floor((start + end) / 2);
  if (arr[mid] > val) return binarySearch(arr, val, start, mid - 1);
  if (arr[mid] < val) return binarySearch(arr, val, mid + 1, end);
  return mid;
}
```

### Or

```js
const binarySearch = (arr, val, start = 0, end = arr.length - 1) => {
  if (start > end) return -1;
  const mid = Math.floor((start + end) / 2);
  return (arr[mid] > val) 
      ? binarySearch(arr, val, start, mid - 1)
      : (arr[mid] < val) 
        ? binarySearch(arr, val, mid + 1, end)
        : mid
}
```

Chalarangelo:
@kingdavidmartins To me, the second one is easier to read, but that might be personal preference.

iamsoorena:
to me first one is far more readable.

skatcat31:
For this one I prefer first because of the way the ternary makes it a little confusing because of line length and indentation for the particular snippet. It doesn't line up well making it hard to notice the differences between the almost same lines

skatcat31:
If we made the ternary line up I'd be indifferent

kingdavidmartins:
Since the 1st one seems to garner more favor from @atomiks due to readablity from `if` + `return`, @iamsoorena, & @skatcat31. I'll update the snippet with the 1st suggested change

skatcat31:
Why is this still in a Changes Requested state?

Chalarangelo:
@skatcat31 Because I totally forgot about it and nobody else seemed to care I suppose. I will merge tomorrow and archive as it falls under the same category as `quickSort` and other algorithmic re-implementations of native ES6 methods, but is nice to have anyways.

PR #503, opened 2018-01-06
[FEATURE] Add chunksArray.md

kriadmin:
## Description
It might look like like `chunk` snippet but it is different. I couldn't write a great description sorry 😞 . Please someone look at the examples and improve the description.
## What does your PR belong to?
- [ ] Website
- [x] Snippets
- [ ] General / Things regarding the repository (like CI Integration)

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (non-breaking improvement of a snippet)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have checked that the changes are working properly
- [x] I have checked that there isn't any PR doing the same
- [x] I have read the **CONTRIBUTING** document.

PR #505, opened 2018-01-06
[FEATURE] indexesOf

kriadmin:
- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:


kriadmin:
@skatcat31 Updated it 😄

atomiks:
`indices` is the correct plural, I think

kriadmin:
@atomiks This is what comes up on googling .
> Both "indexes" and "indices" are acceptable plural forms of the word "index" or to refer to more than one index. 
But if you want I can update it

kingdavidmartins:
I believe naming the snippet something like `indexOfAll` goes a long way especially since it's similar to `Array.indexOf` and it indicates that it returns all the indices found

atomiks:
With reduce:
```js
const indicesOf = (arr, val) => {
  const indices = arr.reduce((acc, el, i) => (el === val ? acc.concat(i) : acc), []);
  return indices.length === 0 ? [-1] : indices;
};
```

But I bet a plain for loop would be most performant:
```js
const indicesOf = (arr, val) => {
  const indices = [];
  for (let i = 0; i < indices.length; i++) if (arr[i] === val) indices.push(i);
  return indices.length === 0 ? [-1] : indices;
};
```

kriadmin:
@atomiks https://jsperf.com/testing-foreach-vs-for-loop see this jsperf comparison of forEach and for loop

kriadmin:
@kingdavidmartins I think returning [-1] is a more consistent API as suggested by @skatcat31

kingdavidmartins:
Ahh gotcha. Jumped to review the code didn't review the comments yet. Makes sense.

atomiks:
`arr.forEach((el,i) => {if(el === val) indexes.push(i)})`

Prettier will reformat this, you can use short circuiting to keep it one line:

`arr.forEach((el, i) => el === val && indexes.push(i))`

kingdavidmartins:
So looks like it is most likely going to be one of the following

```js
const indexOfAll = (arr, val) => {
  const indices = [];
  arr.forEach((el, i) => el === val && indices.push(i))
  return indices.length === 0 ? [-1] : indices;
};
```

### OR

```js
const indexOfAll = (arr, val) => {
  let allIndex = arr.map((v, i) => v === val ? i : false).filter(x => x !== false);
  return allIndex.length === 0 ? [-1] : allIndex;
}
```

### OR

```js
const indexOfAll = (arr, val) => {
  const indices = arr.reduce((acc, el, i) => (el === val ? acc.concat(i) : acc), []);
  return indices.length === 0 ? [-1] : indices;
};
```

<strong>*Note:*</strong> 
- Not a fan of the 3rd snippet due to `.concat()` which isn't very performant 
- Feel like the 2nd snippet is more readable
- Believe the 1st is more performant than the others

atomiks:
2nd one is O(2n) isn't it? The first is O(n)

Why not do this btw? Since `0` is falsy.

`return indices.length ? indices : [-1];`

kriadmin:
@kingdavidmartins I think the first one is more readable.  I vote for the first one.

kingdavidmartins:
@atomiks Yea, agreed

> 2nd one is O(2n) isn't it? The first is O(n)

However the following wouldn't work if `0` get removed

> Why not do this btw? Since 0 is falsy.
```js
return indices.length ? indices : [-1];
```

kingdavidmartins:
Which snippet version do you want to vote for @atomiks? I'm honestly indifferent when it comes to the 1st & 2nd. So which ever snippet you vote for, I will vote in agreement. So the update can be made

kingdavidmartins:
Also there's still also duplicate commits happening @kriadmin as pointed out by me in PR #503

kriadmin:
@kingavidmartins Yes I know. Check my comment on #503. :(

atomiks:
I vote for this

```js
const indexOfAll = (arr, val) => {
  const indices = [];
  arr.forEach((el, i) => el === val && indices.push(i))
  return indices.length ? indices : [-1];
};

**Comment by kriadmin** [comment]
@kingdavidmartins I updated it but I couldn't check if they work due to being on phone . please check them

**Comment by Chalarangelo** [comment]
I made some improvements to the snippet description and I will merge. But, on a final note, there is another technique which we *could have* used - utilizing `Array.indexOf()`'s second argument to quickly loop over the array:

```js
const indexOfAll = (arr, val, start = 0) => {
  const i = arr.indexOf(val, start);
  if(i !== -1) return [i].concat(indexOfAll(arr,val,i+1));
  else return start === 0 ? i : [];
};
```

I tested this with a [jsPerf](https://jsperf.com/indexofall-505) (Suggestion) and it is quite a lot slower. Hoever, that is the fault of `Array.concat()`, so I tried with a helper function, too, and it is marginally faster:

```js
const indexOfAll = (arr, val) => {
  const indices = [];
  const _indices = (_arr, _val, start = 0) => {
    const i = _arr.indexOf(_val, start);
    if(i !== -1) {
      indices.push(i);
      _indices(_arr,_val,i+1);
    }
  }
  _indices(arr,val);
  return indices.length? indices : [-1];
};
```

While this provides a tiny speed improvement, it significantly harms readability, so I'd rather we keep the implementation we have. Just wanted to share what I tried, if anyone can use it for something else. 😄

snippets/indexOfAll.md
```diff
@@ -0,0 +1,20 @@
...
+Returns all indices of `val` in an array. If `val` never occurs, returns `[-1]`.
+
+Use `Array.forEach()` to loop over elements and `Array.push()` to store indices for matching elements.
+Return `[-1]` if `length` of the array of indices is `0`, otherwise return the array of indices.
+
+``` js
+const indexOfAll = (arr, val) => {
+  const indices = [];
+  arr.forEach((el, i) => el === val && indices.push(i))
+  return indices.length ? indices : [-1];
```
skatcat31: I think you misunderstood me:
```js
return indices.length ? indices : -1;
```

skatcat31: This way you either get back an array, in which case do what you need to, or a -1, no third case. This is a consistent testable API in that Array.isArray will gladly work on.

I just meant you should return either the array of indeces or the -1, not a singular value if there is only one index

Chalarangelo: I think returning an array in all cases feels more consistent, as you can then use it as an array no matter what. Might be just me, though.

skatcat31: currently
```js
indexOfAll([1,2,3], 4).filter(a => a != -1).map(console.log)
```

If this is the case, we should return an empty array, not an array of -1, otherwise the filter would be needed for any chaining

Chalarangelo: @skatcat31 An empty array is a decent idea that improves the snippet, could you send a PR for that? (I'm on mobile currently)

skatcat31: @Chalarangelo I'm in  the middle of deploying and have time sensitive issues after, so I will not be able to( don't have this repo on my work PC )

Chalarangelo: You can edit the snippet right from the browser, @skatcat31. The main reason I suggest this is so that you can take credit for the change, which was your idea to begin with! 😉

skatcat31: ... I meant to just make a PR but apparently I edited it directly on master... Web apps are confusing -.-

I can't figure out how to bring my side along fork to date and create an MR from it. Is what I did an issue?( this was due to a lack of knowledge of how Github does it's built in tools )

Chalarangelo: Nop, it's a change discussed and approved, no problem at all!

PR #511, opened 2018-01-08
[FEATURE] Add reject

Chalarangelo:
## Description
**Lodash** #100 -> https://lodash.com/docs/4.17.4#reject

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/reject.md
+++ b/snippets/reject.md
@@ -1,0 +1,17 @@
+### reject
+
+Returns the elements of an array for which the passed function returns `false`.
+
+Use `Array.filter()` and the not operator (`!`) to return the elements in the array for which `fn` returns `false`.
+
+```js
+const reject = (arr, fn) => arr.filter((e,i,arr) => !fn(e,i,arr));
+```
+
+```js
+const users = [
+  { 'user': 'barney', 'age': 36, 'active': false },
+  { 'user': 'fred',   'age': 40, 'active': true }
+];
+reject(users, o => !o.active); // [{ 'user': 'fred',   'age': 40, 'active': true }]
+```
--- a/tag_database
+++ b/tag_database
@@ -128,6 +128,7 @@ randomNumberInRange:math,utility,random
 readFileLines:node,array,string
 redirect:browser,url
 reducedFilter:array
+reject:array,object,function
 remove:array
 reverseString:string,array
 RGBToHex:utility


snippets/reject.md
```diff
@@ -0,0 +1,17 @@
+### reject
+
+Returns the elements of an array for which the passed function returns `false`.
+
+Use `Array.filter()` and the not operator (`!`) to return the elements in the array for which `fn` returns `false`.
+
+```js
+const reject = (arr, fn) => arr.filter((e,i,arr) => !fn(e,i,arr));
```
skatcat31: breaks binding of context provision of the filter method

Chalarangelo: Meaning `this` is broken in there?

skatcat31: I can't think of a single use case where I'd use this over filter with a proper predicate. If someone is using boolean logic and can't design a proper predicate, what use would this function be for them?

Chalarangelo: I only implemented this because lodash has it and, to be honest, I felt the same. Should we just can this and mark in #100 to note we are not implementing?

Chalarangelo: We also have `negate` which should be enough with composition, let's kill this one then.

iamsoorena:
I just approved that it works right. I have no suggestion about whether or not we should add this. :)

snippets/reject.md
skatcat31: the awkward part with negate is that it also kills the bindable, and does exactly what this snippet is with composition:
```js
const reject = (f, a) => a.filter(negate(f)) // kills bindable unless negate is implemented as a function versus an arrow function
```

Chalarangelo: In all honesty these start to get confusing and both are terribly niche and forgettable. I think we shall keep `negate` just as an example (not a great one, but still) and remove `reject` and any other methods that are just boolean mumbo-jumbo. How difficult is it for a developer to add a `!` anyways?

Chalarangelo:
I'm closing and marking in #100 that this will not be implemented. Might mark a few more as well.

snippets/reject.md
mariastervic: Lodash has a lot of fluff methods that are not useful. It's better to avoid this and a few others such as the `flattenMap` ones in Object.

PR #512, opened 2018-01-08
[FEATURE] Add partition

Chalarangelo:
## Description
**Lodash** #100 -> https://lodash.com/docs/4.17.4#partition

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:

### DIFF
--- a/snippets/partition.md
+++ b/snippets/partition.md
@@ -7,7 +7,7 @@ Use `Array.push()` to add elements for which `fn` returns `true` to the first ar
 
 ```js
 const partition = (arr, fn) =>
-  arr.reduce((acc, val, i, arr) => {fn(val,i,arr) ? acc[0].push(val) : acc[1].push(val); return acc;},[[],[]]);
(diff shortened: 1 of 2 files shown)


Chalarangelo:
@atomiks I thought about it when I was writing the snippet, it looks slightly confusing like that, I feel. Plus we have a few snippets with the same kind of style, so I decided to go with that style for consistency.

atomiks:
@Chalarangelo see the edit- you can also stick the ternary inside the `acc[]`, which makes it shorter

Chalarangelo:
`Array.concat()` is terriby slow, so we avoid using it for such cases in general. Sticking the ternary inside the `[]` is not a bad idea, however the comma operator next to that will just be a bit too much and the rest of the snippets don't use it anyways, so I guess just the ternary should be enough of an improvement.

skatcat31:
Reduce is a weird monster, but this works as advertised. To only awkward part is needing the parens, which harms readability but not enough to not consider it

PR #518, opened 2018-01-09
[FEATURE] noVowels.md

kriadmin:
## Description
add noVowels.md .  Can we have `repl` to be function?

- [ ] Website
- [x] Snippets

## Types of changes

## Checklist:


Chalarangelo:
What valid use-cases does this snippet have? I don't think I have ever needed this.

fejes713:
I got this as kata on code wars and I was like I am going to add this and then I taught:
> Who is even going to use this?

kriadmin:
@Chalarangelo @fejes713 @kingdavidmartins It is a sibling of `countVowels`. And if we replace `repl` with a   `fn` than it's use cases might increase 😕

Chalarangelo:
I'm not sure `countVowels` is super useful either, unless you are doing something super specific in a string.

kriadmin:
@Chalarangelo @fejes713 @kingdavidmartins 
Than we can have a poll for both of them.
-If they win they go in snippets
-If they lose they go in snippets_archive

Voting:
👍 : in favor of snippets
👎 : not in favor of snippets

Ok?

fejes713:
> something super specific in a string
Like counting vowels maybe 😆  @Chalarangelo

kingdavidmartins:
Although development specific, I believe `countVowel` has use when developing string/word based games. Where the scoring system treats the string/word vowels higher or lower based on number of vowels found

I think this snippet is redundant since one can use countVowel to check if a string has noVowels by checking if `countVowel(str) === 0`

kriadmin:
@kingdavidmartins Did u actually look at the code and what it does 😒

atomiks:
The name should be `replaceVowels` in that case

Chalarangelo:
Both are novelty and should be archived as far as I can tell. All they do is use the simplest of regexes to count something in a string or replace it. I haven't seen many use-cases outside of games or very specific applications, so I think we'd better send them to the archive and be done with them.

kingdavidmartins:
> Both are novelty and should be archived as far as I can tell. All they do is use the simplest of regexes to count something in a string or replace it. 

@Chalarangelo to be fair what you/us might reason as simple. Might not be for others

Chalarangelo:
@kingdavidmartins I agree with your statement, but this is a regex you can very easily find if you need and, as we already have discussed, the use-cases *might* be very limited anyways. Having them in the archive makes sure people can still find them, but de-clutters the main page, so that people can find the better snippets more easily.

kingdavidmartins:
@kriadmin Yea, I did. I just assumed you must up the description and was going to suggest you return `true` or `false` if it has `noVowels` but then rather followed through with the following

> I think this snippet is redundant since one can use countVowel() to check if a string has noVowels by checking if countVowel(str) === 0

Which I believe was vise versa. You actually messed up on the naming & not the description and implementation.

@Chalarangelo makes sense

PR #536, opened 2018-01-12
4 New Snippets

Emgo-Dev:
Maintain Digits
Convert to 12 hour format
Format Colon Time
Get Meridiem Time

## Description

**Resolves** #(issue number) 

**Lodash[ACTION]** #100 -> https://lodash.com/docs/4.17.4#(method) 

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

- [ ] I added the prefix [UPDATE: `method.md`] or [ADD: `method.md`]

## Checklist:


Emgo-Dev:
I see, sorry the guidelines changed a lot in only a few weeks. I think I made all the necessary changes. Can you take a look now?

Snippet file name, title, and function name all match in camelCase. Properly wrote descriptions, removed edge cases from functions, removed any dependencies in functions (copied the code from two other submitted snippets into one that requires them to work).

snippets/getColonTime.md
```diff
@@ -0,0 +1,13 @@
...
+Returns transformed time integers from provided date object to "colon time" representation. Formats hour to 12 hour and maintains digit length of colon time format (HH:MM:SS).
+
+```js
+let toColonTime = ( date ) => {
+    let times = [ date.getHours(), date.getMinutes(), date.getSeconds() ];
+    times[0] = times[0] === 0 || times[0] === 12 || times[0] === 24 ? 12 : times[0] % 12;
+    
+    return times.map( time => time.toString().padStart(2, "0") ).join(":");
+}
+// toColonTime12(new Date()) -> "08:38:00"
```
skatcat31: The examples should be in a different js block. Please update your snippets to match

# Example

> ### toColonTime
> 
> Returns transformed time integers from provided date object to "colon time" representation. Formats hour to 12 hour and maintains digit length of colon time format (HH:MM:SS).
> ```js
> const toColonTime = ( date ) => {
>     let times = [ date.getHours(), date.getMinutes(), date.getSeconds() ];
>     times[0] = times[0] === 0 || times[0] === 12 || times[0] === 24 ? 12 : times[0] % 12;
>     
>     return times.map( time => time.toString().padStart(2, "0") ).join(":");
> }
> ```
> ```js
> toColonTime12(new Date()) -> "08:38:00"
> ```

snippets/getMeridiemOfInteger.md
```diff
@@ -0,0 +1,15 @@
+### toMeridiem
+
+Uses modulo operator (`%`) to transform an integer to a 12 hour clock format. Stringifies transformed integer with concatenated meridiem suffix. Maintains 12 hour format principles with conditional check (0am - 12am).
+
+```js
+const toMeridiem = num => {
+    const meridiems = ["am", "pm"];
+    let period = num > 11 ? 1 : 2;
```
skatcat31: Using an array and a reference is error prone if there are few cases. Being explicit here is better:
```js
let period = num > 11 ? 'pm' : 'am'
```

As a case in point: JS uses 0 index, you used 1 index leading to either `XXam` or `XXundefined`

snippets/maintainDigits.md
```diff
@@ -0,0 +1,14 @@
+### setDigits
+
+Uses `String.padStart()` or `String.padEnd()` to maintain digit length of a provided integer. Returns stringified integer.
+
+Control direction (0 = before integer, 1 = after integer).
+
+```js
+const setDigits = ( num=0, len=1, direction=0 ) => {
```
skatcat31: This API is extremely confusing. They should be required to pass a number.

snippets/maintainDigits.md
```diff
@@ -0,0 +1,14 @@
...
+    return direction > 0 ? num.toString().padStart(len, "0") : num.toString().padEnd(len, "0");
```
skatcat31: Since this is a single line you can use the implicit return:
```js
const example = _ => true
```
```js
example() // true
```

skatcat31: `padStart` and `padEnd` are not in ES6 but ES2017. That should be noted in the description, even though it is supported in most browsers

snippets/to12HourFormat.md
```diff
@@ -0,0 +1,13 @@
+### to12Hour
+
+Uses the modulo operator (`%`) to transform an integer to a 12 hour clock format.
+
+```js
+const to12Hour = ( num ) => {
+
+    return num === 0 || num === 12 || num === 24 ? 12 : num % 12;
```
skatcat31: can be rewritten with implicit return

skatcat31:
As another note I notice a couple of these are date formatting. @Chalarangelo @kingdavidmartins @Pl4gue @atomiks would like some input on them being included.

We should start a discussion on the inclusion of date formatting snippets. Regardless we should at least move these date formatting snippets to the archive due to their ubiquity and constant use if nothing else.

snippets/maintainDigits.md
```diff
@@ -0,0 +1,14 @@
+### setDigits
```
Chalarangelo: This is an extremely confusing method and doesn't provide any real value, as you can achieve the exact same thing with native methods. The returned result could be easily unusable in many situations and there could be tons of problems using this in anything (try passing an array as input - `setDigits([10,2],5,1)` returns `"010,2"` which is highly confusing and should not work at all). I'm voting to remove the snippet entirely and move on.

snippets/getMeridiemOfInteger.md
```diff
@@ -0,0 +1,15 @@
+### toMeridiem
```
Chalarangelo: As stated in #538 I suggest this and `to12hour` are merged into one snippet that converts to 12h format with period appended optionally. Something like this:
```js
const to12h = (hours, period = true) =>
  (hours% 12 === 0 ? 12 : hours% 12) + (period ? hours> 11 ? 'pm' : 'am' : 0); 
```

snippets/getMeridiemOfInteger.md
Emgo-Dev: You're right, that slipped my attention. I rewrote how it quickly to keep it simpler. I'll fix that.

snippets/getMeridiemOfInteger.md
Emgo-Dev: Sure if you'd prefer that. I wrote most of these functionally so they'd call one another when that action was needed, but I removed their dependencies when submitting them.

snippets/maintainDigits.md
Emgo-Dev: You're right. I wrote it in mind with maintaining digit length, for example in the colon time format. Like I'm mentioning above these were written dependent one another in some ways. This can also be used to maintain digit length of a real price rather than having it rounded to the nearest tenth cent.

Just wanted to throw them all and see what stuck haha. Thanks for your review everyone.

snippets/maintainDigits.md
Emgo-Dev: Should I rewrite it without relying on those instead?

snippets/getColonTime.md
Emgo-Dev: I've realized I can reduce this function to
```js
const getColonTimeFromDate = date => date.toTimeString().slice(0, 8)
```

Not sure if it's worth submitting at all.

Emgo-Dev:
Just took a look at #538.

From your specifications for date formatting functions it seems the previous commit of ```getColonTimeFromDate()``` would be preferred which uses individual methods ```Date.getHours()```, ```Date.getMinutes()``` rather than ```String.slice()``` to select only the colon time from ```Date.toTimeString()```.

> @Chalarangelo in #538 
> toMeridiem and to12Hour should be merged to provide a back and forth converter (something like convertTime). 

I can't find an example of these functions. I know ```toCapitalize``` has a parameter that controls for forcing all following characters to be lowercase, which is what I think you mean.

snippets/getColonTimeFromDate.md
```diff
@@ -0,0 +1,11 @@
+### getColonTimeFromDate
+
+Formats hour, minute, and second time integers from Date into stringified colon representation. Formats hour integer to 12 hour clock and maintains digit lengths to match (HH:MM:SS).
```
kingdavidmartins: Hey @Emgo-Dev 👋 😄 

Thanks for taking the time out to submit this PR

I believe `new Date()` by default returns 24 hrs & not 12

snippets/getColonTimeFromDate.md
```diff
@@ -0,0 +1,11 @@
...
+
+```js
+const getColonTimeFromDate = date => date.toTimeString().slice(0, 8)
```
kingdavidmartins: Hey @Emgo-Dev 👋 😄 

Also I believe going about the following makes things much easier

Changing the method name to `digitalClock24`

```js
const digitalClock24 = _ => new Date().toTimeString().slice(0, 8);
```

Which can now be invoked by doing the following

```js
digitalClock24() // "13:07:00"

```

snippets/getColonTimeFromDate.md
Emgo-Dev: Yes it does which detracts from the simplicity of just slicing the colon time from the ```Date``` itself, because formatting the hour would require more slicing work.

It comes down to the preferences of the repo. Slicing a provided string serves little educational purpose as well, where constructing a string has more to look at.

snippets/getColonTimeFromDate.md
Emgo-Dev: Yeah that makes a lot more sense. I think my original terminology was a bit more accurate too. I was following the changes made to my toOrdinal() submission. The file name was changed to get ordinal suffix from integer. So I just followed that.

snippets/getColonTime.md
tomByrer: Worth submitting IMHO; some don't realize one can just slice.
Keep in mind that many program JavaScript who are still novices; either new to programming or program other languages.  Having a copy/paste reference is handy!

snippets/maintainDigits.md
tomByrer: Just old mobile & IE11 users can't use pad; they can use a polyfill.

snippets/getColonTime.md
Emgo-Dev: I had originally preferred constructing the colon time string from individual values because it allowed you to format those values beforehand such as formatting the hours to 12 hour format. To format this you'd need to format the result of the function. I felt it was more functional the other way, but this would still work. I imagine this function would be an 'endpoint' though, so rather than have this function encapsulate parameters being formatted, you'd format the result of the function.

PR #540, opened 2018-01-13
[FEATURE] Add colorize

Chalarangelo:
## Description

Output colorization in node console is quite common, so I tried my hand on it. Should work cross-platform. 

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/colorize.md
+++ b/snippets/colorize.md
@@ -1,0 +1,35 @@
+### colorize
+
+Add special characters to text to print in color in the console (combined with `console.log()`).
+
+Use template literals and special characters to add the appropriate color code to the string output.
+For background colors, add a special character that resets the background color at the end of the string.
+
+```js
+const colorize = (...args) => (
+  {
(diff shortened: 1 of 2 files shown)


snippets/colorize.md
```diff
@@ -0,0 +1,35 @@
+### colorize
+
+Add special characters to text to print in color in the console (combined with `console.log()`).
+
+Use template literals and special characters to add the appropriate color code to the string output.
+For background colors, add a special character that resets the background color at the end of the string.
+
+```js
+const colorize = (...args) => (
+  {
```
skatcat31: While it will get garbage collected, this will cause a memory space for every log of `colors * string`. In high performance applications this is a bad way of doing it due to the large amount of memory high traffic will consume. Each color or option should be it's own function and it should just be a micro library so that there are no excess memory allocations.

PR #542, opened 2018-01-13
[FEATURE] Add observeMutations.md

Chalarangelo:
## Description

A method that observes mutations in the DOM and runs a callback.

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/observeMutations.md
+++ b/snippets/observeMutations.md
@@ -1,23 +1,27 @@
 ### observeMutations
 
-Runs the provided callback for each mutation on the specified element.
+Returns a new MutationObserver and runs the provided callback for each mutation on the specified element.
 
 Use a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to observe mutations on the given element.
 Use `Array.forEach()` to run the callback for each mutation that is observed.
 Omit the third argument, `options`, to use the default [options](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#MutationObserverInit) (all `true`).
 
 ```js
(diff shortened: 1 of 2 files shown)


snippets/observeMutations.md
```diff
@@ -0,0 +1,23 @@
+### observeMutations
+
+Runs the provided callback for each mutation on the specified element.
+
+Use a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to observe mutations on the given element.
+Use `Array.forEach()` to run the callback for each mutation that is observed.
+Omit the third argument, `options`, to use the default [options](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#MutationObserverInit) (all `true`).
+
+```js
+const observeMutations = (element, callback, options) => new MutationObserver(mutations => mutations.forEach(m => callback(m))).observe(element, Object.assign(
```
atomiks: Should we start using the spread syntax? It kinda makes this easier. I know it's not part of ES yet but it should be soon.

You might also want to return the reference to the mutation observer to allow it to be disconnected.

```js
const observeMutations = (el, callback, options) => {
  const observer = new MutationObserver(mutations => mutations.forEach(callback));
  observer.observe(el, {
    childList: true,
    attributes: true,
    attributeOldValue: true,
    characterData: true,
    characterDataOldValue: true,
    subtree: true,
    ...options
  });
  return observer;
};
```

Chalarangelo: As long as it's not part of ES, we'd better stick to `Object.assign()`. It's not too hard to read anyways. About returning the mutation observer, you are absolutely right, that would be better, thanks for the suggestion, I'll update the code.

PR #552, opened 2018-01-17
[FEATURE] Hash methods

Chalarangelo:
## Description

Added two snippets for generating hashes from given values/objects (one for browser, one for Node). The Browser one might be slightly advanced, but I think they are both quite valuable.

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/hashBrowser.md
+++ b/snippets/hashBrowser.md
@@ -18,5 +18,5 @@ const hashBrowser = val =>
 ```
 
 ```js
-hashBrowser(JSON.stringify({a :'a', b: [1,2,3,4], 'foo': {c: 'bar'}})).then(hash => console.log(hash)); // '04aa106279f5977f59f9067fa9712afc4aedc6f5862a8defc34552d8c7206393'
+hashBrowser(JSON.stringify({a :'a', b: [1,2,3,4], 'foo': {c: 'bar'}})).then(console.log); // '04aa106279f5977f59f9067fa9712afc4aedc6f5862a8defc34552d8c7206393'
 ```
--- a/snippets/hashBrowser.md
+++ b/snippets/hashBrowser.md
@@ -1,0 +1,22 @@
+### functionName
+
+Creates a hash for a value using the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) algorithm. Returns a promise.
+
+Use the [SubtleCrypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) API to create a hash for the given value.
+
+```js
+const hashBrowser = val =>
+  crypto.subtle
+    .digest('SHA-256', new TextEncoder('utf-8').encode(val))
+    .then(h => {
+      let hexes = [],
+        view = new DataView(h);
+      for (let i = 0; i < view.byteLength; i += 4)
+        hexes.push(('00000000' + view.getUint32(i).toString(16)).slice(-8));
+      return hexes.join('');
+    });
+```
+
+```js
+hashBrowser(JSON.stringify({a :'a', b: [1,2,3,4], 'foo': {c: 'bar'}})).then(hash => console.log(hash)); // '04aa106279f5977f59f9067fa9712afc4aedc6f5862a8defc34552d8c7206393'
+```
--- a/snippets/hashBrowser.md
+++ b/snippets/hashBrowser.md
@@ -1,4 +1,4 @@
-### functionName
+### hashBrowser
 
 Creates a hash for a value using the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) algorithm. Returns a promise.
 
--- a/snippets/hashNode.md
+++ b/snippets/hashNode.md
@@ -1,0 +1,14 @@
+### functionName
+
+Creates a hash for a value using the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) algorithm.
+
+Use `crypto` API to create a hash for the given value.
+
+```js
+const crypto = require('crypto');
+const hashNode = val => crypto.createHash('sha256').update(val).digest('hex');
+```
+
+```js
+hashNode(JSON.stringify({a :'a', b: [1,2,3,4], 'foo': {c: 'bar'}})); // '04aa106279f5977f59f9067fa9712afc4aedc6f5862a8defc34552d8c7206393'
+```
--- a/snippets/hashNode.md
+++ b/snippets/hashNode.md
@@ -1,14 +1,15 @@
 ### hashNode
 
-Creates a hash for a value using the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) algorithm.
+Creates a hash for a value using the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) algorithm. Returns a promise.
 
 Use `crypto` API to create a hash for the given value.
 
 ```js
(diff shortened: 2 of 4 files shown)


snippets/hashBrowser.md
```diff
@@ -0,0 +1,22 @@
+### functionName
```
kingdavidmartins: Update snippetTitle to `hashBrowser` currently says `functionName`

snippets/hashNode.md
```diff
@@ -0,0 +1,14 @@
+### functionName
```
kingdavidmartins: Update snippetTitle to `hashNode` currently says `functionName`

Chalarangelo:
@kingdavidmartins Fixed both, hadn't noticed that, thanks for pointing it out!

kingdavidmartins:
Hey @Chalarangelo 👋 😄 Yea No Problem

Can you also do me a favor & update `tdd.js`?

And add `hashNode` to the errSnippets array on line 32
 
Update 

```js
const errSnippets = ['JSONToFile', 'readFileLines', 'UUIDGeneratorNode'];
```

To 
```js
const errSnippets = ['JSONToFile', 'readFileLines', 'UUIDGeneratorNode', 'hashNode'];
```


The `npm test` command will break because I didn't support extracting requiring modules which I will quickly fix  & submit a PR request to

Chalarangelo:
@kingdavidmartins I'll do so in a bit!

snippets/hashBrowser.md
```diff
@@ -0,0 +1,22 @@
...
+      let hexes = [],
+        view = new DataView(h);
+      for (let i = 0; i < view.byteLength; i += 4)
+        hexes.push(('00000000' + view.getUint32(i).toString(16)).slice(-8));
+      return hexes.join('');
+    });
+```
+
+```js
+hashBrowser(JSON.stringify({a :'a', b: [1,2,3,4], 'foo': {c: 'bar'}})).then(hash => console.log(hash)); // '04aa106279f5977f59f9067fa9712afc4aedc6f5862a8defc34552d8c7206393'
```
skatcat31: # Requested Change
Functions are first order and do not need a wrapper in JS
```js
Promise.resolve('test').then(console.log)
```

snippets/hashNode.md
```diff
@@ -0,0 +1,14 @@
+### hashNode
+
+Creates a hash for a value using the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) algorithm.
+
+Use `crypto` API to create a hash for the given value.
+
+```js
+const crypto = require('crypto');
```
skatcat31: # Question
In node based snippets should they have the require statement in them un-commented?

snippets/hashNode.md
```diff
@@ -0,0 +1,14 @@
...
+const hashNode = val => crypto.createHash('sha256').update(val).digest('hex');
```
skatcat31: # Note
We could make this async and return a promise
```js
const hashNode = val =>
  new Promise( resolve => setTimeout(() => resolve(crypto.createHash('sha256').update(val).digest('hex')),0))
```

snippets/hashNode.md
Chalarangelo: I was thinking that we could actually have the require statement inside the function itself. I think that can work and de-clutter the global namespace, right?

snippets/hashBrowser.md
Chalarangelo: I will fix that, thanks for the heads up!

snippets/hashNode.md
Chalarangelo: I'd rather make the browser one non-async (which is not possible), as this is a lot easier to use than the promise-based version. Usually hashing won't take extremely long and, in most cases I have needed it, it's something you do not want to have running on the background but rather have done immediately. Might be wrong, though.

snippets/hashNode.md
kingdavidmartins: I will quickly submit a PR that will parse handle require. @skatcat31 I honestly think leaving it there is ok. Just need to be able parse it & handle it accordingly

skatcat31: @Chalarangelo do NOT put require statements in functions. It will lead to a lot of overhead that in the worse case scenario could lead to disk access if the module has been fully garbage collected and de-instanced between each use. A require statement in node actually hits a cache algorithm that is very slow. Hence why requires should only be done once and kept around.

If the module in node has been designed badly it can lead to a worse case scenario of a memory leak. DRY coding practices should be adhered to as much as possible. Some good resources on the subject include Vitaly's write up on it in pg-promise and the bluebirdjs write up on it. More information can also be found in the node best practices repo, etc.

Requires and imports are almost analogous, so same applies there

Chalarangelo: @skatcat31 Had no idea, thanks a ton for clarifying this! I suggest we keep them as they are, then, commenting them provides no significant benefit imho.

snippets/hashNode.md
skatcat31: This make the function have the same footprint as the other hashing algorithm so  that you could actively replace one with the other without any other changes to your code. However if we don't need it async, at least resolving it in a promise gives it that capability. I just wanted to replicate the browser behavior as much as possible so that when someone uses it in an async environment( such as a webserver or a request heavy script with multiple instances per something ) they wouldn't hold up other parts of the script while it runs the SHA and Buffer parts of the code unless the eventloop was empty.

NOTE: secure hashing and most crypto functions are actually quite slow( comparatively ). It's why I'm a fan of using the built in apis to make them operate later. If you're using it for a one time use script then it's perfectly fine, however in a webserver you'll either want a callback, or a promise so that it can run when other things are done because of how awfully slow it can be.

snippets/hashNode.md
skatcat31: @Chalarangelo fair, since it does hit the caching algorithm it should reuse. the problem comes in ones that requires instances

snippets/hashNode.md
Chalarangelo: Convincing argument, I'll swap it out to an async version for consistency and the rest of the reasons outlined above.

kingdavidmartins: > However if we don't need it async, at least resolving it in a promise gives it that capability. I just wanted to replicate the browser behavior as much as possible so that when someone uses it in an async environment( such as a webserver or a request heavy script with multiple instances per something ) they wouldn't hold up other parts of the script while it runs the SHA and Buffer parts of the code unless the eventloop was empty.

To be honest @skatcat31 but at that point it really depends on how the user is using this current snippet. Hopefully the user would be playing around with said snippet and using it for educational purposes and not trying to use this exact code in a production/server environment. Because there is soo many way's these can go wrong depending on the use case

skatcat31: @kingdavidmartins fair point! At that point it should be
> However if we don't need it async, at least resolving it in a promise gives it that capability.

and I should have stopped with that argument. Hopefully at least having similar functions that behave similarly helps the learning process along

kingdavidmartins: That being said security/crypto in general is never really handled by a beginner. So I don't think there's really anything to worry about

kingdavidmartins: > Hopefully at least having similar functions that behave similarly helps the learning process along

@skatcat31 Yup Yup Agreed!!

PR #557, opened 2018-01-17
[FEATURE] btoa and atob

Chalarangelo:
## Description

Adds two snippets for Node, `atob` and `btoa`, similar to the browser methods [`atob`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/atob) and [`btoa`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa).

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/atob.md
+++ b/snippets/atob.md
@@ -1,0 +1,13 @@
+### atob
+
+Decodes a string of data which has been encoded using base-64 encoding.
+
+Create a `Buffer` for the given string with base-64 encoding and use `Buffer.toString('binary')` to return the decoded string.
+
+```js
+const atob = str => new Buffer(str, 'base64').toString('binary');
+```
+
+```js
+atob('Zm9vYmFy'); // 'foobar'
+```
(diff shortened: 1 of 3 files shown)


kingdavidmartins:
Doubt beginners will look up 
- atob()
- btoa()

Seems highly improbable

However believe there's value if both have simple names so a beginner / users can quickly search up base64 Encode / Decode & get it

Chalarangelo:
Actually, beginners do look for those when using node, as they exist in browsers and not in Node. I have found myself multiple times in the past looking for a solution to a problem, only to find out a solution that uses `btoa`. Fast forward ten minutes later adn I realize that Node doesn't have this built-in. Main reason I named them like this is that people who know where to use them, know what they are called and they usually have seen them in their browser form.

skatcat31:
I agree with @Chalarangelo, @kingdavidmartins I have on several occasions needed to explain that `btoa` and `atob` don't exist in Node -.-

It's always surprised me they find those so easily

kingdavidmartins:
@Chalarangelo  & @skatcat31 Makes sense!

PR #558, opened 2018-01-18
[ENHANCEMENT] [UPDATE: `flip.md`] Use idiomatic js for flip function 

itaditya:
## Description

Used idiomatic javascript to avoid complex array functions. The rest operator is used to collect all the remaining argument of the function. So instead of taking the whole list and then popping the first element (this is expensive), we simply retain the first argument as a variable `first` and the rest of the arguments are stored as array in `rest` variable. Then we just change the order of arguments.

I think this code is easy to grok and is much faster and more ES6.

```js
const flip = fn => (first, ...rest) => fn(...rest, first);
```

Note - 
Intentionally used the word `rest` instead of `args` because I felt for this case rest better helps people understand that , rest holds all the args after first arg. If I have used `args` then it partly implies that first is not an arg which is not true
[flip function](https://30secondsofcode.org/#flip)

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/flip.md
+++ b/snippets/flip.md
@@ -5,7 +5,7 @@ Flip takes a function as an argument, then makes the first argument the last.
 Return a closure that takes variadic inputs, and splices the last argument to make it the first argument before applying the rest.
 
 ```js
-const flip = fn => (...args) => fn(args.pop(), ...args);
+const flip = fn => (first, ...rest) => fn(...rest, first);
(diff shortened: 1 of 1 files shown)


itaditya:
Thanks, loved to contribute, will be on the lookout for more :smiley:

PR #562, opened 2018-01-20
[ADD: zipWith.md]

kriadmin:
## Description

**Resolves** #100 

**Lodash[ACTION]** #100 -> https://lodash.com/docs/4.17.4#zipwith

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

- [x] I have referenced the `method` to the lodash backlog.

## Checklist:

### DIFF
--- a/snippets/zipWith.md
+++ b/snippets/zipWith.md
@@ -19,4 +19,5 @@ const zipWith = (...arrays) => {
 
 ``` js
 zipWith([1, 2], [10, 20], [100, 200], (a,b,c) => a + b + c); // [111,222]
+zipWith([1, 2, 3], [10, 20], [100, 200], (a,b,c) => (a != null ? a : 'a') + (b != null ? b:'b') + (c != null ? c : 'c')); // [111, 222, '3bc]
 ```
--- a/snippets/zipWith.md
+++ b/snippets/zipWith.md
@@ -1,0 +1,22 @@
+### zipWith
+
+This method is like [zip](https://30secondsofcode.org/#zip) except that it accepts a function (`fn`) as the last value to specify how grouped values should be combined. 
+
+The function is invoked with the elements of each group: `(...group)`.
+
+``` js
+const zipWith = (...arrays) => {
+  const length = arrays.length;
+  let fn = length > 1 ? arrays[length - 1] : undefined;
+  fn = typeof fn == 'function' ? (arrays.pop(), fn) : undefined;
+  const maxLength = Math.max(...arrays.map(x => x.length));
+  const result =  Array.from({ length: maxLength }).map((_, i) => {
+    return Array.from({ length: arrays.length }, (_, k) => arrays[k][i]);
+  })
+  return fn ? result.map(arr => fn(...arr)) : result;
+}
+```
+
+``` js
+zipWith([1, 2], [10, 20], [100, 200], (a,b,c) => a + b + c); // [111,222]
+```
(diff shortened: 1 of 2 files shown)


kriadmin:
@skatcat31 Updated

skatcat31:
Hmm something I just noticed this defaults to zip if there is no function in the last argument. I'm not certain if this should be an intended behavior pattern to avoid two snippets that do the same thing, or if we should replace zip with this snippet. @Chalarangelo @kingdavidmartins @fejes713 any input on that pattern?

skatcat31:
otherwise great submission!

Chalarangelo:
My opinion is to replace `zip` with this implementation, defaulting to the original `zip` behavior without a function in the arguments, like we have done with other snippets. Seems easy enough to understand in my opinion.

kingdavidmartins:
@skatcat31 Although I agree with you with the following

> Hmm something I just noticed this defaults to zip if there is no function in the last argument. I'm not certain if this should be an intended behavior pattern to avoid two snippets that do the same thing.

One thing I think we also need to take into account is how quickly a beginner can really pick & understand what the code is doing. Although `zipWith` can replicate `zip` if 2nd arg is omitted. I think if we completely replace `zip` with said code we will be adding more code which could reduce the readability/simplicity of them understanding the code. 

I also realized that what we believed as "simple" may really not be the case when a few people I met at a meet up who started learning JS couldn't really grasp some of the snippets I thought were relatively easy. I believe having something our users can build off to learn from so they can see how things connect and etc is useful. Rather than just removing `zip` entirely.  

The way I see it `zipWith` is a more advance version of `zip` such as users can have more control of how things are zipped. 

**Note:** I was think having a little pointer in the description can go a long way for snippets like these pointing users who feel the current snippet is too hard to a easier simpler version.

Like `if this snippets seems a bit more than you can handle checkout out our zip snippet`

@Chalarangelo Thoughts?

Chalarangelo:
Iterlinking snippets will lead to a situation where we have to maintain links atop of the snippet list, which leads to all sorts of problems when archiving, renaming, removing etc. Marking snippets as `advanced` was a decent first step, maybe we want to add a difficulty rating via a system, so it's easy to understand the relative level of difficulty for a JS user. That being said, what might be hard for me to grasp might be super easy for someone else and vice versa, so in general this is very subjective and won't work out too well.

Keeping both snippets, in this case, makes more sense probably as the `zipWith` one is complex enough to require some extra time to read through. Sticking an `advanced` tag can certainly help, plus `zip` is right above or below it anyways, so discoverability is a non-issue here.

kingdavidmartins:
>  plus zip is right above or below it anyways, so discoverability is a non-issue here.

@Chalarangelo haha 😆 Of course I totally forgot. Makes Sense!

I also think an advance tag can go some ways. 

Maybe anythings with over a certain amount of lines maybe 5, and two/or more of the following closures, reduce, `...` operator, or.....

I don't just brainstorming here. Any other ideas? 

Thoughts @skatcat31 ?

skatcat31:
I think they idea of keeping this as is and keeping zip might be a good idea if we mark this advanced since it shows people something most of our other snippets don't, the special function type

kriadmin:
@skatcat31 @Chalarangelo @kingdavidmartins I don't think this snippet is too advanced and can also be called very easy to understand in comparision to the lodash code. And I checked the behavior of `_.zipWith` and it is the same so I think we can keep them both.

Issue #564, opened 2018-01-21
[BUG] tomorrow() is wrong two hours of the year


NielsLeenheer:
The `tomorrow()` function is based on the assumption that a day is always 60 * 60 * 24 = 86400 seconds long. But it is not. Due to daylight savings time most countries have one day each year that 23 hours long and one day that is 25 hours long.

So take change from winter time to summer time in The Netherlands:
On Sunday, 25th of March 2018 at 02:00 the clock skips and hour to 03:00 and the day will be just 23 hours. So if you call the function the day before on 24th of March at 23:01 and add 24 hours, it will not arrive at 23:01 of the 25th, but instead at 00:01 of the 26th. That is not tomorrow, but the day after tomorrow.

And the change from summer time to wintertime in The Netherlands:
On Sunday, 27th of October 2018 at 03:00 the clock goes back to 02:00 and the day will be 25 hours long. So if you call the function on the 27th at 00:59 and add 24 hours, it will not arrive at 00:59 the next day, but on 23:59 of the same day.

So for two hours each year, this function will give the wrong result. For one hour it will give the same day and for one hour it will give the day after tomorrow.

Using the build-in getDate and setDate function of the Date object is safer.

```
const tomorrow = () => { 
    let t = new Date();
    t.setDate(new Date().getDate() + 1);
    return t.toISOString().split('T')[0]; 
};
```

NielsLeenheer:
Actually the problem is worse than this.

You should not use toISOString() for this. That function converts the local time to UTC and there is no reason to assume the date in UTC time is the same as your local timezone.

Take for example:

Assume it is 09:00 in Sydney, Australia. Today it has a 11 hour timezone difference with UTC. 

````
new Date()
> Mon Jan 22 2018 09:00:00 GMT+1100 (AEDT)
````

Now if I determine the date using toISOString().

````
new Date().toISOString()
> "2018-01-21T22:00:00.000Z"

new Date().toISOString().split('T')[0]
> "2018-01-21"
````

The date is off by one day. So if I were to get the date of tomorrow and use the correct method to do that, than I would still get today as the result.

So the correct way is probably something like this:

````
const tomorrow = () => { 
    let t = new Date();
    t.setDate(new Date().getDate() + 1);
    return t.getFullYear() + '-' + String(t.getMonth() + 1).padStart(2, '0') + '-' + String(t.getDay()).padStart(2, '0')
};
````

skatcat31:
[...]

PR #565, opened 2018-01-22
[FIX: #564] Fix inaccurate tomorrow() function

NielsLeenheer:
## Description
Fix a problem due to the assumption that days are always 24 hours long, while due to DST one day a year is 23 hours and one day a year is 25 hours long. This lead this function to give the current date during one hour a year and the day after tomorrow for another hour a year.

Additionally, the function only worked correctly when the user was in the same timezone as UTC. For example in Amsterdam which 1 hour from DST it only worked 23 hours a day. And in Sydney which is 11 hours away from DST, the function only worked correctly for 13 hours each day. During the hours it didn't work correctly it would give the current date instead of tomorrows.

**Resolves** #564  

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/tomorrow.md
+++ b/snippets/tomorrow.md
@@ -1,10 +1,20 @@
 ### tomorrow
 
 Results in a string representation of tomorrow's date.
-Use `new Date()` to get today's date, adding `86400000` of seconds to it(24 hours), using `Date.toISOString()` to convert Date object to string.
+Use `new Date()` to get today's date, adding one day using `Date.getDate()` and `Date.setDate()`, and converting the Date object to a string.
 
 ```js
-const tomorrow = () => new Date(new Date().getTime() + 86400000).toISOString().split('T')[0];
+const tomorrow = () => {
+  let t = new Date();
+  t.setDate(t.getDate() + 1);
+  return (
+    t.getFullYear() +
+    '-' +
+    String(t.getMonth() + 1).padStart(2, '0') +
+    '-' +
+    String(t.getDay()).padStart(2, '0')
+  );
+};
 ```
 
 ```js
--- a/snippets/tomorrow.md
+++ b/snippets/tomorrow.md
@@ -7,13 +7,7 @@ Use `new Date()` to get today's date, adding one day using `Date.getDate()` and 
 const tomorrow = () => {
   let t = new Date();
   t.setDate(t.getDate() + 1);
-  return (
-    t.getFullYear() +
-    '-' +
-    String(t.getMonth() + 1).padStart(2, '0') +
-    '-' +
-    String(t.getDate()).padStart(2, '0')
-  );
+  return `${t.getFullYear()}-${String(t.getMonth() + 1).padStart(2, '0')}-${String(t.getDate()).padStart(2, '0')}`;
 };
 ```
 
--- a/snippets/tomorrow.md
+++ b/snippets/tomorrow.md
@@ -12,7 +12,7 @@ const tomorrow = () => {
     '-' +
     String(t.getMonth() + 1).padStart(2, '0') +
     '-' +
-    String(t.getDay()).padStart(2, '0')
+    String(t.getDate()).padStart(2, '0')
   );
 };
 ```


snippets/tomorrow.md
```diff
@@ -1,10 +1,20 @@
...
-const tomorrow = () => new Date(new Date().getTime() + 86400000).toISOString().split('T')[0];
+const tomorrow = () => {
+  let t = new Date();
+  t.setDate(t.getDate() + 1);
+  return (
+    t.getFullYear() +
+    '-' +
+    String(t.getMonth() + 1).padStart(2, '0') +
+    '-' +
+    String(t.getDay()).padStart(2, '0')
```
skatcat31: `getDay` is 0-6(M-F), `getDate` is 0-End of Month. If we're meaning to emulate the ISOString you need to use `t.getDate()`

snippets/tomorrow.md
```diff
@@ -1,10 +1,20 @@
...
 
 Results in a string representation of tomorrow's date.
-Use `new Date()` to get today's date, adding `86400000` of seconds to it(24 hours), using `Date.toISOString()` to convert Date object to string.
+Use `new Date()` to get today's date, adding one day using `Date.getDate()` and `Date.setDate()`, and converting the Date object to a string.
 
 ```js
-const tomorrow = () => new Date(new Date().getTime() + 86400000).toISOString().split('T')[0];
+const tomorrow = () => {
+  let t = new Date();
+  t.setDate(t.getDate() + 1);
```
kingdavidmartins: Hey @NielsLeenheer 👋 😄 

Thanks for taking the time to submit this PR

However here are my suggestions 

Rather than spreading everything out and making `-` appear on their own line.

Why not keep things nice & short as possible by doing some variant of the following. 

```js
const tomorrow = () => {
  let time = new Date();
  let month = `${time.getMonth()+1}`.padStart(2, '0');
  let day = `${time.getDate()+1}`.padStart(2, '0');
  return `${time.getFullYear()}-${month}-${day}`;
};
```

Chalarangelo: This will definitely break on the last day of each month, as it will go from `Dec 31` to `Dec 01`. Btw in this example, the year will also be broken. A way to make it shorter would be:

```js
const tomorrow = () => {
  let t = new Date();
  t.setDate(t.getDate() +1);
  return `${t.getFullYear()}-${String(t.getMonth() + 1).padStart(2, '0')}-${String(t.getDate()).padStart(2, '0')}`;
};
```

kingdavidmartins: @Chalarangelo ahhh yea, Nice catch!

I quickly came up with a sloppy solution. I am sure this can be refactored. But it takes care of the edge cases you mentioned
```js
const tomorrow = () => {
  let time = new Date();
  let maxDay = { 01: 31, 02: 28, 03: 31, 04:30, 05: 31, 06: 30, 07: 31, 08:31, 09: 30, 10: 31, 11: 30, 12: 31}; 
  let month = `${time.getMonth()+1}`.padStart(2, '0');
  let nextMonth = `${time.getMonth()+2}`.padStart(2, '0');
  let day = `${time.getDate()}`.padStart(2, '0');
  let isLastDayOfMonth = maxDay[month] === day;
  let tomorrow = `${time.getDate()+1}`.padStart(2, '0');
  let isNewYearEve = month == 12 && maxDay[month] === day;
  let year = isNewYearEve ? time.getFullYear() + 1 : time.getFullYear();
  return `${year}-${isNewYearEve ? 01 : isLastDayOfMonth ? nextMonth : month}-${isLastDayOfMonth ? 01 : tomorrow}`;
};
```

NielsLeenheer: @kingdavidmartins and @Chalarangelo 

The use of 
```js
t.setDate(t.getDate() + 1);
```
on a separate line is very deliberate. 

This will **not** break on month or year boundaries. Using this line on `Nov 30` will in fact set it to `Dec 1` and not `Nov 1`. It will also take care of year boundaries automatically and update `Dec 31, 2017` to `Jan 1, 2018`. This behaviour is in the spec. This is exactly what the `Date` object should do and every implementation follows this.

So you should not handle the boundaries yourself, they `Date` object does that for you. But if you do, you are bound to make mistakes due to wrong assumptions. For example in that last solution there is an assumeption that every February will have 28 days. But it doesn't due to leap days. 

@Chalarangelo 

Using templates to generate the return value is indeed shorter and changing that would be okay. But in my opinion it does hurt readability. And the non-templated version is just 4 bytes longer.

skatcat31: :+1: on the Date rollover functionality, that is very often forgotten about. Should not be changed.

As for "shortness" I think moving to the template will help enough. When that change happens it should be fine

skatcat31: example of rollover taken from console of Chrome:
```js
const b = new Date()
b //Sun Dec 31 2028 22:42:13 GMT-0800 (Pacific Standard Time)
b.setDate(32) //1862030533926
b //Mon Jan 01 2029 22:42:13 GMT-0800 (Pacific Standard Time)
```

kingdavidmartins: @NielsLeenheer Woah! totally forgot. Nice!! Thanks for pointing that out 🎉 ⚡️ 

Haven't really been reviewing dates documentation or using native dates methods as much since momentjs 😞 

Def going to go over the documentation more in depth later

Can confirm I also get the rollover result.

As for using template literals, I'm with @skatcat31.  Even though it will not be significatly shorter. It still will be shorter, have less lines, and promote the use of ECMAScript 2015

Update using template literals and should be good to go 😄 

Suggestion 1 ~ More readable

```js
const tomorrow = () => {
  let t = new Date();
  t.setDate(t.getDate() + 1);
  let month = `${t.getMonth()+1}`.padStart(2, '0');
  let day = `${t.getDate()}`.padStart(2, '0');
  return `${t.getFullYear()}-${month}-${day}`;
};
```

Suggestion 2 ~ Less lines

```js
  let t = new Date();
  t.setDate(t.getDate() + 1);
  return `${t.getFullYear()}-${String(t.getMonth()+1).padStart(2, '0')}-${String(t.getDate()).padStart(2, '0')}`;
};
```

PR #571, opened 2018-01-26
delay.md's heading was defer

coolreader18:
Basically that. [Here's](https://cdn.pbrd.co/images/H4DuuaT.png) a screenshot if someone wants it.


skatcat31:
Thanks for catching that!

Issue #526, opened 2018-01-10
Test ~ Log


fejes713:
Similarly to the #100, we can keep track of snippets that are missing test cases.
## Here is the list of snippets that need their test cases implemented:
*Be sure to tick the ones that you or someone else solved. If you don't have access to edit this file be sure to let us know in the comments about the changes*
- [x] anagrams
- [x] arrayToHtmlList
- [x] ary
- [x] atob
- [x] attempt
- [x] average
- [x] averageBy
- [x] bind
- [x] bindAll
- [x] bindKey
- [x] bottomVisible
- [x] btoa
- [x] byteSize
- [x] call
- [x] capitalize
- [x] capitalizeEveryWord
- [x] castArray
- [x] chainAsync
- [x] chunk
- [x] clampNumber
- [x] cloneRegExp
- [x] coalesce
- [x] coalesceFactory
- [x] collectInto
- [x] colorize
- [x] compact
- [x] compose
- [x] composeRight
- [x] copyToClipboard
- [x] countBy
- [x] countOccurrences
- [x] createElement
- [x] createEventHub
- [x] currentURL
- [x] curry
- [x] debounce
- [x] decapitalize
- [x] deepClone
- [x] deepFlatten
- [x] defaults
- [x] defer
- [x] delay
- [x] detectDeviceType
- [x] difference
- [x] differenceBy
- [x] differenceWith
- [x] digitize
- [x] distance
- [x] drop
- [x] dropRight
- [x] dropRightWhile
- [x] dropWhile
- [x] elementIsVisibleInViewport
- [x] elo
- [x] equals
- [x] escapeHTML
- [x] escapeRegExp
- [x] everyNth
- [x] extendHex
- [x] factorial
- [x] fibonacci
- [x] filterNonUnique
- [x] findKey
- [x] findLast
- [x] findLastIndex
- [x] findLastKey
- [x] flatten
- [x] flip
- [x] forEachRight
- [x] formatDuration
- [x] forOwn
- [x] forOwnRight
- [x] fromCamelCase
- [x] functionName
- [x] functions
- [x] gcd
- [x] geometricProgression
- [x] get
- [x] getDaysDiffBetweenDates
- [x] getScrollPosition
- [x] getStyle
- [x] getType
- [x] getURLParameters
- [x] groupBy
- [x] hammingDistance
- [x] hasClass
- [x] hasFlags
- [x] hashBrowser
- [x] hashNode
- [x] head
[...]

PR #583, opened 2018-01-30
[#526] Added some test cases

tcmal:
Adds 5 new test cases for previously untested snippets.

Addresses #526 

## Description

Wrote test cases for binarySearch(), bottomVisible(), call(), chainAsync() and collatz()

**Addresses ** #526 

- [ ] Website
- [ ] Snippets
- [X] Tests

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:


snippets/chainAsync.md
```diff
@@ -4,10 +4,12 @@ Chains asynchronous functions.
 
 Loop through an array of functions containing asynchronous events, calling `next` when each asynchronous event has completed.
 
+The tenerary function checks the next function exists before calling it, otherwise it will exit.
+
 ```js
 const chainAsync = fns => {
   let curr = 0;
-  const next = () => fns[curr++](next);
+  const next = () => (fns[++curr] ? fns[curr](next) : false);
```
Chalarangelo: We are not handling bad input in most snippets, so I don't think we should make this change here. It's better to allow the method to just throw an error and let developers handle it, instead of exiting unexpectedly like that.

test/bottomVisible/bottomVisible.test.js
```diff
@@ -1,12 +1,29 @@
...
+};
+
+const window = {
+  scrollY: 2160
+}
+
+// Override the function to use these fake properties
+const bottomVisible = () =>
+document.documentElement.clientHeight + window.scrollY >=
+(document.documentElement.scrollHeight || document.documentElement.clientHeight);
```
Chalarangelo: Either use `jsdom` or revert these to their original. We currently have some trouble with the browser-specific snippets, so it'd be better to just revert changes to this file until we figure it all out.

snippets/chainAsync.md
tcmal: I don't think the changes make it exit unexpectedly, only when it reaches the end of the list, as is the expected behaviour.

kingdavidmartins: Hey @tcmal 👋 😄 

Thanks for submitting a PR

However I believe your code doesn't work as expected and should leave as it. As stated by @Chalarangelo 

```js
const chainAsync = fns => {
  let curr = 0;
  const next = () => (fns[++curr] ? fns[curr](next) : false);
  next();
};
```

The following doesn't run as expected
```js
chainAsync([
  next => {
    console.log('0 seconds');
    setTimeout(next, 1000);
  },
  next => {
    console.log('1 second');
  }
]);
```

`1 second` is printed to the console automatically instead of working as expected

tcmal: Ah ok, think I misunderstood original review. Have commited change that appears to fix this on my end.

PR #602, opened 2018-02-10
fix broken test in tomorrow.test.js

kingdavidmartins:
## Description

fixes failing the `npm test` command

## What does your PR belong to?
- [ ] Website
- [ ] Snippets
- [ ] General / Things regarding the repository (like CI Integration)
- [x] Tests

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (non-breaking improvement of a snippet)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have checked that the changes are working properly
- [x] I have checked that there isn't any PR doing the same
- [x] I have read the **CONTRIBUTING** document.

PR #603, opened 2018-02-10
Update tomorrow

kingdavidmartins:
## Description

**Resolves**  #602  

- [ ] Website
- [ ] Snippets
- [x] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/tomorrow.md
+++ b/snippets/tomorrow.md
@@ -10,7 +10,7 @@ const tomorrow = (long = false) => {
   const ret = `${t.getFullYear()}-${String(t.getMonth() + 1).padStart(2, '0')}-${String(
     t.getDate()
   ).padStart(2, '0')}`;
-  return !long ? ret : `${ret}T:00:00:00`;
+  return !long ? ret : `${ret}T00:00:00`;
 };
 ```
 
--- a/snippets/tomorrow.md
+++ b/snippets/tomorrow.md
@@ -7,12 +7,12 @@ Use `new Date()` to get today's date, adding one day using `Date.getDate()` and 
 const tomorrow = () => {
   let t = new Date();
   t.setDate(t.getDate() + 1);
-  return `${t.getFullYear()}-${String(t.getMonth() + 1).padStart(2, '0')}-${String(
+  return `${String(t.getMonth() + 1).padStart(2, '0')}-${String(
     t.getDate()
-  ).padStart(2, '0')}`;
+  ).padStart(2, '0')}-${t.getFullYear()}`;
 };
 ```
 
 ```js
-tomorrow(); // 2017-12-27 (if current date is 2017-12-26)
+tomorrow(); // 12-27-2017 (if current date is 12-26-2017)
 ```
--- a/snippets/tomorrow.md
+++ b/snippets/tomorrow.md
@@ -4,15 +4,17 @@ Results in a string representation of tomorrow's date.
 Use `new Date()` to get today's date, adding one day using `Date.getDate()` and `Date.setDate()`, and converting the Date object to a string.
 
 ```js
-const tomorrow = () => {
+const tomorrow = (long = false) => {
   let t = new Date();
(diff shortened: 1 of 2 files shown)


kingdavidmartins:
@Chalarangelo now all the test run 😄 

Your were right your test were perfect 👍  it wasn't the test but the snippet.

Please let me know your results. 

Everything should work on your end. if not ....

<strong>*Note:*</strong> *It can also be a timezone thing.*

kingdavidmartins:
@skatcat31 Because the original snippet `tomorrow.md` behaves the following way when passes in `new Date()`

```js
new Date(tomorrow()).getDate() // returns the curremt date -> 11
```

checkout -> https://repl.it/repls/RotatingTightBluefish

alt text

But when formatted the proposed way it returns the following which is what we expected

```js
new Date(tomorrow()).getDate() // returns the curremt date -> 12
```

checkout -> https://repl.it/repls/ChartreuseUnwrittenPerch

alt text

skatcat31:
Ah previous version parses with ISO8601 in timezone parser whereas second one parses as english date which works apparently

Chalarangelo:
I'd rather have the ISO8601 version, instead of the English Date one. I mean we had a long discussion about conforming to ISO8601, this is our first chance to honor that decision.

skatcat31:
The problem is I kind of see both sides due to the parsing of the intermediate value. It gets weird when you think about it. Maybe put a Boolean for the second argument? Make it output with TZ to make it parse correctly

mariastervic:
The existing version with ISO8601 worked fine, why are you changing this?

skatcat31:
@mariastervic if you parse the returned tomorrow value it parses at local time, resetting it to "today" instead of keeping it tomorrow.

The englishdate organization parses as "tomorrow" when sent into `Date`.

skatcat31:
I think we should have a Boolean as a second argument, defaults to false for ISO, true for longest length ISO8601 with an appended `T00:00:00Z` so that parsing is doable thorough all date libraries

skatcat31:
I hate Github's site the more I use Gitlab...
Taken from Master
```js
const tomorrow = (long = false) => {
  let t = new Date();
  t.setDate(t.getDate() + 1);
  const tomorrowString = `${t.getFullYear()}-${String(t.getMonth() + 1).padStart(2, '0')}-${String(
    t.getDate()
  ).padStart(2, '0')}`;
  return !long ? tomorrowString : `${tomorrowString}T00:00:00Z`;
};
```

Chalarangelo:
I like this idea, @skatcat31. Any objections by anyone? @kingdavidmartins?

skatcat31:
Just to display the importance of a single letter...
```js
const tomorrow = (long = false) => {
  let t = new Date();
  t.setDate(t.getDate() + 1);
  const tomorrowString = `${t.getFullYear()}-${String(t.getMonth() + 1).padStart(2, '0')}-${String(
    t.getDate()
  ).padStart(2, '0')}`;
  return !long ? tomorrowString : `${tomorrowString}T00:00:00`;
};
```

kingdavidmartins:
@skatcat31 haha I was actually going to suggest you omit the `Z` but you already caught it 😄

kingdavidmartins:
I'm all for it @skatcat31 @Chalarangelo

Chalarangelo:
Alright, let's do that then and update tests and everything about this snippet, so that this PR can be finally merged!

skatcat31:
I'm not at home so I can't regenerate the test files, but I've updated the snippet to have the long modifier to return a full ISO 8601 compatible string, or we could just default to always returning that compatible string

skatcat31:
Wrong comment button, ignore close...

Chalarangelo:
Bump! Where do we stand on this?

Chalarangelo:
My GitKraken has basically broken, so I am going to merge and run the tester updates on the Travis CI, hope this doesn't cause any problems.

PR #613, opened 2018-02-24
[FEATURE] pad - string padding

commenthol:
Adds string padding function.
- `pad` pads on both sides

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/pad.md
+++ b/snippets/pad.md
@@ -1,0 +1,16 @@
+### pad
+
+Pads `string` on the left and right side if it's shorter than `length`.
+String is *NOT* truncated on both sides if it exceeds length.
+
+```js
+const pad = (string, length = 8, char = ' ') =>
+  string.padStart((string.length + length) / 2, char).padEnd(length, char);
+module.exports = pad;
+```
+
+```js
+pad('cat'); //> '  cat   '
+pad(String(42), 4, '0'); //> '004200'
+pad('foobar', 3); //> 'foobar'
+```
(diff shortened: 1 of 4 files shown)


Chalarangelo:
These methods are natively implemented, so I don't think we should add them.

commenthol:
Did not know so far. Thanks for telling me. 
Albeit `pad` is not available. Do you want me to leave just the pad  function?

Chalarangelo:
I'm away for the weekend, so I cannot review it right now, but if we keep anything it will just be the `pad` snippet. If any other members of the core team can review it and provided that this is not natively implemented, I suppose we should keep it.

snippets/pad.md
```diff
@@ -0,0 +1,16 @@
+### pad
+
+Pads `string` on the left and right side if it's shorter than `length`.
+String is *NOT* truncated on both sides if it exceeds length.
+
+```js
+const pad = (string, length = 8, char = ' ') =>
+  string.padStart(string.length + (length - string.length) / 2, char).padEnd(length, char);
```
kriadmin: `string.length/2` should be `Math.floor(string.length/2)` just to be safe against people looking for fun.

snippets/pad.md
```diff
@@ -0,0 +1,16 @@
...
+module.exports = pad;
```
kriadmin: Remove this. We don't need it.

snippets/pad.md
commenthol: Could you give me an example what "fun" I might look for? Just being curious...

PR #616, opened 2018-02-25
Create recordAnimationFrames.md

atomiks:
## Description

An interesting / esoteric snippet I came up with to debug a problem with someone's library (smooth scroller when clicking an anchor tag)

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/recordAnimationFrames.md
+++ b/snippets/recordAnimationFrames.md
@@ -2,8 +2,10 @@
 
 Invokes the provided callback on each animation frame.
 
-Use recursion. Provided that `running` is `true`, continue invoking `window.requestAnimationFrame()` which invokes the
-provided callback. Return an object with two methods `start` and `stop` to allow manual control of the recording. Omit the second argument, `autoStart`, to implicitly call `start` when the function is invoked.
+Use recursion. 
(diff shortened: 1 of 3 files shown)


Chalarangelo:
I like the idea, but the snippet is probably far too specific, so I cannot see too many valid use-cases. I'm voting to move this to the archive. Also, there are a ton of missing semicolons in the snippet and the length is a tiny bit on the long side (which is not a real issue, provided how short each line is, but it might look intimidating). I'd like to hear more feedback from the team on this.

atomiks:
I use semi-colonless style, but it will get fixed by linter/prettier anyway so I didn't bother

Chalarangelo:
Ok, so the only issue remaining is the usability and if it should be added to the main listing or the archive as a result? @atomiks can you think of any common use-cases where one might want this? I suspect I am not really giving the snippet as much credit as I should.

atomiks:
Well for the problem I had, something was occuring on every animation frame (the scroll position was changing), so I wanted to "record" each frame to see what the value of `window.scrollY` was to debug it.

I guess its main use is anything to do with animation. You might want to grab the value of something (such as styles of an element, etc) while it's animating, which this easily allows you to do.

PR #617, opened 2018-02-25
[ENHANCEMENT] smooth scrolling on iOS devices

rainshen49:
- [x] Website
- [ ] Snippets
- [ ] Tests

## Types of changes

## Checklist:


Chalarangelo:
As MDN says:
> **Non-standard**
> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

So I'm against adding this, unless it solves some real big issue on the website.

atomiks:
@Chalarangelo that message is meaningless, it's an important/widely-used property to fix overflow scrolling on iOS

Chalarangelo:
[caniuse](https://caniuse.com/#search=-webkit-overflow-scrolling) does not even recognize it, is it really that big of a deal? (Never been on an iOS device, I don't even understand what we are talking about here)

My assumption was that if MDN says it's technically invalid, there's no reason to have it, right?

atomiks:
On iOS, when scrolling scrollable divs (not the body however) without it, the scrolling has no inertia which makes scrolling feel really sluggish. It's non-standard but only applies to iOS and shouldn't be going away soon, so it's fine to use.

Chalarangelo:
Ok, reopening and merging. Whatever quirky thing that is, no problem adding one line of CSS I guess if it doesn't affect anything else.

fejes713:
@Chalarangelo iOS handles native scrolling a bit differently so this should solve that problem.

PR #628, opened 2018-03-07
[FEATURE] is NaN type function

fejes713:
## Description
This one was missing from `type` category. It is a bit trickier compared to other snippets, so let me explain:

We use NaN to define a number that isn't really a number `"foo" / 2 = NaN`. `typeof(NaN) -> "number"`.

**NaN compared to anything is false**, even to itself. There is native `isNaN` function but it isn't stable, here are the examples with explanations:

_native_ isNaN(NaN) _returns_ `true` (this one is okay)
_native_ isNaN(1) _returns_ `false` (this is okay too)
_native_ isNaN("1") _returns_ `false` (this is okay too)
_native_ isNaN("foo") _returns_ `true` (this is **not okay** obviously, it is because of JavaScript coercion) - Number("foo") returns NaN. Therefore isNaN("foo") turns into isNaN(NaN)

The solution to the problem:
In JavaScript the `NaN` is the only value that is treated as following `NaN == NaN returns false`. Therefore, we can use strict inequality to compare such value. The following solution is often asked on job interviews. 

✅ I've updated tests accordingly. New tests are passing, however, there were 2 tests that fail from before: `pad` and `toCurrency`

 

- [ ] Website
- [x] Snippets
- [x] Tests

## Types of changes

## Checklist:


atomiks:
`Number.isNaN()` is the new native alternative, but I think it's still an interesting concept, where `NaN` is the only value that is not equal to itself.

Chalarangelo:
This is natively implemented as [`Number.isNaN()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN), so we don't really need to add it ourselves. Let ES6 handle the NaN weirdness, trust me it's for the best.

PR #631, opened 2018-03-12
[WEB] Avoid forced reflows inside search function

brettapeters:
## Description

This PR improves the performance of the snippet search function on the website. The `search` was running slowly (~1 second per `keyup` event) due to forced reflows caused by alternating style reads and writes for each iteration of its internal loops. For example, on each iteration through the loop of snippet nodes, there was a style read due to calling `x.innerText` on a DOM node, then a style write due to `x.style.display = ...`. This forced a reflow on each time through the loop, slowing down the total time taken to search through all nodes for matches. This profile depicts one call of `search` with the argument `"if"` prior to the fix.

before

Instead, replacing `x.innerText` with `x.textContent` avoids forced reflows and allows the browser to batch all style updates when the loop finishes. This profile shows the same call to `search` as above, but without the intermediate reflows:

after

I found that this change removed the noticeable delay between typing a character and seeing the updated list of results. 

- [x] Website
- [ ] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/static-parts/index-start.html
+++ b/static-parts/index-start.html
@@ -16,7 +16,7 @@
       const search = (node) => {
         let matchingTags = [];
         Array.from(node.parentElement.parentElement.getElementsByTagName('a')).forEach(x => {
-          let data = [x.innerText.toLowerCase(), ...x.getAttribute('tags').split(',')].map(v => !!(v.indexOf(node.value.toLowerCase()) + 1));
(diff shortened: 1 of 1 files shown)


brettapeters:
Thanks! It was a good learning opportunity for me.

skatcat31:
Did not know that was a property I could use. Great find!

Issue #629, opened 2018-03-07
Questions about `sortedIndex` and `sortedLastIndex`


HowardTangHw:
**sortedIndex**
Returns the `lowest` index at which value should be inserted into array in order to maintain its sort order

**sortedLastIndex**
Returns the `highest` index at which value should be inserted into array in order to maintain its sort 
order.

but 
```javascript
sortedIndex([5, 3, 2, 1], 4); // 1

sortedLastIndex([5, 3, 2, 1], 4); // 0

```

Is the answer correct? Or is my understanding of the mistake

Chalarangelo:
`sortedIndex` is correct. `sortedLastIndex` *should* return the same, but it does not seem to do so. Let me check this out.

HowardTangHw:
const sortedLastIndex = (arr, n) => {
  const isDescending = arr[0] > arr[arr.length - 1];
  const index = arr.reverse().findIndex(el => (isDescending ? n <= el : n >= el));
  return index === -1 ? 0 : arr.length - index;
};

Chalarangelo:
I remember distinctly trying to use this exact code and breaking for certain cases, while `-1` did somehow solve problems. I'm quite certain I missed something at some point. I will need some edge cases to test this on, but this will certainly be fixed soon.

PR #632, opened 2018-03-12
[FIX:#629] Fix invalid test for sortedLastIndex & Fix sortedLastIndex

tcmal:
## Description

Change the test of sortedLastIndex to expect a return value of 4 and stopped sortedLastIndex subtracting 1 before returning

**Resolves** #629 

- [ ] Website
- [X] Snippets
- [X] Tests

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/test/sortedLastIndex/sortedLastIndex.test.js
+++ b/test/sortedLastIndex/sortedLastIndex.test.js
@@ -6,7 +6,7 @@ test('Testing sortedLastIndex', (t) => {
   //Please go to https://github.com/substack/tape
   t.true(typeof sortedLastIndex === 'function', 'sortedLastIndex is a Function');
   t.equal(sortedLastIndex([10, 20, 30, 30, 40], 30), 4, 'Returns the highest index to insert the element without messing up the list order');
-	//t.deepEqual(sortedLastIndex(args..), 'Expected');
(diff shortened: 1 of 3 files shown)


test/sortedLastIndex/sortedLastIndex.test.js
```diff
@@ -5,7 +5,7 @@ test('Testing sortedLastIndex', (t) => {
   //For more information on all the methods supported by tape
   //Please go to https://github.com/substack/tape
   t.true(typeof sortedLastIndex === 'function', 'sortedLastIndex is a Function');
-  t.equal(sortedLastIndex([10, 20, 30, 30, 40], 30), 3, 'Returns the highest index to insert the element without messing up the list order');
+  t.equal(sortedLastIndex([10, 20, 30, 30, 40], 30), 4, 'Returns the highest index to insert the element without messing up the list order');
```
Chalarangelo: The test is wrong now, as it should return `4`. This is due to the fact that inserting at index equal to `3` would make it so that the array would be `[10, 20, 30, 30 (the inserted entry), 30, 40]` while it should be `[10, 20, 30, (the inserted entry), 30, 30, 40]`.

PR #633, opened 2018-03-15
[FEATURE] Add mapAsync snippet

Birkenstab:
## Description
Maps the values of an array using the async function provided. Returns a promise that resolves with an array that contains all mapped values

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/mapAsync.md
+++ b/snippets/mapAsync.md
@@ -1,0 +1,17 @@
+### mapAsync
+
+Maps the values of an array using an async function.
+Returns a promise that resolves with an array of the mapped values
+
+Use `Promise.all()` to wait till all promises have completed
+Use `Array.map` to map the array
+
+```js
+const mapAsync = (arr, fn) => 
+  Promise.all(arr.map(fn));
+```
+
+```js
+const asyncFn = (val) => new Promise(resolve => setTimeout(() => resolve(val ** 2), 20));
+await mapAsync([12, 5, 8, 3], asyncFn); // [144, 25, 64, 9]
+```
--- a/tag_database
+++ b/tag_database
@@ -147,6 +147,7 @@ lcm:math,recursion
 longestItem:array,string,utility
 lowercaseKeys:object
 luhnCheck:math,utility
+mapAsync:array,function
 mapKeys:object,function
 mapObject:array,object,advanced
 mapValues:object,function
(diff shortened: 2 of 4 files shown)


skatcat31:
I'm not such a fan since it's obfuscating already hard to understand concepts and is merely a wrapper

```js
var a = [1,2,3]
var b = a.map(i => setTimeout(() => Promise.resolve(i*2),0))
Promise.all(b)
```

behaves the same, and it's obvious what you're doing there, making debugging MUCH simpler than

```js
mapAsync(a, fn)
```

which begs the question of "Which order is it async?" Is the Array async and the function executing after it all resolves, or is the mapping Async?

Chalarangelo:
@skatcat31 I am not entirely certain about the implementation specifics and the way this behaves, but the main question that begs asking for me is if we really need this snippet and if it has valid and common enough use-cases.

skatcat31:
Valid common use case is less typing in a common factor, however it is confusing no matter which way you implement it because mapAsync does not describe how it's deciding to map and there can be no standardized way to describe it with a short name that would actually make it more useful.

```js
a = [1,2,3]
mapAsync(a, fn)
// vs
Promise.all(a.map(fn))       // implimented the way it is here as map all then resolve
// vs
Promise.all(a).then(b => b.map(fn))       // another mapAsync( resolve all, then map )
// vs
new Promise((resolve) => setTimeout(() => resolve(a),0)).then(b => b.map(fn))        // another mapAsync( push, then map )
```

But as you can see all of these could be called mapAsync which is the problem. which GREATLY limits the usefulness of this snippet and whether we'd need it( confusion over functionality is NEVER good )

Chalarangelo:
I'm pretty sure this would confuse users as it does confuse me even after I took a long hard look at it. Does it qualify as archive material or should we just close the PR altogether, @skatcat31 ?

skatcat31:
We should close altogether due to the ambiguous nature of most functions called mapAsync and how little it ends of up saving in terms of development since if you know you're going to map to an array of Promises, you also know you're going to us a Promise.all, and at that point non promise transformations are no longer able to be chained and you'd return to a different context where the Promise.all would be called from then and continued on.

skatcat31:
To be fair, this use pattern comes up A LOT, but it is not something that can easily be named or reasoned, so closed it is

Birkenstab:
Ok, no problem

PR #648, opened 2018-04-11
[FEATURE] hz

atomiks:
## Description

Returns the number of times a function executed per second after `iterations` times. 

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/hz.md
+++ b/snippets/hz.md
@@ -3,7 +3,9 @@
 Returns the number of times a function executed per second. 
 `hz` is the unit for `hertz`, the unit of frequency defined as one cycle per second.
 
-Use `window.performance.now()` to get the difference in milliseconds before and after the iteration loop to calculate the time elapsed executing the function `iterations` times. Return the number of cycles per second by converting milliseconds to seconds and dividing it by the time elapsed. Omit the second argument, `iterations`, to use the default of 100 iterations.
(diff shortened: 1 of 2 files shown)


atomiks:
Interesting application:

```js
const isAnimationFriendly = (fn, millisecondBudget = 10) => 
  hz(fn) > 1000 / millisecondBudget
```

"Animation friendly" means the function can execute within `10ms` (`millisecondBudget`), which is the budget for creating performant animations (16.666ms for 60 fps, but with browser overhead it's said to be 10ms).

skatcat31:
For Chrome vs Node speed: One of them has a DOM engine, the other does not. AFAIK that is what causes such a large difference in speed between the two. When the DOM engine is told to be turned off, Chrome can often be seen being just as fast as NodeJS( PhantomJS uses a rendering engine and a headful Chrome, Headless Chrome does not use a rendering engine ): https://hackernoon.com/benchmark-headless-chrome-vs-phantomjs-e7f44c6956c

Chalarangelo:
@skatcat31 has a good point. That's what I was thinking as well. `performance.now()` is pretty accurate and it actually depends on the environment, so it should give a decently accurate measurement about performance in said environment. Therefore, I think we are good to go with this.

Chalarangelo:
By the way, lucky snippet: As soon as we merged it, it's featured on the landing page. 😆

PR #646, opened 2018-04-10
[FEATURE] renameKeys

yazeedb:
## What does your PR belong to?
- [ ] Website
- [x] Snippets
- [ ] General / Things regarding the repository (like CI Integration)
- [ ] Tests

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (non-breaking improvement of a snippet)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have checked that the changes are working properly
- [x] I have checked that there isn't any PR doing the same
- [x] I have read the **CONTRIBUTING** document.

PR #652, opened 2018-04-19
Add when()

yazeedb:
Inspired by [Ramda](http://ramdajs.com/docs/#when), I absolutely love this function. It's even better, however, if you compose it by only supplying the first two parameters.

That's why my implementation returns a function expecting `x`.

If accepted, would we consider currying this function? I think the answer's "no" based on [previous discussion](https://github.com/Chalarangelo/30-seconds-of-code/pull/646#issuecomment-380253755) that this is a learning library. 

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/when.md
+++ b/snippets/when.md
@@ -1,7 +1,8 @@
 ### when
 
-Tests a value, `x`, against a predicate function. If `true`, return `fn(x)`. Else, return `x`.
-`when(pred, whenTrue)` returns a function expecting `x` for better composability.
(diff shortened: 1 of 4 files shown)


Chalarangelo:
The function looks alright as-is and, due to it being short and its general complexity low, I believe it's easy to understand the slightly more advanced footprint as it's the only part of it that is not immediately obivous. The example is also quite helpful, so I think it's quite easy to understand everything just by having a quick look at it.

People visiting the website and/or repository can combine this with adapters, such as `curry` to get the desired result.

PR #671, opened 2018-05-30
ELO snippet formatting + performance

OutThisLife:
Not assigning `arr.length` to a variable in a for loop means that the read is done on every iteration.

## Description

**Resolves** #(issue number) 

- [x] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/elo.md
+++ b/snippets/elo.md
@@ -15,17 +15,20 @@ const elo = ([...ratings], kFactor = 32, selfRating) => {
   const expectedScore = (self, opponent) => 1 / (1 + 10 ** ((opponent - self) / 400));
   const newRating = (rating, i) =>
     (selfRating || rating) + kFactor * (i - expectedScore(i ? a : b, i ? b : a));
+    
   if (ratings.length === 2) {
     return [newRating(a, 1), newRating(b, 0)];
-  } else {
-    for (let i = 0; i < ratings.length; i++) {
-      let j = i;
-      while (j < ratings.length - 1) {
-        [ratings[i], ratings[j + 1]] = elo([ratings[i], ratings[j + 1]], kFactor);
-        j++;
-      }
+  } 
+  
+  for (let i = 0, l = ratings.length; i < l; i++) {
+    let j = i;
+    
+    while (j < l - 1) {
+      [ratings[i], ratings[j + 1]] = elo([ratings[i], ratings[j + 1]], kFactor);
+      j++;
     }
   }
+  
   return ratings;
 };
 ```
--- a/snippets/elo.md
+++ b/snippets/elo.md
@@ -22,7 +22,7 @@ const elo = ([...ratings], kFactor = 32, selfRating) => {
     let j = i;
     while (j < len - 1) {
       j++;
-      [ratings[i], ratings[j + 1]] = elo([ratings[i], ratings[j + 1]], kFactor);
+      [ratings[i], ratings[j]] = elo([ratings[i], ratings[j]], kFactor);
     }
   }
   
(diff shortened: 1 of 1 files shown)


Chalarangelo:
Thanks for taking the time to contribute!

Not sure if the performance enhancement is significant enough to be worth the changes (which could be argued to harm readability just a tiny bit). Firefox performance shows a different of about 1%, whereas Chrome has already optimized this in its engine, so there is absolutely no performance difference. 

JSPerf for this PR can be found [here](https://jsperf.com/elo-comparsion).

I'll mark as `opinions needed`, see what others think.

skatcat31:
There's other readability we can address on this question before that... These changes do bring in the proper placing and return structure of a guard statement(if statement that short circuits) meaning the else can be removed.

The 'being less readable' though is a little of a misnomer. For loops are anything BUT readable. `(Decleration; Comparison; Step)` is a footprint that will always be... awful. It's why you see in a LOT of places people taking declaration out ENTIRELY before the for loop and using footprints like `(; index < length; index++)` or using do...while loops before the days of the Array methods.

So my opinion is that it's already unreadable... I do see a need to rename l as length though so it's not lost what it is... possibly change it to index and jndex(common inside jokes) if we want to improve readability.

After all readable programs read like stories. It's why promises and composition are so popular:

Promise.resolve(compare(a,b)).then(winner_takes_all)

So in the end, declaring it can help redability. It probably just needs to be fleshed out a bit so that we don't get coded confusing code like

```js
return i < l && a < b ? a : r
```

skatcat31:
Example of making it more readable:

```js
for (let index = 0, length = ratings.length; index < length; index++) {
  let jndex = index;
 
  while (jndex < length - 1) {
    [ratings[iindex], ratings[jndex + 1]] = elo([ratings[index], ratings[jndex + 1]], kFactor);
    jndex++;
  }
}
```

This is eschewing more updated coding practices to illustrate a point about readability being relative

snippets/elo.md
```diff
@@ -15,17 +15,20 @@ const elo = ([...ratings], kFactor = 32, selfRating) => {
...
-  } else {
-    for (let i = 0; i < ratings.length; i++) {
-      let j = i;
-      while (j < ratings.length - 1) {
-        [ratings[i], ratings[j + 1]] = elo([ratings[i], ratings[j + 1]], kFactor);
-        j++;
-      }
+  } 
+  
+  for (let i = 0, l = ratings.length; i < l; i++) {
```
skatcat31: While it leads to a speed improvements I think we do to much short cutting of names into abstracts here. We shoudl probably expand the names to make this more readable than a bunch of single letters

snippets/elo.md
```diff
@@ -15,17 +15,20 @@ const elo = ([...ratings], kFactor = 32, selfRating) => {
...
+    let j = i;
+    
+    while (j < l - 1) {
+      [ratings[i], ratings[j + 1]] = elo([ratings[i], ratings[j + 1]], kFactor);
+      j++;
```
skatcat31: If we're constantly using j+1 we should just `j++` before we execute this code or `++j` on the first call so we don't have to constantly put j+1

snippets/elo.md
Chalarangelo: `i` and `j` as iterators are fine (they are also part of the guidelines). `len` or `length` for the length should be a bit better than the current name.

snippets/elo.md
Chalarangelo: Yeah, `j++` seems logical.

fejes713: > There's more that can be done in terms of readability, but a modest speed improvement and a proper guard statement is appreciated.

I agree with this one. Speaking about readability this snippet is categorized as advanced as far as I remember so readability was always a bit tough here. We assume that more experienced devs are going to read these anyways so ... Let's leave readability to some other PR like @skatcat31  mentioned.

Thanks for contributing!

PR #672, opened 2018-05-30
Add toHash snippet

skatcat31:
## Description

A thing I've found myself using a LOT recently to create single pass closures so that I can grab a referenced object based on an identifier in another array. Useful for unfolding loops so you only have to loop over a data set once before referencing them later

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

- [x] My code follows the code style of this project. -- I may have to many spaces in my code formatting... I keep forgetting how semi standard sets that...


Chalarangelo:
I like the idea, but description could be a bit more descriptive and I think we are missing an explanation. Also, isn't the name a bit confusing or am I misunderstanding the functionality? 

*Don't get me wrong, we are gonna merge this, we just need to polish it up a tiny bit*

skatcat31:
The name is to show that an iterator passed in returns a hash(keyed data store)

I could probably clean up the description and add an explination

skatcat31:
Updated with a description, and making it Array Like

skatcat31:
sadly what brought this idea about is how often I'm joining multiple data-sources together now and how often I only want one loop over each instead of `O(nm)` loops

PR #675, opened 2018-06-05
Update `arrayToHtmlList` function

BrianDGLS:
## Description

The existing method has an unnecessary closure and string concatenation with `querySelector`.

To update I have removed the closure and used `getElementById` instead of `querySelector` combined with concatenating `'#'` to the supplied id.

Working example: [https://codepen.io/BrianDGLS/pen/MXyQoK](https://codepen.io/BrianDGLS/pen/MXyQoK)

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/arrayToHtmlList.md
+++ b/snippets/arrayToHtmlList.md
@@ -2,14 +2,12 @@
 
 Converts the given array elements into `<li>` tags and appends them to the list of the given id.
 
-Use `Array.map()`, `document.querySelector()`, and an anonymous inner closure to create a list of html tags.
+Use `Array.map()` and `document.getElementById(id)` to create a list of html tags.
 
 ```js
-const arrayToHtmlList = (arr, listID) =>
-  (el => (
-    (el = document.querySelector('#' + listID)),
-    (el.innerHTML += arr.map(item => `<li>${item}</li>`).join(''))
-  ))();
+const arrayToHtmlList = (arr, listID) => (
+  el = document.getElementById(listID),
+  arr.map(item => el.innerHTML += `<li>${item}</li>`))
 ```
 
 ```js


ZAYEC77:
@BrianDGLS 
why did you close your MR?
your solution looks clearer than current as for me

skatcat31:
@ZAYEC77 while it's more readable(and we should update this to the `getElementByID` method), it makes a global variable(huge no-no), and causes constant draws and is MUCH slower due to multiple modifications to the `innerHTML` attribute, which causes re-render and since it do done synchronously and without an animation frame will cause massive slowdown on lists of any significant size.

This can however be boiled down into a much more similar and simpler meshing of the two:

```js
const arrayToHtmlList = (arr, listID, format = ( item => `<li>${item}</li>`) ) =>
  document.getElementById(listID).innerHTML += arr.map(format).join('')
```

Of course there are still improvements that could be made here regarding performance probably... but this should only cause a single re-render.

ZAYEC77:
@skatcat31 Thanks for the response.
What about solution like this:
```js
const arrayToHtmlList = (arr, listID) => (
    document.getElementById(listID).innerHTML = arr.reduce((acc, item) => acc + `<li>${item}</li>`, ''))
```

or more readable variant:

```js
const arrayToHtmlList = (arr, listID) => {
    const html = arr.reduce((acc, item) => acc + `<li>${item}</li>`, '')
    document.getElementById(listID).innerHTML = html
}
```

or `join` will be faster than concatenation at each iteration of `reduce`?

skatcat31:
it's all flavor at this point. We should address it's form though. I'm going to migrate to a new issue about an update and whether we should change the footprint.

PR #680, opened 2018-06-05
make `difference` function a one liner

BrianDGLS:
## Description

Makes the difference function a one liner.

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/difference.md
+++ b/snippets/difference.md
@@ -5,10 +5,7 @@ Returns the difference between two arrays.
 Create a `Set` from `b`, then use `Array.filter()` on `a` to only keep values not contained in `b`.
 
 ```js
-const difference = (a, b) => {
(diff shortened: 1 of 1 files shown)


BrianDGLS:
Alternatively the second arg of `Array.prototype.filter` could be leveraged here to supply a `this` context.

The downside here is a the use of `function`.

```javascript
const difference = (a, b) => a.filter(function(x) { return !this.has(x); }, new Set(b));
```

Docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

skatcat31:
reducing this one to a safe one liner would involve an IIFE and harm readability. This is one of those cases where one lining it harms the developer, not the language

PR #688, opened 2018-07-08
Dig

hszeto:
## Description
I put together a snippet that will dig up key/value pair in a deeply nested Json and I hope that others will find it useful.

When I run `npm run tester`, dig.js and dig.test.js were created, but also I got the following:
```
> node ./scripts/tdd.js

npm ERR! Test failed.  See above for more details.
Tester: 6913.834ms
```
Not much details were given.

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

- [ ] I have checked that there isn't any PR doing the same


snippets/dig.md
```diff
@@ -0,0 +1,56 @@
...
+      }
+    }
+  }
+};
+
+dig(data, 'nickName'); // "Johnny"
+dig(data, 'hobby');    // "coding"
+dig(data, 'snacks');   // ["chips","candy","coke"]
+dig(data, 'l4');       // "4 levels deep."
+dig(data, 'blahblah')  // undefined
```
Chalarangelo: The examples are extremely lengthy, please shorten then to a few lines with a less complex data structure (usually examples are just one line).

test/testlog
```diff
@@ -3,2034 +3,2058 @@
 # Starting...
 # 339 test suites found.
```
Chalarangelo: We usually don't push the testlog as it changes pretty much on every run, but I can take care of that change before I merge it, so no worries here.

tag_database
```diff
@@ -55,6 +55,7 @@ detectDeviceType:browser
 difference:array,math
 differenceBy:array,function
 differenceWith:array,function
+dig:object,recursion
```
Chalarangelo: Tags are looking good.

snippets/dig.md
```diff
@@ -0,0 +1,56 @@
+### dig
+
+Return the value in a nested JSON object by the key.
+
+Given the key name (or target), it will look up the key in the object recursively and return the first value if found.
+
+```
+const dig = (obj, target) => {
+  if (!obj) return;
```
Chalarangelo: Type and error checking is something we avoid doing usually. I'd rather have this line removed.

snippets/dig.md
```diff
@@ -0,0 +1,56 @@
...
+
+  if (obj[target]) return obj[target];
```
Chalarangelo: I'm pretty sure, if you remove the error check above, you can wrap this whole thing in a one-liner with the ternary operator (`?`) or two and avoid the curly brackets around the function as well as the `return` statement.

snippets/dig.md
```diff
@@ -0,0 +1,56 @@
+### dig
+
+Return the value in a nested JSON object by the key.
+
+Given the key name (or target), it will look up the key in the object recursively and return the first value if found.
```
Chalarangelo: Please explain the methods used here, such as recursion of `dig` and what happens in it.

snippet-template.md
```diff
@@ -11,4 +11,4 @@ const functionName = arguments =>
 
 ```js
 functionName('sampleInput'); // 'sampleOutput'
-```
+```
```
Chalarangelo: Apparently the last line changed... I don't see anything different, but I don't care about this at all.

snippets/dig.md
```diff
@@ -0,0 +1,31 @@
+### dig
+
+Return the value in a nested JSON object by the key.
+
+Given the key name (or target), it will map through each key in the object and look for object type.
+If value is an object, dig is called recusively until the first matching key/value is found.
+
+```
+const dig = (obj, target) => {
```
skatcat31: This can be greatly shortened using or( `maybe || eventually` ), `Object.values())`, and `Array.prototype.find()`

```js
dig = (obj, target) => obj[target] || Object.values(obj).map( val => typeof val !== 'object' ? false : dig(val, target)).find(Boolean)
```

skatcat31: Good catch on the falsy!

I haven't used reduce lately to get information that might not exist so I've forgotten that's a good '1 pass' method. TBH a loop here is actually a better idea but not quite ES6+ enough for our style rules, so I keep forgetting reduce is a good compromise... the problems of doing 'always exists' development and using maybes... XD

skatcat31: Is the check for `undefined` on the recursion necessary?

atomiks: I just realised the `undefined` check beneath was pointless. Also I'm not sure about the top check either... Check the edit.

hszeto: @atomiks I moved the `if (typeof val === 'object')` to a filter. I think it makes the `reduce` easier to read. What do you think about this version? 

```
const dig = (obj, target) =>
  target in obj 
    ? obj[target] 
    : Object
        .values(obj)
        .filter(value => typeof value === 'object')
        .reduce((acc, val) => acc !== undefined ? acc : dig(val, target), undefined)
```

skatcat31: @hszeto reduce is to prevent multiple loops over the same data. We are 'reducing' the whole array to a single value. Filtering before that is an extra loop and code that we can avoid, but does make it more readable. It has a worse case of O(2n) and a best case of O(n+1) with a 'not found' case of O(n). Why go for something we don't know how long it could take with especially huge datasets possible instead of something we know how long it will take( O(n) )?

hszeto: @skatcat31 I am 100% with you in terms of performance. It's never a good idea to add more loops. In my mind, I was thinking of ways to make the code easier to read and understand for users to learn in 30 seconds.... I didn't put performance in priority. My bad. :P Anyway, I will make another adjustment and commit.

PR #709, opened 2018-08-10
Update symmetricDifference

EthanOrange:
No const
One time new Set()
One line code style

## Description

**Resolves** #(issue number) 

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/symmetricDifference.md
+++ b/snippets/symmetricDifference.md
@@ -5,11 +5,9 @@ Returns the symmetric difference between two arrays.
 Create a `Set` from each array, then use `Array.filter()` on each of them to only keep values not contained in the other.
 
 ```js
-const symmetricDifference = (a, b) => {
-  const sA = new Set(a),
-    sB = new Set(b);
-  return [...a.filter(x => !sB.has(x)), ...b.filter(x => !sA.has(x))];
-};
+const symmetricDifference = (a, b) => [
+  ...new Set([...a.filter(v => !b.includes(v)), ...b.filter(v => !a.includes(v))])
+];
 ```
 
 ```js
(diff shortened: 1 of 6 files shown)


skatcat31:
I'm not sure this was ever working as intended in either version...

From Wikipedia:

```
For example, the symmetric difference of the sets {1,2,3} and {3,4} is {1,2,4}.
```

We may want to check that the symmetric difference is correct since we should not get ANY repeats as a repeat would mean it's in both sets wouldn't it? I'm going to do some more review. We may have a bug in this snippet

Chalarangelo:
A repeat means it's twice in one of the arrays, @skatcat31.

skatcat31:
If that's what we want then we should make new snippet for unique symmetrical differences

Chalarangelo:
@skatcat31 That sounds like a bit of a convoluted name but it's what I was thinking of, too. Tests should also be more refined and better defined to make sure these edge cases are documented properly.

skatcat31:
Then we'll want some changes on this one so that it doesn't unique

Chalarangelo:
I'll follow up on this PR tomorrow, probably keep the original snippet as-is and use the PR's code to create a new snippet for `uniqueSymmetricalDifference`. I'll also add tests and examples to showcase the difference between the two.

EthanOrange:
I really want to complete this code snippet.
but my English is very bad. I can't express the difference between `symmetricDifference` and `uniqueSymmetricalDifference`. 😔

skatcat31:
`uniqueSymmetricalDifference` can be fixed to `uniqueSymmetricDifference` so that they are parodied names.

PR #711, opened 2018-08-13
[ENHANCEMENT] Optimized Longest Item function

Bruce-Feldman:
## Description

Changed longestItem function to run in O(n) rather than O(nlog(n)).
Note: The longest item returned will now be the last one in the array rather than an unspecified one (since Array.sort is not necessarily stable).

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/longestItem.md
+++ b/snippets/longestItem.md
@@ -2,10 +2,10 @@
 
 Takes any number of iterable objects or objects with a `length` property and returns the longest one.
 
-Use `Array.sort()` to sort all arguments by `length`, return the first (longest) one.
+Use `Array.reduce()` to collect the longest element.  Returns [] for empty array.
 
 ```js
-const longestItem = (...vals) => [...vals].sort((a, b) => b.length - a.length)[0];
+const longestItem = (...vals) => [...vals].reduce((a, x) => (a.length > x.length ? a : x), []);
 ```
 
 ```js
--- a/test/longestItem/longestItem.js
+++ b/test/longestItem/longestItem.js
@@ -1,2 +1,2 @@
-const longestItem = (...vals) => [...vals].sort((a, b) => b.length - a.length)[0];
+const longestItem = (...vals) => [...vals].reduce((a, x) => (a.length > x.length ? a : x), []);
 module.exports = longestItem;


skatcat31:
performance gain explanation: It went from a sort(multi pass) to a reduce(single pass) which can gain exponential speed. The return of the empty array however is breaking since this is a duck casing snippet. A reduce call would have produced functionally equivalent code and kept the context expectations(I do after all receive an array I can modify, with this version if they're all empty, I will not receive the expected array and it will not modify in place if desired). We should probably fix that

Chalarangelo:
@skatcat31 Good catch, that seems wrong. We should do something like this then:

```js
const longestItem = (...vals) => [...vals].reduce((a, x) => ((a ? a.length : 0) > x.length ? a : x), undefined);
```

Seems a bit weird, but I think it's a valuable thing to showcase as an edge case in `Array.reduce()`.

skatcat31:
@Chalarangelo yours is still passing an optional initial. If the initial is note passed to reduce, it will instead use the first element...

```js
const longestItem = (...vals) => [...vals].reduce((a, x) => a.length > x.length ? a : x);
```

skatcat31:
We just now need to confirm that if done the same array would be returned from both versions if the arrays are all same length(which we might not be able to do due to the way sort is semi unpredictable about a length of ten). In which case we should decide if it should be first or last returned(`>` or `>=` respectively)

Bruce-Feldman:
@skatcat31
```
const longestItem = (...vals) => [...vals].reduce((a, x) => a.length > x.length ? a : x);
```
will throw a type error with no arguments.  This is different than the alternatives that return some value (originally undefined, which I think is probably the best).

skatcat31:
interesting. I'm guessing that works because it forces a first argument so the array will always have a length. Good catch!

Chalarangelo:
The `(val, ...vals)` footprint looks a bit weird to newcomers, but it's a nice opportunity to teach people about that sort of thing. I think this is probably the best possible option, right?

Bruce-Feldman:
Some more thoughts:

With the benefit of more reflection, I want to point out that my suggestion of `const longestItem = (val, ...vals) => [val, ...vals].reduce((a, x) => a.length > x.length ? a : x);` replicates the original functionality of ` const longestItem = (...vals) => [...vals].sort((a, b) => b.length - a.length)[0]; ` to the best of my ability to test.

Fortunately (or unfortunately), these functions both have the behaviour that, if they are passed exactly one non-conforming element (e.g. null, undefined, a number), they return that element without exception.  However, if that non-conforming element is passed as part of 2 or more parameters, then the functions bail with a TypeError.

If we want to enforce the fact that all provided parameters must have a length property (and still return undefined when passed no parameters), we can use a modified version of @skatcat31 
suggestion:
```
const longestItem = (...vals) => [...vals].reduce((a, x) => a.length > x.length ? a : x, vals[0]);
```

atomiks:
I don't see the point of `(...vals) => [...vals]`, same problem with the previous PR.

skatcat31:
I'm thinking we may want to land with

```js
const longestItem = (val, ...vals) => [val, ...vals].reduce((a, x) => x.length > a.length ? x : a);
```

This should force the fact that if a is the same length as x, a will be returned instead leading to a first found return like @Chalarangelo suggested

Bruce-Feldman:
@atomiks I don't understand your objection.  Could you please elaborate?

Chalarangelo:
I think the footprint @skatcat31 suggests is probably the best one we can use. If nobody objects, we should get started on writing tests about it.

skatcat31:
tests cases we'll need:

returns first found of all similar:

```js
let a = [];
longestItem(a, [], []) === a == true;

a = [1];
longestItem(a, [2], [3]) === a == true;
longestItem(a, [2], [3])[0] === 1 == true;
longestItem('1', '2', '3') === '1' == true;
longestItem('2', '1', '3') === '2' == true;
```

returns undefined if no arguments passed due to being a reduce call on a single element:
```js
longestItem() === undefined == true // just test in browser, working
```


throws on multiple undefined passed due to type errors
```js
try{
  longestItem(undefined, undefined);
  return false;
} catch (e) {
  return true;
} // working, just tested in browser
```

does not throw on mixed types as long as none are undefined
```js
try{
  let arr = [];
  let ab = 'ab'
  longestItem('a', 1, arr) === arr == true //  works because longest length is 1
  longestItem(ab, 1, arr) === ab == true // works because longest length is 2
  return true;
} catch (e) {
  return false;
}
```

we could also easily make this throw on any undefined as well and just invert those cases:
```js
const longestItem = (val, ...vals) => (val.length, [val, ...vals].reduce((a, x) => x.length > a.length ? x : a));
```

is there any test case I missed?

Chalarangelo:
Continue the discussion under #717 please.

PR #712, opened 2018-08-13
[ENHANCEMENT] Simplified nth Element

Bruce-Feldman:
## Description

Simplified the implementation of nth Element to avoid unnecessary branching.
Updated documentation to reflect reality - the function would return undefined (not []) if the index was not in range.  This is maintained in the current implementation.

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/nthElement.md
+++ b/snippets/nthElement.md
@@ -3,11 +3,11 @@
 Returns the nth element of an array.
 
 Use `Array.slice()` to get an array containing the nth element at the first place.
-If the index is out of bounds, return `[]`.
+If the index is out of bounds, return `undefined`.
 Omit the second argument, `n`, to get the first element of the array.
 
 ```js
-const nthElement = (arr, n = 0) => (n > 0 ? arr.slice(n, n + 1) : arr.slice(n))[0];
+const nthElement = (arr, n = 0) => (n === -1 ? arr.slice(n) : arr.slice(n, n + 1))[0];
 ```
 
 ```js
--- a/test/nthElement/nthElement.js
+++ b/test/nthElement/nthElement.js
@@ -1,2 +1,2 @@
-const nthElement = (arr, n = 0) => (n > 0 ? arr.slice(n, n + 1) : arr.slice(n))[0];
+const nthElement = (arr, n = 0) => (n === -1 ? arr.slice(n) : arr.slice(n, n + 1))[0];
 module.exports = nthElement;


Chalarangelo:
This changes behavior in the snippet itself, as the edge-case returns `undefined` instead of `[]` now. To be honest, this is not an unwelcome change as it feels more consistent with the codebase and it was never well documented and tested properly to begin with. I will run some tests, see if everything works as expected and probably merge this afterwards.

Bruce-Feldman:
I can't seem to reproduce nthElement's returning [] on out of bounds values (https://repl.it/repls/ProductiveAwesomeLocation).  Could you describe the inputs that cause this behaviour so I can try to keep the functionality consistent?

Chalarangelo:
It's quite possible the snippet's description was wrong in this case. I can't seem to reproduce it either.

Bruce-Feldman:
Thanks for confirming.

skatcat31:
Wasn't the `n, n+1` case for performance concerns with large arrays? If that's the case wouldn't it be better to update the description to note that and introduce that idea that large arrays are sometimes problems?

Chalarangelo:
@skatcat31 We might need to run some perfs on arrays with a half a million objects. I think it's not that much a performance issue, but rather a memory issue. Might be worth checking out and finding what kind of limitations there are.

skatcat31:
... come to think of it isn't this just a really badly performing example of infinitely many other ways to implement this?

```js
array.find((val, ind) => ind == n)
array[n]
etc()
```

Bruce-Feldman:
This all makes sense.  Should we optimize for the case where n is negative as well?

skatcat31:
> Should we optimize for the case where n is negative as well?

That's a great reason to keep `.slice`!

Bruce-Feldman:
I've been thinking about how to implement the negative case with minimal memory cost, it seems as though:
```
[1, 2].slice(-1, 0) // []
[1, 2].slice(-2, -1) // [1]
```
This means that a working version should probably look something like:
```
const nthElement = (arr, n = 0) => (n === -1 ? arr.slice(n) : arr.slice(n, n + 1))[0];
```
This has the additional benefit that it will now reject out of range negative values as well.

skatcat31:
That version has other issues as well sadly(out of bounds breaking, more memory usage on positive index, etc.)

probably more like:

```js
const nthElement = (arr, n = 0) => (n == 0 ? arr.slice(n) : arr.slice(n, n + 1))[0];
```

Bruce-Feldman:
@skatcat31 
The version you propose doesn't work with n = -1:
```
const nthElement = (arr, n = 0) => (n == 0 ? arr.slice(n) : arr.slice(n, n + 1))[0];
nthElement([1], -1) // undefined
```
Did you mean n < 0 instead of n == 0?  Then we are right back where we started, which is fine with me.  I guess the question would be if we cared about optimizing memory consumption of the current implementation for large negative values.

That said, I don't understand your objections with the following:
```
const nthElement = (arr, n = 0) => (n === -1 ? arr.slice(n) : arr.slice(n, n + 1))[0];
```
It should not use extra memory on positive indexes.  I did note that it breaks backwards compatibility - large negative indices now return undefined.  I am not sure if this is a feature or a bug (I am leaning towards a feature).

skatcat31:
ah never mind I misread the snippet. You're right I shouldn't have any objections.

Bruce-Feldman:
Is there something I can do to help this PR get through?

skatcat31:
Nope you've done your part. I try not to merge things unless they fix something because I want @Chalarangelo and the others to get the credit. I do more of the 'what are we missing or what concerns do we have here' stuff. Sometimes if they're busy I'll merge if it's been a week since last change. However this one looks good and has had enough deliberation

Chalarangelo:
Ah, it's August and I'm sort of lazy lately. I try to merge PRs daily but then I get distracted a lot. Thanks @skatcat31 for merging this one! 👍

Bruce-Feldman:
Thanks for getting to this!

PR #714, opened 2018-08-13
Update mapObject

EthanOrange:
short code and easy to kown it how to work

## Description

**Resolves** #(issue number) 

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:


EthanOrange:
context issues?
in async function?

skatcat31:
There are ways to do this that are simpler than the inner context closure declaration that I used that are just as compatible. When we created this snippet it was a smosh of a `zipObject` snippet that combined the two steps usually separated into a closure. We left it that way after another suggestion like this as a technical showcase, however there are many updates that can be made to it.

As for the provided snippet it is similar to one provided before. I think @Chalarangelo might be misreading something since this just removes the array value nesting in favor of api chaining, something we've run into context problems with before. More in depth testing and a new description would also be needed. This snippet is ripe for updating, but we may want to keep an archive version so that people still have a place to learn about IIFEs in fat arrow functions

Issue #690, opened 2018-07-16
[MAJOR FEATURE] Glossary


Chalarangelo:
To enhance the website and repository and provide more value to newcomers that might be unfamiliar with JS, programming, web development or simply some terminology, we will be launching a glossary of JS,programming and webdev terms on a separate page and file. Here's the list of things we need to do before we launch this feature, as well as an incomplete list of terms we should document:

## The chores

- [x] Create a `librarian` script (I'd love to call it that, but we can call it something more sensible), that adheres to our current style, build a README in `/glossary`. assigned: _@flxwu_
- [x] Add a `keyword_database` and its respective script for glossary terms to make search and term tagging more dynamic. assigned: _@flxwu_
- [x] Update our `webber` script to generate a `glossary` page, similar to any tag page.
- ~~Launch with a couple of Medium articles to promote this new feature (I'll definitely write a comprehensive guide to JS and the frontend for newcomers, someone else can write another one).~~
- ~~If possible, use a script to add short tooltips to our snippets whenever a word from the glossary is present to further enhance the learning experience.~~ *Maybe later*

## The template

Before moving forward, the barebones guidelines for glossary terms are:

- Keep it short and simple.
- File names have the first letter capitalized, are in snake-case and we should create a file for each term. Titles should have the first letter capitalized and be in space case.
- Links might be added later down the line to **trusted resources only, such as MDN** (and maybe W3Schools), at the bottom of the file as a list.
- Probably no code is necessary and explanation length should be around 2-3 sentences (one line per sentence as usual).

## The term list

In general, I would like this to be a complete-ish guide to JS, programming and webdev. What goes in it?

- Terms that are JS-specific (such as keywords - `this` comes to mind).
- General terms related to programming, algorithms and programming paradigms (OOP, JSON, currying, functional programming).
- Terms that are commonly used in the JS ecosystem (callbacks, AJAX).
- Well-known resources and packages in the JS ecosystem (React, npm, MDN).
- Terms related to web development, HTML, CSS and other complementary technologies (DOM, selectors).

A very incomplete list of terms that should be documented:

- *All the JS keywords we can document, add them to the list as we go*
- [x] Callback
- [x] String
- [x] Template literals
- [x] Boolean
- [x] Integer
- [x] Array
- [x] Function
- [x] Object
- [x] First-order function
- [x] Functional programming
- [x] Object-oriented programming
- [x] Prototypical inheritance
- [x] Dynamic programming
[...]

PR #718, opened 2018-08-17
[GLOSSARY] Glossary additions

Chalarangelo:
## Description

In continuation of the development of #690.

List of terms (13):

- Variable
- Constant
- NPM
- Yarn
- Express
- React
- Git
- Repository
- Argument
- Function
- ES6
- IP
- HTML

- [ ] Website
- [ ] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/glossary/Argument.md
+++ b/glossary/Argument.md
@@ -1,3 +1,4 @@
 ### Argument
 
 An argument is a value passed as an input to a function and can be either a primitve or an object.
+In JavaScript, functions can also be passed as arguments to other functions.
--- a/glossary/Argument.md
+++ b/glossary/Argument.md
@@ -1,4 +1,4 @@
 ### Argument
 
-An argument is a value passed as an input to a function and can be either a primitve or an object.
+An argument is a value passed as an input to a function and can be either a primitive or an object.
 In JavaScript, functions can also be passed as arguments to other functions.
--- a/glossary/Argument.md
+++ b/glossary/Argument.md
@@ -1,0 +1,3 @@
+### Argument
+
+An argument is a value passed as an input to a function and can be either a primitve or an object.
--- a/glossary/Function.md
+++ b/glossary/Function.md
@@ -1,0 +1,5 @@
+### Function
+
+Functions are self-contained blocks of code with their own scope, that can be called by other code and are usually associated with a unique identifier.
+Functions accept input in the form of arguments and can optionally return an output. 
+JavaScript functions are also objects.
--- a/glossary/Function.md
+++ b/glossary/Function.md
@@ -1,5 +1,5 @@
 ### Function
 
 Functions are self-contained blocks of code with their own scope, that can be called by other code and are usually associated with a unique identifier.
-Functions accept input in the form of arguments and can optionally return an output. 
+Functions accept input in the form of arguments and can optionally return an output (if no `return` statement is present, the default value of `undefined` will be returned instead). 
 JavaScript functions are also objects.
--- a/glossary/Git.md
+++ b/glossary/Git.md
@@ -1,4 +1,4 @@
 ### Git
 
 Git is an open-source version control system, used for source code management.
-Git allows users to clone and edit repositories on their local machines, without having to change the master repository.
+Git allows users to copy (clone) and edit code on their local machines, before merging it into the main code base (master repository).
--- a/glossary/Git.md
+++ b/glossary/Git.md
@@ -1,0 +1,4 @@
+### Git
+
+Git is an open-source version control system, used for source code management.
+Git allows users to clone and edit repositories on their local machines, without having to change the master repository.
--- a/glossary/React.md
+++ b/glossary/React.md
@@ -1,0 +1,4 @@
+### React
+
+React is a frontend framework, that allows developers to create component-based user interfaces.
+React separates view and state, utilizing a virtual DOM to update the user interface.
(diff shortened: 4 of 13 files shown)


glossary/Argument.md
```diff
@@ -0,0 +1,3 @@
+### Argument
+
+An argument is a value passed as an input to a function and can be either a primitve or an object.
```
flxwu: You may also add here that in JS, you can pass a function as function argument as well as functions are first-class objects in JavaScript

glossary/Function.md
```diff
@@ -0,0 +1,5 @@
+### Function
+
+Functions are self-contained blocks of code with their own scope, that can be called by other code and are usually associated with a unique identifier.
+Functions accept input in the form of arguments and can optionally return an output. 
+JavaScript functions are also objects.
```
flxwu: could amend it here as well

glossary/Git.md
```diff
@@ -0,0 +1,4 @@
+### Git
+
+Git is an open-source version control system, used for source code management.
+Git allows users to clone and edit repositories on their local machines, without having to change the master repository.
```
flxwu: this is pretty vague... I would reword it 😅  also, I would change the last sentence to "the remote master repository"

glossary/React.md
```diff
@@ -0,0 +1,4 @@
+### React
+
+React is a frontend framework, that allows developers to create component-based user interfaces.
```
flxwu: would add "dynamic" here as keyword

glossary/Repository.md
```diff
@@ -0,0 +1,3 @@
+### Repository
+
+In a version control system, a repository (or repo for short) is a data structure that stores metadata for a set of files.
```
flxwu: i.e. a project? 😅

glossary/Yarn.md
```diff
@@ -0,0 +1,4 @@
+### Yarn
+
+Yarn is a package manager for the JavaScript programming language.
```
flxwu: maybe add "by facebook" here?

glossary/Function.md
Chalarangelo: For this term, I think stating that they're object is sufficient..

glossary/Argument.md
Chalarangelo: Definitely worth stating here, I'll update the term.

glossary/Git.md
Chalarangelo: Honestly, it's based on Wikipedia and other sources. It's the least vague I could write. Could you just reword it on the same branch? Maybe you can explain it better than that.

glossary/React.md
Chalarangelo: Will be done.

glossary/Repository.md
Chalarangelo: Yup.

glossary/Yarn.md
Chalarangelo: Honestly, I prefer not to say who is behind what as packages have a slighttendency to change ownership over time and it might get outdated or sound like unnecessary promo. I'm thinking we can link to some external content from the glossary like said packages etc. so it will be obvious to people what is what.

Chalarangelo:
@flxwu Chanes made as requested and comments added for the terms I did not update.

glossary/Function.md
skatcat31: > Functions are special objects in JavaScript that have their own local scope and can be called by other pieces of code. They are usually assigned a unique identifier, but sometimes an anonymous function is used. Functions can optionally accept inputs and return outputs.

Combines both but may be confusing. It should also be noted that a Function call will always return. If one is not explicit, the implicit return of `undefined` is returned.

glossary/Argument.md
skatcat31: I'd put it as a note like
> ... or an object(a function is an object)

or some other way

Chalarangelo: Hmmm, I could rephrase the first sentence to be something along the lines of *An argument is a value passed as an input to a function and can be either a primitve, a function or an object.* Seems less weird that way.

glossary/Function.md
Chalarangelo: I prefer the phrasing asis. The return functionality, however, should be explained. I'll update.

glossary/Git.md
Chalarangelo: Updated. Tried to keep the terminology present as it could be of use to beginners.

glossary/Yarn.md
Chalarangelo: Currently it only works with JS. Plus, this is a JS-specific repo, so we can quietly omit that information at least until this is a stable feature.

Chalarangelo:
@skatcat31 Made some updates, read my comments for the rest.

PaulReiber:
primitve

Chalarangelo:
@PaulReiber fixed, thanks for the heads up!

glossary/Yarn.md
skatcat31: In that case it might be better as

> Yarn is a package manager made by Facebook. It can be used as an alternative to the npm package manager and is compatible with the public NPM registry.

reason the second NPM is capitalized: It's actually the company name, no the tool

Chalarangelo:
@skatcat31 @flxwu I think most terms are ready to go, what do you guys think?

PR #719, opened 2018-08-20
[FEATURE] Create pipeLog.md

3dos:
## Description
I added a new snippet

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/pipeLog.md
+++ b/snippets/pipeLog.md
@@ -1,6 +1,6 @@
 ### pipeLog
 
-Use `console.log` in a pipeline as this function simply encloses it and returns the passed value.
+Use `console.log` in a pipeline as this function simply encloses it and returns the passed value. This is especially useful for debugging when you want to log a variable's value before its usage.
 
 Logs a value and returns it.
 
--- a/snippets/pipeLog.md
+++ b/snippets/pipeLog.md
@@ -1,0 +1,13 @@
+### pipeLog
+
+Use `console.log` in a pipeline as this function simply encloses it and returns the passed value.
+
+Logs a value and returns it.
+
+```js
+const pipeLog = data => console.log(data) || data
+```
+
+```js
+pipeLog(1); // logs `1` and returns `1`
+```


flxwu:
Thanks for the PR!

TBH, I don't really see the value in writing in in one statement instead of first logging and then returning the val. Opinions? @Chalarangelo @fejes713

3dos:
Hi, thanks for the comment. It's a little helper function I use in development while using Ramda's `pipe` or `compose` without breaking the pipeline so I thought it would be useful among the other snippets.

Obviously feel free to reject this PR :)

flxwu:
right for debugging purposes it indeed would be quite convenient 👍

3dos:
Thanks for the update in the description !

snippets/pipeLog.md
```diff
@@ -0,0 +1,13 @@
+### pipeLog
+
+Use `console.log` in a pipeline as this function simply encloses it and returns the passed value. This is especially useful for debugging when you want to log a variable's value before its usage.
+
+Logs a value and returns it.
+
+```js
+const pipeLog = data => console.log(data) || data
```
Chalarangelo: Semistandard dictates adding a semicolon (`;`) at the end of the statement.

Chalarangelo: I am also not entirely certain that this is the only way to achieve this. I think something along the lines of `console.log(data), data` would have the same effect.

snippets/pipeLog.md
```diff
@@ -0,0 +1,13 @@
+### pipeLog
+
+Use `console.log` in a pipeline as this function simply encloses it and returns the passed value. This is especially useful for debugging when you want to log a variable's value before its usage.
+
+Logs a value and returns it.
```
Chalarangelo: This should be the first part of the description of the snippet, detailing the functionality.

snippets/pipeLog.md
```diff
@@ -0,0 +1,13 @@
+### pipeLog
+
+Use `console.log` in a pipeline as this function simply encloses it and returns the passed value. This is especially useful for debugging when you want to log a variable's value before its usage.
```
Chalarangelo: This should be expanded to explain the how of this snippets inner workings and be less about use-cases and more about the details of how the code works (although there is very little to see here anyways). It should also come after the functionality description.

snippets/pipeLog.md
skatcat31: as stated above, a comma may actually be the preferred way to handle this.

Chalarangelo:
@skatcat31 `console.debug` could be a better option actually.

snippets/pipeLog.md
3dos: I didn't know about the comma operator (every time, it's a good time to learn). Now you mention it, it totally makes sense. But I get an `Unexpected use of comma operator` from es-lint while using

```js
const debug = obj => (console.log(obj), obj);
```

I can't find any help on the [`no-sequence`](https://eslint.org/docs/rules/no-sequences) page of es-lint docs.

skatcat31:
No, it couldn't. `console.verb` is a pattern of logging. They are usually semantic and have some differences, but in the end most do the same thing: write to standard out. It does not allow stepping, it does not allow reassignment or modification, it does not allow the true power of debugging. Log debugging is NOT a good way to debug. With powerful tools built into every browser and NodeJS itself with a simple `debugger` they aren't comparable, especially if I can't step out of a function to see its calling context. That's more often than not what really matters. I'd be hard pressed to ever justify a `console` statement, and in fact most developers are. It's often why lint rules have it set at `error`.

Still that's besides the point. I was not against this, I'm just saying it's not necessarily something I'd put in the main repo do to the intended use of the snippet and thus I'm not the best judge of it. It's helpful for people who don't know of the `debugger` statement, but it should be noted that `console` logging to inspect something is considered a bad practice and that you should use a debugger instead.

As a point in it's favor however I was going back through the Ramda gitter and noticed about 3 or 4 questions about how to do just this. So take it as you will.

Chalarangelo:
@skatcat31 It's a code smell definitely. I think this belongs in the archive more than anything else.

3dos:
@skatcat31 thank you for your insight. I'll definitely take a look at the `debugger`. Yet another thing I learned today \o/

snippets/pipeLog.md
flxwu: Did you read through the whole docs page? 😉 ``
Rule Details

This rule forbids the use of the comma operator, with the following exceptions:

    In the initialization or update portions of a for statement.
    If the expression sequence is explicitly wrapped in parentheses.

Chalarangelo:
Merging with intent to move to the archive.

PR #721, opened 2018-08-25
[FEATURE]  add deepFreeze

fejes713:
## Description
A very useful snippet for freezing complex objects.

screen shot 2018-08-25 at 6 03 45 pm

- [ ] Website
- [x] Snippets
- [ ] Tests

## Types of changes

## Checklist:


snippets/deepFreeze.md
```diff
@@ -0,0 +1,38 @@
+### deepFreeze
+
+Deep freezes an object.
+
+Calls `Object.freeze(obj)` recursively on all unfrozen properties of obj that are `typeof` function or object.
```
Chalarangelo: Phrasing improvement suggestion:

Call `Object.freeze(obj)` recursively on all unfrozen properties of `obj` that are identified as functions or objects, using `typeof`.

Just my opinion, there might be an even better way to phrase this.

snippets/deepFreeze.md
```diff
@@ -0,0 +1,38 @@
...
+      obj.hasOwnProperty(prop) &&
+      obj[prop] !== null &&
+      (typeof obj[prop] === 'object' || typeof obj[prop] === 'function') &&
+      !Object.isFrozen(obj[prop])
+    ) {
+      deepFreeze(obj[prop]);
+    }
+  });
+
+  return obj;
```
Chalarangelo: I think that most of this could be converted to a ternary operator, but readility might be terrible.

snippets/deepFreeze.md
```diff
@@ -0,0 +1,38 @@
...
+};
+```
+
+```js
+'use strict';
+
+const a = Object.freeze([1, [2, 3]]);
+
+a[0] = 3; // not allowed
+a[1][0] = 4; // allowed but shouldn't be
```
Chalarangelo: Comparing to `Object.freeze()` in the example is a tiny bit unnecessary, I think.

tag_database
```diff
@@ -48,6 +48,7 @@ debounce:function
 decapitalize:string,array
 deepClone:object,recursion
 deepFlatten:array,recursion
+deepFreeze:object,recursion
```
Chalarangelo: Maybe this fits under `advanced`? It's not very difficult, but the concepts are a bit more advanced.

test/deepFreeze/deepFreeze.test.js
```diff
@@ -0,0 +1,6 @@
+const expect = require('expect');
+const deepFreeze = require('./deepFreeze.js');
+
+test('deepFreeze is a Function', () => {
+  expect(deepFreeze).toBeInstanceOf(Function);
+});
```
Chalarangelo: We definitely need at least a few tests, but I will pitch in on these if you have trouble adding them due to time limitations or anything else.

fejes713: Sorry, I had very little time + I had no clue how to test it. 😢

tag_database
fejes713: I wouldn't put it under advanced,  but you might be right too. 

TBH I see deepFreeze as an essential snippet when writing functional JS.

snippets/deepFreeze.md
fejes713: Agree. Should have put that into the PR description.

snippets/deepFreeze.md
fejes713: > but readility might be terrible.

It would. This isn't very long either.

skatcat31:
why are the packages being rolled back or frozen?

snippets/deepFreeze.md
```diff
@@ -0,0 +1,38 @@
...
+
+```js
+const deepFreeze = obj => {
+  Object.freeze(obj);
+
+  Object.getOwnPropertyNames(obj).forEach(function(prop) {
+    if (
+      obj.hasOwnProperty(prop) &&
+      obj[prop] !== null &&
+      (typeof obj[prop] === 'object' || typeof obj[prop] === 'function') &&
```
skatcat31: Why not use
```js
obj[prop] instanceof Object
```

skatcat31: This would require a change to the snippet description since you wouldn't be using typeof anymore

fejes713:
@flxwu  I probably made a stupid mistake - I've pulled the changes, made the changes, and then did fresh `npm i`. Can you do that `felix git magic` so we revert those changes?

flxwu:
@fejes713 sure, 
```
git checkout master -- package-lock.json
```

tag_database
Chalarangelo: Let's keep it as not `advanced` for now and we'll see what the feedback says.

snippets/deepFreeze.md
Chalarangelo: I agree that `instanceof` is probably better and shorter. Let's give it a shot.

snippets/deepFreeze.md
Chalarangelo: If we go with `instanceof` as suggested maybe it can work with a ternary. I'll check and report back with reults.

test/deepFreeze/deepFreeze.test.js
Chalarangelo: I'll add some tests.

snippets/deepFreeze.md
```diff
@@ -0,0 +1,38 @@
+### deepFreeze
+
+Deep freezes an object.
+
+Calls `Object.freeze(obj)` recursively on all unfrozen properties of obj that are `typeof` function or object.
+
+```js
+const deepFreeze = obj => {
+  Object.freeze(obj);
```
Chalarangelo: This can be moved into the return statement, as the passed `obj` will be returned from `Object.freeze()`.

snippets/deepFreeze.md
```diff
@@ -0,0 +1,38 @@
...
+
+  Object.getOwnPropertyNames(obj).forEach(function(prop) {
+    if (
+      obj.hasOwnProperty(prop) &&
+      obj[prop] !== null &&
```
Chalarangelo: Provided we only want `obj[prop] instanceof Object` properties, do we need to check for `null`? I mean it's (as far as a I can tell) impossible for a property to be `null` and `instanceof Object` at the same time.

Chalarangelo:
Having toyed with the code a bit, I propose that we change the snippet to this:

```js
const deepFreeze = obj => {
  Object.getOwnPropertyNames(obj).forEach(prop => {
    if (obj.hasOwnProperty(prop) && obj[prop] instanceof Object && !Object.isFrozen(obj[prop]))
      deepFreeze(obj[prop]);
  });
  return Object.freeze(obj);
}
```
**Changes:**

- No `null` checking, it's not possible to have a property that is both an `Object` and also `null`.
- Return the frozen object at the end, no extra statement at the start.
- Use `instanceof`.


**Questions:**

- Combining `getOwnPropertyNames` and `hasOwnProperty` essentially returns `Object.keys()`, right? If so, we could trim the `if` statement significantly.

snippets/deepFreeze.md
skatcat31: You are correct in assuming that null is not an instance of Object

Chalarangelo: @skatcat31 Check my proposal below, I think it makes the code a lot shorter and easier to understand.

skatcat31:
For reference, this is an implicit return version using `instanceof` and `Object.keys`

```js
const deepFreeze = obj =>
  Object.keys(obj).forEach(prop =>
    !obj[prop] instanceof Object || Object.isFrozen(obj[prop])
      ? null
      : deepFreeze(obj[prop])
  )
|| Object.freeze(obj);
```

There are some stylistic choices for kerning alignment(moving the final return to the same kerning level) that may want to be fixed if we adopt this version. The ternary was inverted so that the short case comes first to be kept in line with our other snippets.

In either case I have a question: Would we want to freeze any functions on these objects? While they are `Objects` and they are enumerable own, it is still a question to consider. Since own and enumerable I left them in since they are directly on the Object and not on the prototype

Chalarangelo:
@skatcat31 The ternary with `Object.keys()` looks a lot more manageable. I think it's not that hard to understand and makes it a lot easier to learn how to do this.

Kerning alignment etc are all handled by our `linter`, so whatever the script likes.

I think we want to freeze functions for consistency. Deep freezing means making the object entirely unmutable if possible, so freezing functions sounds like a good fit. We could add a flag for it, but that makes the snippet a lot longer and more complicated for everyone.

skatcat31:
```js
const deepFreeze = (obj, freezeFunctions = true) =>
  Object.keys(obj).forEach(prop =>
    !obj[prop] instanceof Object || Object.isFrozen(obj[prop])
      ? null
      : !obj[prop] instanceof Function || !freezeFunctions ?
        ? null
        : deepFreeze(obj[prop])
  )
|| Object.freeze(obj);
```

adds a few lines and a confusing default(true so that all freeze on default) and a confusing enclosed ternary. Like @Chalarangelo said it might not be a good idea to allow the flag. It gets a little confusing

Chalarangelo:
30 seconds of code is not about a ton of configuration, we usually go for the most common use-case. Freezing functions sounds like the way to go.

skatcat31:
Then I'd recommend the version without the flag

flxwu:
@fejes713 Please reset the lockfile and maybe adapt the snippet to the one @Chalarangelo and @skatcat31 agreed on:

```
const deepFreeze = obj =>
  Object.keys(obj).forEach(prop =>
    !obj[prop] instanceof Object || Object.isFrozen(obj[prop])
      ? null
      : deepFreeze(obj[prop])
  )
|| Object.freeze(obj);
```
which looks most reasonable to me as well.

fejes713:
@flxwu I've updated the snippet, tests are still missing, and I've run `git checkout master -- package-lock.json` so hopefully the package lock is not all right.

flxwu:
Perfect! 👌  @fejes713. What about `test/testlog`?

Chalarangelo:
@flxwu I resolved the conflict there, no worries.

PR #726, opened 2018-08-30
Simplify initialize2DArray

siarhei-zharnasek:
Simplify initialize2DArray [ENHANCEMENT]

No need to use Array.prototype.map and Array.prototype.fill because of native Array.prototype.from callback

- [ ] Website
- [X] Snippets
- [ ] Tests

## Types of changes
- [X] Enhancement (non-breaking improvement of a snippet)

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/snippets/initialize2DArray.md
+++ b/snippets/initialize2DArray.md
@@ -2,7 +2,7 @@
 
 Initializes a 2D array of given width and height and value.
 
-Use `Array.map()` to generate h rows where each is a new array of size w initialize with value. If the value is not provided, default to `null`.
+Use `Array.from()` callback to generate h rows where each is a new array of size w initialized with value `val`. If the value is not provided, default to `null`.
 
 ```js
 const initialize2DArray = (w, h, val = null) =>


skatcat31:
I think tagging as on hold is hepful. Fill is very esoteric, and Array.from is just as esoteric. I have used it once, and that was before NodeList iterators were attached to the type, which I usually just get around by calling or apply the Array Prototype to them.

I think updating to the newer code is helpful, but if we do we should document it a bit more in the snippet markdown, which may make the snippet explanation a bit long. Maybe instead we just include a link to the MDN article for it because once you learn it it's really self explanitory

Chalarangelo:
I really think the new code is a bit hard to understand and requires more effort to figure out if you do not know about the `from` callbacks. `map` and `fill` are a lot more self-explanatory, as you see `map` creating the arrays and `fill` filling them like we do for 1D arrays. For consistency's sake and to make things easier to understand, I think we should stay on `map` and `fill`. 

We could link to an alternate version with `fill`... Which raises an issue here: **could we implement a system that allows for alternate or previously used implementations of snippets to be listed in the website?** That sounds like an interesting feature and something that can be useful if you want to learn JS in depth or not have all the features available to you right away.

skatcat31:
Addressing using Fill and Map:

It adds a syntactical way to explain how it's working on each element which greatly increases readability. however between map and fill, we're REALLY hitting the performance hard. This could probably be done much better by ForEach to address the performance(in testing, when I got to 2D Arrays of size 1000 x 1000 I started noticing a performance difference that only got more pronounced as the arrays got bigger: https://jsperf.com/foreach-fill-versus-map

While it may not look like it here, there is a significant difference in the time taken. However anyone using Arrays that large.... needs to step back and analyze why. At that point Map creating a secondary Array is of little concern.

The compared ForEach version

```js
const initialize2DArray = (w, h, val = null) =>
  Array.from({ length: h })
  .forEach((_, index, array) =>
    array[index] = Array.from({ length: w }).fill(val)
  );
```

The proposed version:

While more streamlined it reads as "experienced only" which is a big strike to it due to the auto magic applied to an overloaded function. It just doesn't read syntacically the more I read it, and after having taken some time and coming back I found myself looking up the `Array.from` documentation again... Which is a bad sign.

Chalarangelo:
`forEach` and `map` are virtually identical in the perf you posted, @skatcat31. At least they are for me. I'm not against the `forEach` modification, but `from`'s callbacks are quite hard to understand. After asking a few colleagues, they all sort of agreed that `from`'s callback magic is a bit weird and that they rarely ever use it, due to not remembering that it exists or that it works that way. So, we should most likely be looking for alternatives other than that.

skatcat31:
Then we should leave it as map since it has the most concise syntax and is semantically easy to understand

PR #746, opened 2018-09-20
Refactor takeRightWhile

Siarhei-Zharnasek:
Refactor takeRightWhile [ENHANCEMENT]

I think it's good place to introduce `reduceRight` in action

- [ ] Website
- [X] Snippets
- [ ] Tests

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] I have updated the documentation accordingly.


skatcat31:
We should look into the performance of constantly iterating over a growing array or if the engines are smart enough now to unroll that automagically. Otherwise the performance can be quite poor.

I remember not pointing out reduce right due to the logical knot it creates in peoples heads. I've seen senior developers get it wrong in production environments several times and that code make it through testing. I'm not a big fan of right reductions like `compose` or `takeRightWhile` if they aren't very verbose or used for mathematical equations because it causes people to think in weird ways.

It has however been a part of the standard for a long time, and while esoteric, is often re-implemented in other libraries, and does have it's uses.

Due to my personal feelings on the confusing nature of NOT seeing the loop and how often I see it wrong in production I will be deferring approval of this to other developers again.

skatcat31:
As a case in point we have missed that our version does not work as intended. Maybe this is a good argument for a reduceRight or rewrite as a whole(with this binding and a proper footprint)

snippets/takeRightWhile.md
```diff
@@ -2,15 +2,10 @@
 
 Removes elements from the end of an array until the passed function returns `true`. Returns the removed elements.
 
-Loop through the array, using a `for...of` loop over `Array.keys()` until the returned value from the function is `true`.
-Return the removed elements, using `Array.reverse()` and `Array.slice()`.
+Loop through the array, using a `Array.prototype.reduceRight` and accumulate element if function returns falsy value.
```
Chalarangelo: `Array.reduceRight`, we do not add the `prototype` part for now.

PR #751, opened 2018-09-25
 [ENHANCEMENT] Simpler solution to mask using padStart

Theomg:
## Description
As asked in https://github.com/30-seconds/30-seconds-of-interviews/pull/107#issuecomment-424019789, I'm making a pull request to the same changes merged in 30 seconds of interviews.

Using `String.padStart` we can make the `mask.md` snippet a bit simpler.

- [ ] Website
- [x] Snippets
- [x] Tests

## Types of changes

## Checklist:

### DIFF
--- a/snippets/mask.md
+++ b/snippets/mask.md
@@ -7,8 +7,7 @@ Omit the second argument, `num`, to keep a default of `4` characters unmasked. I
 Omit the third argument, `mask`, to use a default character of `'*'` for the mask.
 
 ```js
-const mask = (cc, num = 4, mask = '*') =>
-  ('' + cc).slice(-num).padStart(('' + cc).length, mask);
+const mask = (cc, num = 4, mask = '*') => `${cc}`.slice(-num).padStart(`${cc}`.length, mask);
 ```
 
 ```js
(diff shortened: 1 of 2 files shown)


skatcat31:
This is a comment on the snippet itself. String slice is a pure function. Why are we copying it?

skatcat31:
`'' + cc` is used twice in the snippet. We could just use `cc.slice()`

Chalarangelo:
`'' + cc` is used to cast to a string if the value provided is not a string, I think. Am I mistaken?

skatcat31:
Fair point, but there has to be a better more readable way to cast it once and use it once

Chalarangelo:
`${cc}` maybe?

Theomg:
@skatcat31 

I did think it was a bit weird that cast was done twice but decided to leave it alone.
Would you prefer if I cast it once in a const and reused such as:
```js
const mask = (cc, num = 4, mask = '*') => {
  const str = c.toString();
  return str.slice(-num).padStart(str.length, mask);
}
```

Chalarangelo:
@Theomg while this has a small performance increase, it makes the snippet longer and arguably uglier. I prefer just casting twice via template but that's my personal opinion. What do you guys think? @atomiks @fejes713 @flxwu @skatcat31

flxwu:
I'd prefer the ``${cc}``, if that works properly.

PR #761, opened 2018-09-30
:bug: Fix deepClone's bug when values includes null

tangweikun:


skatcat31:
Granted coercing an instance to a falsey value requires some deep knowledge of JS... but it could be done if someone were idiotic enough. We should ideally check
```js
typeof obj[key] === 'object' && obj[key] !== null
```

PR #762, opened 2018-10-01
[FIX] Packager script: Babel 7, named exports

atomiks:
## Description
The packager script was broken in several ways:

- `...` object rest spread operator in `nest.md` (now part of ES2018) was breaking Babel 6 without a transform plugin. Babel 7 works
- Exports were not named and could not be imported like: `import { snippet } from '30-seconds-of-code'`
- `dist` directory should be ignored in the repo and only published to npm

## PR Type
- [ ] Snippets, Tests & Tags (new snippets, updated snippets, re-tagging of snippets, added/updated tests)
- [x] Scripts & Website & Meta (anything related to files in the [scripts folder](https://github.com/30-seconds/30-seconds-of-code/tree/master/scripts), how the repository's automated procedures work and the website)

## Guidelines


skatcat31:
Does the ignore and delete fix the problem with the modules? Since the file has been tracked once, if it's added again won't it track?

skatcat31:
@Chalarangelo dist being in the repo is actually considered bad practice since dist only needs to go to distribution channels, which Github is not.

For that matter however if you do want to upload dist to the repo it would be good to note it since it's not common, and bloats the NPM download and project sizes

Chalarangelo:
@skatcat31 The `npm` download is essentially `dist` itself, right? How does it bloat NPM? Am I missing something there?

I am not against removing it or anything, I just thought it's not a problem and is useful to check if builds are ok in Travis. However, we could set up Travis to create the folder, make the builds and just notify us for errors.

skatcat31:
@Chalarangelo often you add a post install step to build the dist folder to prevent duplicate byte download

skatcat31:
it's why `node-gyp` is so popular

Chalarangelo:
Aaaaand I need to look this whole thing up, my publishing skills are not that good yet. 😝

mariastervic:
If this is still up for discussion, I think that the dist folder is useful asset to have in the repo. I have been downloading the compiled modules sometimes when I did not have npm on a computer I was working on, so I would like the folder to stay.

skatcat31:
@mariastervic that is a fair point worth considering

Chalarangelo:
Let's keep `dist` until we come up with a more robust plan to distribute the code from that directory? We shouldn't make hasty decisions. 

Apart from the `dist` controversy, the PR seems like a very welcome change, so let's change `.gitignore` and merge?

atomiks:
@mariastervic the files in that folder had not been updated for a very long time though 😟 

You don't need the `dist` folder because of the `unpkg` CDN which is a mirror for npm (as long as the releases get updated somewhat often)

https://unpkg.com/30-seconds-of-code@1.1.0/dist/

`.npmignore` should be updated to ignore everything except `package.json` & `dist`.

Chalarangelo:
The issue here is not that we have a `dist` folder, but rather that the files in it do not get pushed from the Travis CRON jobs. The `dist` folder was never meant to be the synced version with the latest release, but rather a nightly build that should sync up with the latest changes in the repo.

skatcat31:
I keep forgetting unpkg is a place...

I think a good decision to be made then is two fold:

- What is the purpose of the dist folder?
- How should we build and deploy the dist folder to users?

I think the first one is pretty obvious:

The dist folder is used to however the actual files for use that are compiled by the build script.

The second id a little more nuanced though because there are two trains of thought:

- Allow Travis nightlies to push to repo the folder(allows for possibly unstable problems)
- Travis builds(releases to NPM) and NPM and Un-Package are the CDNs where people should get the dist folder from

Both have merits, but maintainability of the repo should probably guide the teams decision.

skatcat31:
With that in mind, my thinking towards removing dist:

Pros:

- Smaller repo
- Faster Git downloads and syncs


Cons:

- Can only get usable snippets from copy paste, NPM, or Un-Package
- No nightlies unless you build it yourself

And my thinking to keeping dist:

Pros:

- Nightly builds
- Git can no be used in an emergency to fetch a copy while other sources are unavailable(somewhat unrealistic, but a backup is a backup)

Cons:

- Git syncing and repository downloads will take longer
- History must now track the dist folder contents aside from just for delivery

Chalarangelo:
Great breakdown of the issue at hand, @skatcat31! 

My two cents are:

* Our current Travis build flow doesn't properly build and push the `dist` folder, which is quite problematic, however I feel like we should rewrite the `packager.js` script to do so, if we go with nightly builds. 
* The `dist` folder is not extremely large, but merge conflicts can arise from it, which has never been an issue so far, so maybe this is not the main concern there.
* Cloning the repo with or without the `dist` folder doesn't seem like a concerning factor either - `dist` is currently 257KB and I doubt it will ever go beyond a couple of MBs in size, so it won't slow down clones all that much.
* `dist` could theoretically be used for the testing, instead of individual `.js` files in the test directory, now that we have named exports. This will also test packaging problems (e.g. a snippet isn't properly loaded in from the module) and we could theoretically check against all the different pacakge builds, so we could use this to our advantage.

### My opinion

Based on all previous discussion, I think we can utilize `dist` for testing the snippets and offload `tdd.js` a little bit, so we don't have to build each snippet file individually. Removing a few hundred files from the test directory and actually moving all the `.test.js` files to the main `test` folder will speed up searching for people who write and update tests (I have found myself scrolling a lot in there on VSCode) and will also guarantee that our builds work as intended, while repository size will be similar to what it is now if not lower (as we will essentially move all the exported code to the `dist` folder and remove a ton of unnecessary folders). Providing nightly builds in `dist` could help us indentify issues earlier and it could also be useful to some people that might still use those files. That being said, I vote towards this approach:

* Fix Travis to push nightly builds to `dist`.
* Remove individual snippets from `test`, use `dist` nightlies for importing, combine `tester` and `packager` into one procedure (separate files still) to make tests work together with nightly builds.
* Move all `.test.js` files to top-level `test` directory to allow for easier search and access.

Thoughts?

Chalarangelo:
If the rest of the team is onboard with the testable code copy/move to `dist`, this could end up being a reasonably big change in the build pipeline that could benefit us a lot in the long run. 

Btw, if we do this, we should finally set up test coverage and remove the ugly testlog. I think tools like Codacy support this and it could help us check everything easily enough. 

**Note:** If we do move and update the test system alongside this, we should just rename the PR and work on top of the work @atomiks has already commited. It's a good place to start and it's a perfectly valid PR already, so let's just improve everything that is related and merge the whole bunch at the same time.

skatcat31:
The idea is solid, I would just like to voice descent of committing a lot at once. Sometimes small things fall through the cracks. It shouldn't be a problem, but I'm paranoid and you should get off mah lawn(read: I'm old)

I'd probably just revert the changes to build scripts and put those in another branch so we can work on that large, and potentially complex change, without having to worry about other things. It is after all seen as good practice to isolate procedures and changes to large systems in Git so you can surgically go back and reverse any mistakes without destroying anything else along the way.

skatcat31:
It gives us, if nothing else, a clean starting point to make the next large group of changes to the build script which means these changes could be merged well before those changes, and safely.

dist/_30s.min.js
```diff
@@ -1 +0,0 @@
...ew Set([...c,...a.filter((a)=>-1===c.findIndex((c)=>b(a,c)))])),uniqueElements:(a)=>[...new Set(a)],untildify:(a)=>a.replace(/^~($|\/|\\)/,`${'undefined'!=typeof require&&require('os').homedir()}$1`),unzip:(a)=>a.reduce((a,b)=>(b.forEach((b,c)=>a[c].push(b)),a),Array.from({length:g(...a.map((a)=>a.length))}).map(()=>[])),unzipWith:(a,b)=>a.reduce((a,b)=>(b.forEach((b,c)=>a[c].push(b)),a),Array.from({length:g(...a.map((a)=>a.length))}).map(()=>[])).map((a)=>b(...a)),validateNumber:(a)=>!isNaN(parseFloat(a))&&isFinite(a)&&+a==a,without:(a,...b)=>a.filter((a)=>!b.includes(a)),words:(a,b=/[^a-zA-Z-]+/)=>a.split(b).filter(Boolean),xProd:(c,a)=>c.reduce((b,c)=>b.concat(a.map((a)=>[c,a])),[]),yesNo:(a,b=!1)=>!!/^(y|yes)$/i.test(a)||!/^(n|no)$/i.test(a)&&b,zip:(...a)=>{const b=g(...a.map((a)=>a.length));return Array.from({length:b}).map((b,c)=>Array.from({length:a.length},(b,d)=>a[d][c]))},zipObject:(a,b)=>a.reduce((a,c,d)=>(a[c]=b[d],a),{}),zipWith:(...a)=>{const b=a.length;let c=1<b?a[b-1]:void 0;c='function'==typeof c?(a.pop(),c):void 0;const d=g(...a.map((a)=>a.length)),e=Array.from({length:d}).map((b,c)=>Array.from({length:a.length},(b,d)=>a[d][c]));return c?e.map((a)=>c(...a)):e}}});
```
Chalarangelo: Why did we remove this in the first place?

PR #764, opened 2018-10-01
[FEATURE] add celsiusToFahrenheit and fahrenheitToCelsius

evyatarmeged:
## Description
Adding two new utility snippets - Fahrenheit to Celsius converter and vice versa

## PR Type

## Guidelines

### DIFF
--- a/snippets/celsiusToFahrenheit.md
+++ b/snippets/celsiusToFahrenheit.md
@@ -1,0 +1,12 @@
+### celsiusToFahrenheit
+
+Celsius to Fahrenheit temperature conversion.
+
+Follows the conversion formula `F =  1.8C + 32`
+```js
+const celsiusToFahrenheit = degrees => 1.8 * degrees + 32;
+```
+
+```js
+celsiusToFahrenheit(33) // '91.4'
+```
--- a/snippets/fahrenheitToCelsius.md
+++ b/snippets/fahrenheitToCelsius.md
@@ -1,0 +1,12 @@
+### fahrenheitToCelsius
+
+Fahrenheit to Celsius temperature conversion.
+
+Follows the conversion formula `C = (F - 32) * 5/9`
+```js
+const fahrenheitToCelsius = degrees => (degrees - 32) * 5/9;
+```
+
+```js
+fahrenheitToCelsius(32); // '0'
+```
--- a/snippets_archive/celsiusToFahrenheit.md
+++ b/snippets_archive/celsiusToFahrenheit.md
@@ -2,11 +2,12 @@
 
 Celsius to Fahrenheit temperature conversion.
 
-Follows the conversion formula `F =  1.8C + 32`
+Follows the conversion formula `F =  1.8C + 32`.
+
 ```js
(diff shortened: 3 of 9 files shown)


skatcat31:
I'm not to sold on this snippet due to the limited scope of application. If you wanted conversions in JS you'd more than likely use a full library so you have other convenient conversions available.

evyatarmeged:
@skatcat31 fair enough, so I guess that a `decimalToBase` conversion snippet also matches the same predicament ?

skatcat31:
These snippets have not yet met their fate. I have merely offered my descent. If they had met their fate yet I would have close. What I may not find useful, a million people might. We'll await more feedback from another core member before closing or proceeding with deliberation.

As for a `decimalToBase`(possibly `numberToBase`?), if it handled enough widely used cases it could be used.

As an example, if this temperature conversion instead asked a unit(second argument) or detected one(change footprint to `String -> String`), and did conversion between the four commonly accepted scales(F, C, K, R) then this snippet would be a fantastic addition even with conversion libraries.

However I'm not certain how good of an idea that would be because at that point you would need four outputs, and possibly an intermediary step to conform for output, so 4 inputs and 4 outputs in 16 cases. I mention it as all types applicable because user input could mean converting from Rankin to Rankin, so you'd want to accept that case as well. Accommodating for those cases would lead to a snippet that might just be too large sadly.

This might make good content for an article or archive though should people want to see such a snippet or examples.

Chalarangelo:
I think these make a decent addition. While limited in scope, this is a reasonably useful conversion. I mean we have rads to degrees and vice versa, I don't see any reason not to have these in the main snippet list as utilities or maybe math snippets?

**Note:** @skatcat31 a lot of the `math` category snippets are esoteric, very scientific or limited in scope. This doesn't mean that they are not valuable to a reasonably large group of people - not the scientists, us average coders who never remember those formulas. More of a cheat sheet, than actually elegant snippets, but still useful in my opinion.

That being said, I think we need another opinion before moving forward. @fejes713 @flxwu @atomiks care to share a thought?

Chalarangelo:
Even if we do not merge in main list, this could still reside in the archive. Waiting on the rest of the team to vote on this...

As far as decimal to base goes, JS has `toString()` which accepts a `radix` (2,8,16), so you can probably just do:
```js
var num = 1000;
num.toString(2); // '1111101000'
num.toString(16); // '3E8'
```

skatcat31:
It seems my reply to @Chalarangelo never pushed up. Yes a lot of our math snippets are scientific. That is a fair point. Like I said though, I'm not denying this snippet if people want it. I guess I'm just getting old(er) and crotchety(er)

Chalarangelo:
@skatcat31 Don't get me wrong, I will probably never use these snippets either, as my field rarely includes such conversions. But I know a lot of people who would. Let's wait for a couple more opinions before deciding if we want to archive or merge into the main list. So far we have no real hard argument for one or the other, so let's give the PR a little while to breathe.

flxwu:
I also think this is a quite rare edge-case - +1 on moving into archive.

fejes713:
Totally agree with you @flxwu. It's time to move this into the archive.

PR #771, opened 2018-10-02
adding sorting algo(Bubblesort) :zap:

enkaypeter:
## PR Type

## Guidelines


skatcat31:
I see a package lock(which I'll have to confirm what our deployment rules for are), and a testlog(which should not be committed with your work as it's more for build purposes).

I'm also noticing you haven't actually submitted the snippet just yet. It's sometimes helpful to see it in progress while it's developed so you can go back over the code, but it's also sometimes helpful to just upload the MVP in a squash/recommit eventually, so it's really no problem, but it's odd that there would be submission before snippet.

Can you please remove the test log from your submission?

enkaypeter:
Yeah sure. i'll remove the testlogs. And the snippets? my bad :man_facepalming: i didn't add it. I'll do all of those in the next commit. Thanks

enkaypeter:
Done :white_check_mark:

skatcat31:
you'd actually have to remove those changes from history. This can be done a number of ways, but the easiest would be to revert those two changes so there is no change in your branch.

enkaypeter:
That should do yeah?

package-lock.json
```diff
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
```
Chalarangelo: Please unstage all changes to `package.json`.

snippets/bubbleSort.md
```diff
@@ -0,0 +1,37 @@
+###bubbleSort
```
Chalarangelo: `bubbleSort` is more of an algorithmic implementation than an everyday-use snippet, as there is already a native function built into JS. This snippet can be merged, but will have to be moved to the archive.

tag_database
```diff
@@ -17,6 +17,7 @@ bindKey:function,object,intermediate
 binomialCoefficient:math,intermediate
 bottomVisible:browser,intermediate
 btoa:node,string,utility,beginner
+bubbleSort:uncategorized
```
Chalarangelo: As the snippet will be moved to archive, tags are probably not needed, so the scripts will automatically handle these. Nothing to worry about in this file.

snippets/bubbleSort.md
enkaypeter: Alright

tag_database
enkaypeter: Ok.

enkaypeter:
> @enkaypeter https://stackoverflow.com/questions/1817766/how-to-revert-to-origins-master-branchs-version-of-file
> 
> WARNING: be careful and read closely. This does have an example about resetting wholly to master.

Hey, did what was stated in the stackoverflow thread.

enkaypeter:
May be i'm not getting what need to be done. The `testlog` file no longer on my branch or the `master` branch. It has been checked out. Could you help point me in the right direction?

skatcat31:
Have you committed and pushed the checked out change so the changes are no longer in the branch?

enkaypeter:
Yeah i have

skatcat31:
It looks like you have not committed and pushed checking out the testlog file yet. Doing that should solve the problem

skatcat31:
Wait, for clarification what do you mean is no longer on the `master` branch?

enkaypeter:
Yeah, on the Master branch and on My branch

skatcat31:
I'm going to close the PR due to the time spent attempting to fix conflicts. We'd welcome the changes in a new branch that does not attempt to commit the built files, and just brings in the changes. Thank you for the understanding, but at this point it's hard to remotely debug, and would be faster to copy the changes to a new branch and to abandon the confusing history to the ether.

skatcat31:
basically 30 seconds of copy and pasting files, versus possibly trying to fix history for hours. We'd suggest starting with a clean copy of the repository as well.

With that said don't let this deter you. This is something I've done a lot with developers(both new and old hats) who are learning Git or are even very experienced. Often times it's better to start over and copy the files then it is to try and fix the history using the tools and getting confused, leading to a much more dangerous situation.

PR #806, opened 2018-10-05
[FEATURE] Add getImages snippet

GravitySalad:
## Description

Adding a snippet to fetch all images from within a desired element.

## PR Type

## Guidelines

### DIFF
--- a/snippets/getImages.md
+++ b/snippets/getImages.md
@@ -1,0 +1,24 @@
+### getImages
+
+Fetches all images from within an element and puts them into an array
+
+Use `Element.getElementsByTagName()` to fetch all `<img>` elements inside `Element`, `Element.getAttribute()` to get the `src` attribute from `<img>`, `Array.includes()` for checking if `src` attribute already exists in `Array`, finally `Array.push()` to add the `src` attribute in the returned `Array`
+
+```js
+const getImages = (elem) => {
+    const pageImages = elem.getElementsByTagName('img');
+    let imageList = [];
+
+    for (i=0, l=pageImages.length; i < l; i++) {
+        const imgSrc = pageImages[i].getAttribute('src');
+
+        if (!imageList.includes(imgSrc)) { imageList.push(imgSrc); }
+    }
+
+    return imageList;
+};
+```
+
+```js
+getImages(document); // ['image1.jpg', 'image2.png', '...']
+```
--- a/tag_database
+++ b/tag_database
@@ -96,6 +96,7 @@ geometricProgression:math,intermediate
 get:object,intermediate
 getColonTimeFromDate:date,intermediate
 getDaysDiffBetweenDates:date,intermediate
+getImages:browser,beginner
 getMeridiemSuffixOfInteger:date,beginner
 getScrollPosition:browser,intermediate
 getStyle:browser,css,beginner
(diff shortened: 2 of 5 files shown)


snippets/getImages.md
```diff
@@ -6,8 +6,7 @@ Use `Element.getElementsByTagName()` to fetch all `<img>` elements inside `Eleme
 
 ```js
 const getImages = (elem, duplicates) => {
-    const imgElements = [...elem.getElementsByTagName("img")];
-    const images = imgElements.map(img => img.getAttribute("src"));
+    const images = [...elem.getElementsByTagName("img")].map(img => img.getAttribute("src"));
 
     return duplicates ? images : [...(new Set(images))];
```
Chalarangelo: This can still be simplified to an one-liner, like so:

```js
const getImages = (elem, duplicates = false) => 
  duplicates ? [...elem.getElementsByTagName("img")].map(img => img.getAttribute("src")) : [...(new Set([...elem.getElementsByTagName("img")].map(img => img.getAttribute("src"))))];
```

Note that you have to run the linter on it and you can maybe make it a bit more readable, I just wrote this as a proof of concept.

atomiks: Using a variable is better because it avoids the repeated code. Just remove the whitespace between it and the return value, and it's good to go 🎉

Chalarangelo: @atomiks Agreed, my code looks slightly ridiculous tbh.

GravitySalad: For what it's worth I tried fitting it onto one line but I personally couldn't find a way to remove the duplicate code or make it look any better without a variable.

snippets/getImages.md
```diff
@@ -0,0 +1,17 @@
+### getImages
+
+Fetches all images from within an element and puts them into an array
+
+Use `Element.prototype.getElementsByTagName()` to fetch all `<img>` elements inside `Element`, `Array.prototype.map()` to map every `src` attribute of their respective `<img>` element. `New Set()` creates a `Set` of unique images which is then converted into an `Array` with the `...` operator.
```
flxwu: It's just `Element.getElementsByTagName()` as `Element` **does not inherit** from `Element.prototype`. See the Array prototype [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/prototype) in comparison to the Element [docs](https://developer.mozilla.org/en-US/docs/Web/API/Element).

flxwu: It's just `Element.getElementsByTagName()` as `Element` **does not inherit** from `Element.prototype`. See the Array prototype [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/prototype) in comparison to the Element [docs](https://developer.mozilla.org/en-US/docs/Web/API/Element).

flxwu: I'd also write "to fetch all ... elements inside *the provided element*" as `Element` is just a base class.

flxwu: And you don't have to mention how you declare the Set as we all know how to write `New Set()` IMHO 🙂 so just "Then, create a `Set` to eliminate duplicates and return the array from it".

GravitySalad:
Shortened the description again and squashed the last two commits into one for clarity.

snippets/getImages.md
flxwu: okay, after researching for a while, I think you're right. My fault, sorry!

flxwu:
I think this is ready for merge now? @atomiks

PR #812, opened 2018-10-06
[FEATURE] Added to title case snippet

domgoodwin:
## Description
Added snippet for to title case done via regex and then maps.

## PR Type
- [ ] Snippets, Tests & Tags (new snippets, updated snippets, re-tagging of snippets, added/updated tests)

## Guidelines
- [ Yes] I have read the guidelines in the [CONTRIBUTING](https://github.com/30-seconds/30-seconds-of-code/blob/master/CONTRIBUTING.md) document.

### DIFF
--- a/snippets/toTitleCase.md
+++ b/snippets/toTitleCase.md
@@ -1,0 +1,21 @@
+### toTitleCase
+
+Converts a string to title case (every word begins with capital)
+
+Split string on spaces and then sets first character of every word to capital
+
+```js
+const toTitleCase = str =>
+  str &&
+  str
+    .match(/[A-Z]{2,}(?=[A-Z][a-z]+[0-9]*|\b)|[A-Z]?[a-z]+[0-9]*|[A-Z]|[0-9]+/g)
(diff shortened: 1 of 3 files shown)


skatcat31:
https://www.grammar-monster.com/lessons/capital_letters_title_case.htm

I'm not sure this adheres to the definitions of Title Case. Further discussion is needed.

skatcat31:
Since Title Case can be a little confusing, and has an extremely long list of words to not capitalize that is dependent on certain cases I vote to note that it is not a full title case definition and it only capitalizes the first letter of each group that is non space.

A full title case definition would be extremely long and confusing due to the cases and definitions.

skatcat31:
So I've noticed a couple of thing:

- I've been unable to see non matching word strings to the regex, maybe I'm missing a pattern that would case it to fail?
- This destroys any middle capital proper nouns

Chalarangelo:
Closing due to inactivity and controversy. Feel free to comment if you want this reopened.

Issue #775, opened 2018-10-02
[BUG] tomorrow.js is extremely difficult to read, and doesn't pass the tester


itchylol742:
## Bug description 

tomorrow.js is extremely difficult to read, and doesn't pass the test

## Steps to reproduce

in tomorrow.js, the code is extremely difficult to understand
```js
const tomorrow = (long = false) => {
  let t = new Date();
  t.setDate(t.getDate() + 1);
  const ret = `${t.getFullYear()}-${String(t.getMonth() + 1).padStart(2, '0')}-${String(
    t.getDate()
  ).padStart(2, '0')}`;
  return !long ? ret : `${ret}T00:00:00`;
};
```
also, it doesn't pass tomorrow.test.js
```js
const t1 = new Date(new Date().getFullYear(), new Date().getMonth(), new Date().getDate() + 1);
const t2 = new Date(tomorrow());
// (other tests)
test('Returns the correct date', () => {
  expect(t1.getDate()).toBe(t2.getDate());
});
```

## Expected behavior

A more readable code, and passing the test.

## Screenshots

pastebin: https://pastebin.com/raw/vR5uK7qP (from testlog)

## Environment

Node 8.10.0
npm 6.4.1
Visual Studio Code 1.27.2
Windows 8.1

gitjeff05:
Should this just use ``t.toLocaleDateString()`` rather than attempting to format for one specific locale? [The ``toLocaleDateString`` method also accepts a locale](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString#Checking_for_support_for_locales_and_options_arguments):
```
const t = new Date()
console.log(t.toLocaleDateString()) // 10/5/2018
console.log(t.toLocaleDateString('de-DE')) // 5.10.2018
```

skatcat31:
@gitjeff05 We want to hard code it at midnight local. It's to represent the first moment of tomorrow so you can try and use it to figure out when tomorrow starts. This is useful for countdowns and other such things that care about midnight(usually when tomorrow is defined)

Chalarangelo:
Seeing as this thread is kind of dead, please move all further discussion to #849.

PR #849, opened 2018-10-17
[FIX] Improve the readability of tomorrow.js (solves #775)

filipdanic:
As discussed in #775, `tomorrow.js` is hard to read. I’ve updated the implementation to simplify things.

I’ve removed the boolean parameter `long`. It’s a code smell and only adds to the complexity of the function. We should encourage folks to use a different method to add a prefix/suffix to a string. 

## PR Type
- [✅ ] Snippets, Tests & Tags (new snippets, **updated snippets**, re-tagging of snippets, added/updated tests)

## Guidelines
- [✅] I have read the guidelines in the [CONTRIBUTING](https://github.com/30-seconds/30-seconds-of-code/blob/master/CONTRIBUTING.md) document.

### DIFF
--- a/snippets/tomorrow.md
+++ b/snippets/tomorrow.md
@@ -2,20 +2,18 @@
 
 Results in a string representation of tomorrow's date.
 
-Use `new Date()` to get today's date, adding one day using `Date.getDate()` and `Date.setDate()`, and converting the Date object to a string.
+First we use `new Date()` to get today's date, then add one day using `Date.getDate()` and mutate the initial `Date` via `Date.setDate()`. The we use `.toISOString` to get the date part of the string – discarding the time part.
 
 ```js
-const tomorrow = (long = false) => {
+const tomorrow = () => {
   let t = new Date();
   t.setDate(t.getDate() + 1);
-  const ret = `${t.getFullYear()}-${String(t.getMonth() + 1).padStart(2, '0')}-${String(
-    t.getDate()
-  ).padStart(2, '0')}`;
-  return !long ? ret : `${ret}T00:00:00`;
+  return t.toISOString().split('T')[0];
 };
 ```
 
 ```js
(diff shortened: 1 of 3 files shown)


skatcat31:
will `Date#toISOString()` cause any issues with it being UTC versus local like `Date#getFullYear` etc?

filipdanic:
Good point – it probably fails on New Year's Eve.

But I’m thinking that replacing `new Date()` with `new Date(Date.now())` should solve that issue?

Chalarangelo:
I'm putting this on hold for a little bit, until we figure it out. Note that the test structure for the snippet needs to be updated to the new format, but that can be done right before the merge from a maintainer, as well.

Chalarangelo:
Based on [this answer on SO](https://stackoverflow.com/a/23081260/1650200) and after some trial and error, I came up with this solution to the issue ([repl here](https://repl.it/repls/EnchantedInsecureIntegers)):
```js
const tomorrow = () => {
  let date = new Date();
  date.setDate(date.getDate() + 1);
  return date;
}
```

In the same repl, I describe a proposal for a `nextDate()` function and showcase some tests. These could solve the problem one and for all, right?

@skatcat31 Any thoughts?

skatcat31:
Also did we ever decide if we were going to use `Class#prototypeMethod()` notation?

RobertAKARobin:
Out of curiosity, how did this make it into `master` with a failing test?

Chalarangelo:
@RobertAKARobin this is one of our oldest snippets, somehow it worked for a while, then it broke at some point. It's probably one of the most problematic snippets in the entire codebase so far.

skatcat31:
To be fair @RobertAKARobin we've considered voting on removing it. It's a problem and with the recent possible changes to DST in PT it might get even more confusing.

PR #855, opened 2018-10-19
Add toTitleCase snippet and update tag_database accordingly

ccjmne:
## Description

I added a `toTitleCase` snippet to convert any string to its title-case equivalent.

It reuses the same "tokeniser" RegExp that is already featured in the `toCamelCase`, `toKebabCase` and `toSnakeCase` snippets.

## PR Type

## Guidelines
- [x] My PR doesn't include any `testlog` changes.

### DIFF
--- a/snippets/toTitleCase.md
+++ b/snippets/toTitleCase.md
@@ -1,0 +1,22 @@
+### toTitleCase
+
+Converts a string to title case.
+
+The algorithm featured here breaks down as follows:
+1. Break the string into words, using a [Regular Expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions)
+2. Capitalize each of them
+3. Stich them back together, using the whitespace `' '` character as a separator
+
+```js
+const toTitleCase = str => str
+  .match(/[A-Z]{2,}(?=[A-Z][a-z]+[0-9]*|\b)|[A-Z]?[a-z]+[0-9]*|[A-Z]|[0-9]+/g)
+  .map(x => x.charAt(0).toUpperCase() + x.slice(1).toLowerCase())
+  .join(' ');
+```
+
+```js
+toTitleCase('some_database_field_name'); // 'Some Database Field Name'
+toTitleCase('Some label that needs to be title-cased'); // 'Some Label That Needs To Be Title Cased'
+toTitleCase('some-package-name'); // 'Some Package Name'
+toTitleCase('some-mixed_string with spaces_underscores-and-hyphens'); // 'Some Mixed String With Spaces Underscores And Hyphens'
+```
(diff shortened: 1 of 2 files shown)


Chalarangelo:
thanks for contributing!

This is still missing tests, but that's alright, I'll add them in after the merge if you don't have time for that!

Siarhei-Zharnasek:
@ccjmne 
`x.charAt(0).toUpperCase() + x.slice(1).toLowerCase()` why didn't you use template string to concatenate strings?
also `x.charAt(0)` it's possible to rewrite like `x[0]`

Chalarangelo:
`x.charAt(0)` is slightly more descriptive to me. The other suggestion sounds like a good idea, though, in general.

ccjmne:
@Siarhei-Zharnasek Thanks for the review!

1. I deliberately chose to not use the template string to concatenate these two bits together, because I think it adds more noise than anything else.
If there were other pieces of strings weaved in between, it'd make much more sense:

```javascript
`${ x.charAt(0).toUpperCase() }${ x.slice(1).toLowerCase() }` // meh
x.charAt(0).toUpperCase() + x.slice(1).toLowerCase()          // looks alright
// vs:
'<strong>' + x.key + '</strong>: <code>' + x.value + '</code>' // meh
`<strong>${ x.key }</strong>: <code>${ x.value }</code>`       // there we go!
```

2. While I agree with @Chalarangelo about `string.charAt(0)`'s intent being clearer than that of `string[0]`, I went for this to handle the empty string `''`:
```javascript
''[0]        // undefined
''.charAt(0) // ''
```

The problem with `undefined` is that, when transformed back into a string context (by concatenating, for example `undefined + ''`), actually evaluates to `'undefined'`.
In this case, it wouldn't matter though, because `x` at this point can't be an empty string.

All that said, I'm happy to change these if you think it'd be better 🙂

snippets/toTitleCase.md
```diff
@@ -0,0 +1,19 @@
+### toTitleCase
+
+Converts a string to title case.
+
+Break the string into words, using a regexp, and combine them capitalizing the first letter of each word and adding a whitespace between them.
+
+```js
+const toTitleCase = str => str
+  .match(/[A-Z]{2,}(?=[A-Z][a-z]+[0-9]*|\b)|[A-Z]?[a-z]+[0-9]*|[A-Z]|[0-9]+/g)
+  .map(x => x.charAt(0).toUpperCase() + x.slice(1).toLowerCase())
```
skatcat31: Causes problems with things like 'mcDolans' becoming 'Mcdolans'.

Ideally since intermediate capitals should be preserved in title case, each word should just have the first letter capitalized

skatcat31:
```suggestion
  .map(x => x.charAt(0).toUpperCase() + x.slice(1))
```

ccjmne: Ha, very good point! Thanks!

PR #889, opened 2018-12-07
[FEATURE] chainAsync update: last function shouldn't get 'next' as argument

RobertAKARobin:
## Description
chainAsync update: last function shouldn't get 'next' as argument. That way it's easier to add a condition within the functions to check whether they're the last one, e.g. `if(next){/* is last function*/}else{/* isn't */}`. This is useful when the array of functions is generated, rather than hard-coded.

My tests pass, but the `npm` scripts outlined in the docs didn't update `test/_30.js`. Not sure what additional step is needed for that.

## PR Type

## Guidelines
- [x] My PR doesn't include any `testlog` changes.

### DIFF
--- a/snippets/chainAsync.md
+++ b/snippets/chainAsync.md
@@ -10,7 +10,7 @@ const chainAsync = fns => {
   const last = fns[fns.length - 1];
   const next = () => {
     const fn = fns[curr++]
-    fn === last ? fn(undefined) : fn(next);
+    fn === last ? fn() : fn(next);
   }
   next();
 };
--- a/snippets/chainAsync.md
+++ b/snippets/chainAsync.md
@@ -6,10 +6,11 @@ Loop through an array of functions containing asynchronous events, calling `next
 
 ```js
 const chainAsync = fns => {
-  let curr = 0, last = fns[fns.length - 1];
+  let curr = 0;
+  const last = fns[fns.length - 1];
   const next = () => {
     const fn = fns[curr++]
-    fn(fn === last ? null : next)
+    fn === last ? fn(undefined) : fn(next);
   }
   next();
 };
--- a/snippets/chainAsync.md
+++ b/snippets/chainAsync.md
@@ -6,8 +6,11 @@ Loop through an array of functions containing asynchronous events, calling `next
 
 ```js
 const chainAsync = fns => {
-  let curr = 0;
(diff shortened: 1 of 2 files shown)


snippets/chainAsync.md
```diff
@@ -6,8 +6,11 @@ Loop through an array of functions containing asynchronous events, calling `next
 
 ```js
 const chainAsync = fns => {
-  let curr = 0;
-  const next = () => fns[curr++](next);
+  let curr = 0, last = fns[fns.length - 1];
```
skatcat31: last should be a const, it is never overwritten
```suggestion
  let curr = 0;
  const last = fns[fns.length - 1];
```

snippets/chainAsync.md
```diff
@@ -6,8 +6,11 @@ Loop through an array of functions containing asynchronous events, calling `next
...
+  const next = () => {
+    const fn = fns[curr++]
+    fn(fn === last ? null : next)
```
skatcat31: This doesn't work like you'd think it does. It passed a null to the function. You should pass an undefined(for proper undefined checks) or modify this line to either call then function with next, or just call the function.

```suggestion
    fn === last ? fn() : fn(next);
```

Note on passing `undefined` versus a blank call(`()`):

I'd have to dig into the documentation to see if this would cause any issues with the `arguments` magic pointer(length of 0 versus length of k)

skatcat31: although that'd be a pretty easy check, I'm currently swamped with papers

RobertAKARobin:
```
function foo(){
  console.log(arguments.length)
}
foo()
// 0
foo(undefined)
// 1
```

RobertAKARobin: To keep the argument length consistent, I'll go with `undefined`.

skatcat31: actually to keep it consistent, you should go with no arguments since that means otherwise you are passing something to it, so any code that relies on `arguments.length === 0` will fail for not receiving props on any possible overloading code that accepts differing lengths of arguments and checks, and even worse this could cause problems with auto curried closures that rely on the arguments length as well. You should call it without arguments

RobertAKARobin: Haha, fair enough.

PR #900, opened 2018-12-23
added new 30 sec

Bhupesh-V:
### Added 30 seconds of STL[Standard Template Library] - unofficial


Bhupesh-V:
So how can i add my 30 seconds of STL repo(unofficial)

PR #902, opened 2018-12-30
[ENHANCEMENT] Replace double-quote characters in CSV fields

m3g4p0p:
Hi everyone! :-)

## Description
As CSV fields are wrapped in double quotes by default, double-quote characters should also be replaced with pairs of quotes which is more or less the standard way to handle these. I was also thinking about an optional escape character argument, but didn't want to over-engineer it at this point.

## PR Type

## Guidelines

### DIFF
--- a/snippets/arrayToCSV.md
+++ b/snippets/arrayToCSV.md
@@ -3,15 +3,17 @@
 Converts a 2D array to a comma-separated values (CSV) string.
 
 Use `Array.prototype.map()` and `Array.prototype.join(delimiter)` to combine individual 1D arrays (rows) into strings.
+Use `String.prototype.replace()` to replace double-quote characters with pairs of quotes.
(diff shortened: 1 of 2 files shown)


skatcat31:
This is not strictly true as they are just escaped by some mechanism. The problem is this is ENTIRELY implementation dependent, hence why Open Office and Libra Office when importing will try to detect it, but allow you to set quote escaping characters. We should put in place a pattern to allow escaping of quotes how the person would like much like the custom delimiter.

skatcat31:
After checking several implementations I can only figure that this is not acceptable without expanding to also allow for quote escaping. The more recent `\"` is used in many of them as well as them auto supporting `""` or allowing for switching. Because of this I will be closing this PR as it would be entirely implementation dependent.

PR #903, opened 2019-01-04
[FIX]: prettyBytes 1000 to 1024

yugasun:
## Description

Thousand of bytes for programmer should be 1024.

## PR Type

## Guidelines


Chalarangelo:
@atomiks I somehow misread your comment thinking it was the other way round. I've done some research and what you are saying is actually very much correct. 

> The IEC further specified that the kilobyte should only be used to refer to 1000 bytes. One thousand and twenty-four kibibytes (1024 KiB) is equal to one mebibyte (1 MiB), where 1 MiB is 10242 bytes.

Therefore I am going to close this and keep the code as-is. Thanks for your time and effort @yugasun!

PR #938, opened 2019-03-15
Delete links to Gitter

fejes713:
We're closing our Gitter and Slack channels - very few used it, and we can be 100% reliable on issues and PR's


slayah2133:
H

PR #959, opened 2019-04-28
Update createDirIfNotExists function

you-think-you-are-special:
## Description
Modified `createDirIfNotExists` snippet. Now it is asynchronous and safe. Those reasons are attached in the snippet description:

```
Using fs.exists() to check for the existence of a directory before calling fs.mkdir() is not recommended. 
Doing so introduces a race condition, since other processes may change the directory's state between the two calls. 
Instead, user code should create the directory directly and handle the error raised if the directory does not exist.

Why not fs.mkdirSync? In busy processes, the programmer is strongly encouraged to use the asynchronous versions of these calls. 
The synchronous versions will block the entire process until they complete — halting all connections.
```

Also, updated `module.js` script. It didn't want to work with multiple `require`

## PR Type
- [x] Scripts & Website & Meta (anything related to files in the [scripts folder](https://github.com/30-seconds/30-seconds-of-code/tree/master/scripts), how the repository's automated procedures work and the website)

## Guidelines


you-think-you-are-special:
It will break the tests. Are you sure?

Chalarangelo:
Why would a snippet change require a change in a script to work?

you-think-you-are-special:
I use two `require` in snippet.
```js
const fs = require('fs');
const util = require('util');
```
Unfortunately, script can't handle this. The script loses second one. Am I wrong?

Chalarangelo:
You should be using something along the lines of 
```js
const fs = require('fs'), util = require('util');
```

you-think-you-are-special:
I've already tried this. It generates that code and won't work

```
const fs = typeof require !== "undefined" && require('fs');
const crypto = typeof require !== "undefined" && require('crypto');
const fs = typeof require !== "undefined" && require('fs'), util = require('util');
```

Chalarangelo:
I am tagging as `on hold` until I can get back to you about the `require` issue, hopefully later today or tomorrow.

you-think-you-are-special:
Why my pull request was closed?

PR #999, opened 2019-07-19
[FEATURE] Add isWeekday code snippets

peter279k:
## Description
- Add the `isWeekday` code snippets.

## PR Type
- [X] Snippets, Tests & Tags (new snippets, updated snippets, re-tagging of snippets, added/updated tests)

## Guidelines
- [X] I have read the guidelines in the [CONTRIBUTING](https://github.com/30-seconds/30-seconds-of-code/blob/master/CONTRIBUTING.md) document.

### DIFF
--- a/snippets/isWeekday.md
+++ b/snippets/isWeekday.md
@@ -1,0 +1,16 @@
+### isWeekday
+
+Results in a boolean representation of a specific date.
+
+Pass the specific date object firstly.
+Use `Date.getDay()` to check weekday then return a boolean.
+
+```js
+const isWeekday = (t = new Date()) => {
+  return t.getDay() >= 1 && t.getDay() <= 5;
+};
+```
+
+```js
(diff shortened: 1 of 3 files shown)


snippets/isWeekday.md
```diff
@@ -0,0 +1,18 @@
+### isWeekday
+
+Results in a boolean representation of a specific date.
+
+Pass the specific date format string firstly. If not, use the current date by default.
+Use `new Date()` to let date format string be date object.
+Use `Date.getDay()` to check weekday then return a boolean.
+
+```js
+const isWeekday = (date = '') => {
```
Chalarangelo: Wouldn't it be better if the default value was `new Date()` instead of `''`, so that you don't have to assign it below?

peter279k: @Chalarangelo. Yes, you'right.

This would be fine to let the `date` is `Date` object by default.

PR #1001, opened 2019-07-19
[FEATURE] Add isInteger code snippets

peter279k:
## Description
- Add `isInteger` code snippets.

## PR Type
- [X] Snippets, Tests & Tags (new snippets, updated snippets, re-tagging of snippets, added/updated tests)

## Guidelines
- [X] I have read the guidelines in the [CONTRIBUTING](https://github.com/30-seconds/30-seconds-of-code/blob/master/CONTRIBUTING.md) document.

### DIFF
--- a/snippets/isInteger.md
+++ b/snippets/isInteger.md
@@ -1,0 +1,19 @@
+### isInteger
+
+Checks if the given argument is a integer.
+
+Use `typeof` to check if a value is classified as a number primitive.
+To check the number is integer, using the `String.indexOf` to make sure it doesn't contain `.`.
+To safeguard against `NaN`, check if `val === val` (as `NaN` has a `typeof` equal to `number` and is the only value not equal to itself).
+
+```js
+const isInteger = val => typeof val === 'number' && String(val).indexOf('.') === -1 && val === val;
+```
+
+```js
+isNumber(1); // true
+isNumber(1.0); // true
+isNumber(1.1); // false
+isNumber('1'); // false
+isNumber(NaN); // false
+```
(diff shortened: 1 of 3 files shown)


peter279k:
The `isNumber` just checks the value is numeric.

To check the numeric values, I think we should implement the `isInteger` or `isFloat`.

And it can check the numeric values are float or integer numbers in advanced.

Chalarangelo:
We used to have both of those and we dropped them for whatever reason. There were a bunch of issues with the way JS parses numbers and many other things as far as I remember. `isNumber` was the most viable option, anything more complicated than that is bound to require some checks that are not 30-seconds long or intuitive enough. (You can check old discussions about this in some of the older issues)

peter279k:
@Chalarangelo, thank you for your explanation.

The advanced numeric value check will be complicated and it's not part of the `30-seconds-code`.

skatcat31:
notes:
It's probably best to avoid doing any sort of advanced logic on basics since they're still instance. You never actually have access to a "basic" value in JS. `1` is actually an instance of Number with a primitive value flag. In JS you can't touch the Primitives from user space, and thus they can always be coerced. Checking `isInteger` or along those lines is "worthless" because at next tick it could be a float and it will still behave exactly the same mathematically to user space. It is better to instead force the coercion without modifying the pointer value if you need an integer, or to throw if it's not exactly modulo by 1 since for all intents and purposes `1.0 === 1`

```js
1%1 // 0
X%1 === 0 // is an integer
Y%1 !== 0 // is a float
1.0 === 1 // true
```

PR #1023, opened 2019-10-02
For doing AND operations

davisraimon:


davisraimon:
Can i go for XOR ?

Chalarangelo:
@davisraimon Generally, logical operators are not that useful in most real-life scenarios, especially as standalone functions. We decided to avoid creating such functions a while ago, due to the fact that they don't really teach people something very useful that they couldn't find with a Google search or already know coming from another language. Therefore, it's best if we refrain from logical operator snippets for the time being.

PR #1037, opened 2019-10-24
fix(pick of Object):add a judgment 

fnlearner:
add a judgment which whether every curr only exists in the obj itself or not


fnlearner:
emmm,so it is unnecessary for that ?

PR #1146, opened 2020-05-15
adding isMobile

ahmad-ali14:
boolean, Returns true if the client uses mobile, otherwise returns false.

`isMobile()` does not affected by screen size, if your using laptop and your screen width 500px it still returning false. 

`isMobile()` uses regex to check userAgent if it's mobile or not.

### DIFF
--- a/snippets/isMobile.md
+++ b/snippets/isMobile.md
@@ -1,0 +1,23 @@
+---
+title: isMobile
+tags: client agent,intermediate
+---
+
+boolean, Returns true if the client uses mobile, otherwise returns false.
+
+`isMobile()` does not affected by screen size, if your using laptop and your screen width 500px it still returning false. 
+
+'isMobile()' uses regex to check userAgent if it's mobile or not.
+
+```js
+const isMobile = () =>
+  {
+    return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
+    window.navigator.userAgent );
+  }
+  
+```
+
+```js
+isMobile(); // false
+```


skatcat31:
some information regarding the future of user agent:
https://medium.com/@galmeiri/get-ready-for-chrome-user-agent-string-phase-out-c6840da1c31e

I'm not a fan of this snippet because of the fact that it requires a known list of "mobile" clients, only interacts wit h the "declared" user agent string, and is not future compatible, and when attempting to code to a "mobile" client mobile can mean many things that this doesn't address.  This will not be a very adaptable snippet either and would require maintenance as time goes on.

If there were a more addressable way to detect what a mobile client were based on the criteria put forth of what the developer wanted I'd be all for this, but because "mobile" doesn't always mean "engine" and can commonly refer to "screen size" and is in all cases arbitrary I'm not certain in the current form this snippet can really be made more adaptable either.

ahmad-ali14:
@skatcat31 honestly I wasn't aware of deprecating UA. this a new info.

Chalarangelo:
Thanks for contributing. User Agents are extremely unreliable, easy to fake and need to be constantly updated with any new information that comes out anytime there's an update. Apart from that, newer devices blur what is mobile and what is not (think about laptops with touchscreens or mobiles that have keyboards attached via a dock), so it's far less desirable to check for mobile UAs. Instead, checking for individual features, screen sizes etc. is much more useful information for most modern web apps. Based on that info, I am going to go ahead and close this PR. Thanks anyways and feel free to open a new PR if you have any other snippets you want to share.

PR #1149, opened 2020-05-20
Add partitionBy

rodmoioliveira:
This is a javascript implementation of this function: https://clojuredocs.org/clojure.core/partition-by

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,7 @@ public
 
 # Mac files
 .DS_Store
+
+# Vim
+*.vim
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -80,4 +80,8 @@ yarn-error.log
 public
 
 # Mac files
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+# Vim
```
rodmoioliveira, line 85: Support for vim users

PR #1156, opened 2020-06-07
create goobox

ahmad-ali14:
string, Returns a string URL that can be used directly as a source of an image.

### DIFF
--- a/snippets/goobox.md
+++ b/snippets/goobox.md
@@ -17,8 +17,7 @@ cconst goobox = (url)=>{
     let drive_regex =/(http(s)*:\/\/)*(www\.)*(drive.google.com\/file\/d\/)/;
 
     if(url.match(dropbox_regex)){
-        let url2 = url.replace(/(http(s)*:\/\/)*(www\.)*/, "https://dl.");
-       return url2;
+       return url.replace(/(http(s)*:\/\/)*(www\.)*/, "https://dl.");
     }
     else if(url.match(drive_regex)){
         let url2 = url.replace(drive_regex, "");
--- a/snippets/goobox.md
+++ b/snippets/goobox.md
@@ -19,12 +19,8 @@ cconst goobox = (url)=>{
     if(url.match(dropbox_regex)){
        return url.replace(/(http(s)*:\/\/)*(www\.)*/, "https://dl.");
     }
-    else if(url.match(drive_regex)){
-        let url2 = url.replace(drive_regex, "");
-        let url3 = url2.match(/[\w]*\//);
-        let file_id = url3[0].replace(/\//,"")
-        let url4 = " https://drive.google.com/uc?id="+file_id
-       return url4;
+    if(url.match(drive_regex)){
+        return `https://drive.google.com/uc?id=${url.replace(drive_regex, "").match(/[\w]*\//)[0].replace(/\//,"")}`;
     }
     else {
        return console.error('Wrong URL, not a vlid drobox or google drive url');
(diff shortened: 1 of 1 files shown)


snippets/goobox.md
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+```
+
+```js
+let url = 'https://drive.google.com/file/d/1PiCWHIwyQWrn4YxatPZDkB8EfegRIkIV/view' // can't be image src.
+
+goobox(url);
+
+//  https://drive.google.com/uc?id=1PiCWHIwyQWrn4YxatPZDkB8EfegRIkIV  // works as image src
```
skatcat31, line 41:
```suggestion
let driveurl = 'https://drive.google.com/file/d/1PiCWHIwyQWrn4YxatPZDkB8EfegRIkIV/view' // can't be image src.

goobox(driveurl); //  https://drive.google.com/uc?id=1PiCWHIwyQWrn4YxatPZDkB8EfegRIkIV // can be image source
```
since the google drive URL is long leaving it on a separate line is okay, but we should put the comment of the result on the same line as the call. We might also be able to omit the `// can be image source` comment?

skatcat31: Is there the possibility of also getting a dropbox URL example? Also, the images need to be vetted to make sure they're a known quantity and always will be

snippets/goobox.md
```diff
@@ -0,0 +1,42 @@
...
+
+```js
+cconst goobox = (url)=>{
+
+    let dropbox_regex = /(http(s)*:\/\/)*(www\.)*(dropbox.com)/;
+    let drive_regex =/(http(s)*:\/\/)*(www\.)*(drive.google.com\/file\/d\/)/;
+
+    if(url.match(dropbox_regex)){
+        let url2 = url.replace(/(http(s)*:\/\/)*(www\.)*/, "https://dl.");
+       return url2;
```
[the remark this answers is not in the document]
ahmad-ali14: you're right, I just made it easier to read.

snippets/goobox.md
```diff
@@ -0,0 +1,42 @@
...
+    }
+    else if(url.match(drive_regex)){
+        let url2 = url.replace(drive_regex, "");
+        let url3 = url2.match(/[\w]*\//);
+        let file_id = url3[0].replace(/\//,"")
+        let url4 = " https://drive.google.com/uc?id="+file_id
+       return url4;
+    }
```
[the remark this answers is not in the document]
ahmad-ali14: totally right;

snippets/goobox.md
ahmad-ali14: what you mean by vetted ??
I'll give a dropbox example.

skatcat31: vetted means approved, and done in such a way that it can be made sure to never be changed to something we don't want

ahmad-ali14: @skatcat31 I've  done:
- added dropbox example.
- used more appropriate photo in my example witch is 30secs logo.
- check the photo here:  https://dl.dropbox.com/s/degru93hceo26m6/30sec.png?dl=0
- I also fixed the sandbox example

PrinceProjects (commented):
This is good stuff

Chalarangelo (changes requested):
Thanks for contributing. However, the snippet has some guideline violations and is also based on 3rd party APIs and code that we have no access to. We avoid such dependencies to minimize maintenance costs. Closing on that basis.

PR #1165, opened 2020-07-08
use `instanceof` to determine if arg is an array

dym-sh:
as per [discussion](https://twitter.com/30secondsofcode/status/1280479521788047360)

### DIFF
--- a/snippets/last.md
+++ b/snippets/last.md
@@ -8,8 +8,7 @@ Returns the last element in an array.
 Check if `arr` is an instance of an `Array`, use `arr.length - 1` to compute the index of the last element of the given array and return it, otherwise return `undefined`.
 
 ```js
-const last = arr => 
-  (arr instanceof Array ? arr[arr.length - 1] : undefined);
+const last = arr => (arr instanceof Array ? arr[arr.length - 1] : undefined);
 ```
 
 ```js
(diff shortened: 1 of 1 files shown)


dym-sh:
why force onto one single line? and use brackets and semicolon in this case

``` js
const last = arr
  => arr instanceof Array
   ? arr[arr.length - 1]
   : undefined
```
:rofl:

Chalarangelo:
@dym-sh Because it's the style we have decided to use in general, so we're sticking to it for consistency reasons.

snippets/last.md
```diff
@@ -5,10 +5,10 @@ tags: array,beginner
 
 Returns the last element in an array.
 
-Check if `arr` is truthy and has a `length` property, use `arr.length - 1` to compute the index of the last element of the given array and return it, otherwise return `undefined`.
+Check if `arr` is an instance of an `Array`, use `arr.length - 1` to compute the index of the last element of the given array and return it, otherwise return `undefined`.
 
 ```js
-const last = arr => (arr && arr.length ? arr[arr.length - 1] : undefined);
+const last = arr => (arr instanceof Array ? arr[arr.length - 1] : undefined);
```
skatcat31, line 11: This needs to be reverted or changed. This is breaking across iframes.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

`When checking for Array instance, Array.isArray is preferred over instanceof because it works through iframes`

PR #1166, opened 2020-07-12
use isArray instead of instanceof

dym-sh:
according to https://github.com/30-seconds/30-seconds-of-code/pull/1165#pullrequestreview-446823247


skatcat31:
I wonder if we want to go back to the duck casing version so it can be used on array likes like Strings too?

PR #1167, opened 2020-07-17
reverse number snippet added

AtharvaThorve:
This is a code snippet to reverse a number in javascript.

### DIFF
--- a/snippets/reverseNumber.md
+++ b/snippets/reverseNumber.md
@@ -1,0 +1,22 @@
+---
+title: reverseNumber
+tags: number,array,beginner
+---
+
+Reverses a number.
+Use `Object.prototype.toString()` to convert number to a string.
+Use the split function `String.prototype.split()` to convert string to array and `Array.prototype.reverse()` to reverse the order of the characters in the string.
+Combine characters to get a string using `String.prototype.join('')`.
+Use the `parseInt()` to convert string to a number and use `Math.sign()` to preserve the signum value of the number.
+
+```js
+const reverseNumber = n => {
+  const reveresedNumber = n.toString().split('').reverse();
+  return parseInt(reveresedNumber) * Math.sign(n);
+}
+```
+
+```js
+reverseNumber(981); // 189
+reverseNumber(-500); // -5
+```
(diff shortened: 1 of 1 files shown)


snippets/reverseNumber.md
```diff
@@ -0,0 +1,22 @@
...
+Reverses a number.
+Use `Object.prototype.toString()` to convert number to a string.
+Use the split function `String.prototype.split()` to convert string to array and `Array.prototype.reverse()` to reverse the order of the characters in the string.
+Combine characters to get a string using `String.prototype.join('')`.
+Use the `parseInt()` to convert string to a number and use `Math.sign()` to preserve the signum value of the number.
+
+```js
+const reverseNumber = n => {
+  const reveresedNumber = n.toString().split('').reverse();
+  return parseInt(reveresedNumber) * Math.sign(n);
```
skatcat31, line 15: breaks on non-int numbers

skatcat31: If this should be limited to just ints we should reflect that

AtharvaThorve:
@skatcat31 required changes have been made.

skatcat31:
Then it's onto the final review. I don't know if the note is really needed, but if there is no problem with it someone with write access will probably either ask for a change or merge it.

raman-maker:
i could not understand.

AtharvaThorve:
@raman-maker in the code we first convert the number to a string then using the `split` function we convert the string to an array of digits. Since array has the prebuilt method `reverse` available then we use it to reverse the array. Then using the `join` function we again convert the array to a string with no delimiter in between the elements. In the end, we convert the string again to a number type using the `parseFloat` function and use the `Math.sign` function to maintain the signum value of the number. <br>
This is a bit complicated procedure with a lot of type conversions in between but this way we can use a lot of prebuilt functions and the code gets shorter and concise. <br>
Another way to do is to use division and modulus to reverse a number but that is much more complicated than this approach.
You can read about the second approach [here](https://krazytech.com/programs/an-html-code-to-reverse-a-given-number-using-javascript)

PrinceProjects (approved):
Awesome

Iloveangelnicole2 (approved):

Iloveangelnicole2 (approved):

Chalarangelo (approved):
Thanks for contributing and sorry for being late to the conversation. Made some minor changes, merging right away!

bwillsc (commented):
Great

PR #1173, opened 2020-07-29
Added format number in JS

piyukore06:


snippets/formatNumber.md
```diff
@@ -0,0 +1,32 @@
...
+    if (count % 3 == 0 && i != 0) {
+      formatted = ',' + formatted;
+    }
+  }
+  return formatted;
+}
+```
+
+```js
+formatNumber(123456); // '123,456'
```
saleymi, line 30: 66067304

piyukore06: Do you mean, should I add that to examples?

snippets/formatNumber.md
```diff
@@ -0,0 +1,32 @@
+---
+title: formatNumber
+tags: date,math,string,intermediate
```
Trinityyi, line 3: Tags should be `string,math,beginner`.

snippets/formatNumber.md
```diff
@@ -0,0 +1,32 @@
...
+---
+
+Returns a number in readable format.
```
Trinityyi, line 6:
```suggestion
Returns a number with commas as thousand separators.
```

snippets/formatNumber.md
```diff
@@ -0,0 +1,32 @@
...
+
+Use `Number.prototype.toString()` to convert a number to string.
+Use `for` loop to loop through the characters in string. Write to a new string variable one character at a time, adding a comma at every third character. Return the newly created string.
+
+
+```js
+const formatNumber = number => {
```
Trinityyi, line 13: The implementation should use `Number.prototype.toLocaleString` instead and be a one-liner with an implicit return. Also, the description should be updated.

snippets/formatNumber.md
Trinityyi: The two examples are more than enough to explain the snippet. No need to add any more.

snippets/formatNumber.md
piyukore06: Nice, thank you for introducing me to toLocaleString. That makes it so easy :)

snippets/formatNumber.md
piyukore06: If we are using `toLocaleString`. I think the description could be 

Returns a number with using the local number format order

its more generic. 
Also I will add examples with multiple locale, so that there's no confusion for someone, who doesn't know about it.

PR #1187, opened 2020-08-19
Add new snippet random_vowel

samalien:
Hello,
I really want to be among you, this project suits me too well, it's a first function that can be considered as a test although it is interesting


Chalarangelo (changes requested):
Thanks for contributing. However, this snippet has multiple issues (formatting, syntax, real-world usage etc.). Please read the guidelines thoroughly before contributing and make sure any added snippets are applicable in real-world scenarios.

PR #1188, opened 2020-08-20
cmp :  to compare two arrays

samalien:
In Javascript, to compare two arrays we need to check that
-the length of both arrays should be same,
-the objects present in it are of the same type
-and each item in one array is equal to the counterpart in another array.
this snippet to compare two arrays


PrinceProjects (approved):
good one

Chalarangelo (changes requested):
Thanks for the contribution. However, this snippet has formatting issues (naming, tags, description), as well as is redundant as a more generic snippet for this case exists - [`equals`](https://www.30secondsofcode.org/js/s/equals). Closing on that basis.

PR #1193, opened 2020-09-02
Feature/fix error and optimize

yinsang:
**counter**
Because of the browser 's property, setTimeout will work after at least 4ms ,so 
```js
counter('#logo-default',1, 2000,1, 2000)
```
will use 8000ms while done.


Chalarangelo (changes requested):
Please contribute fixes to individual snippets in separate PRs, so that we can review them more easily. Thank you. Closing on that basis.

PR #1194, opened 2020-09-02
update: add comments and avoid default fileName repeat

yinsang:


Chalarangelo (changes requested):
There's no reason to pass a default here. Also the limitation mentioned is inherent to `JSON.stringify()`, I don't think this needs to be in the example.

yinsang:
Pass object only will get 'undefined.json', I don't think it's a good way.
Repeat pass arguments without fileName will cover the passed file data.
limitation can be compatibled by function but may too complex. 
A comment is needed.

PR #1195, opened 2020-09-02
optmize: change + to easy understanding

yinsang:


Chalarangelo (commented):
I think the original intent here was to showcase the `+` operator, but I'll take a look into the `Number()` option, it might be easier to read and understand. I'll get back to you in a bit.

Chalarangelo:
Will leave as-is to showcase the unary operator.

PR #1197, opened 2020-09-02
fix: counter error course of browser property

yinsang:
**counter**
Because of the browser 's property, setTimeout will work after at least 4ms ,so 
```js
counter('#logo-default',1, 2000,1, 2000)
```
will use 8000ms while done.

### DIFF
--- a/snippets/counter.md
+++ b/snippets/counter.md
@@ -5,26 +5,45 @@ tags: browser,advanced
 
 Creates a counter with the specified range, step and duration for the specified selector.
 
-Check if `step` has the proper sign and change it accordingly.
+Inscrease and reduce.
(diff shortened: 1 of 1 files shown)


Chalarangelo (changes requested):
I don't think this is necessary. This works as intended, however, we could add a note in the explanation about this. Changing the implementation is not required and also breaks backwards compatibility.

yinsang:
Change Comment is a good way.
But 4ms had written in [ES Standard](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-settimeout).

> If timeout is less than 0, then set timeout to 0.

> If nesting level is greater than 5, and timeout is less than 4, then set timeout to 4.

PR #1198, opened 2020-09-02
del: useless variable

yinsang:


Chalarangelo (changes requested):
I think the issue here is not with the existing example, but rather the comment. It should be `// regExp !== regExp2`, instead.

Chalarangelo (approved):

PR #1199, opened 2020-09-02
comments: point out disadvantage of JavaScript prototype

yinsang:


Chalarangelo (changes requested):
This isn't implied anywhere, nor do we provide any example that show something like this. Closing.

PR #1201, opened 2020-09-02
update: change hack way to normal

yinsang:
Set day to 0 is a hack way, recommend to use 1.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date
https://tc39.es/ecma262/#sec-date-objects

### DIFF
--- a/snippets/dayOfYear.md
+++ b/snippets/dayOfYear.md
@@ -10,9 +10,9 @@ Use `Math.floor()` to appropriately round the resulting day count to an integer.
 
 ```js
 const dayOfYear = date =>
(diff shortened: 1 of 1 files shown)


Chalarangelo:
From MDN:

> If the parameter you specify is outside of the expected range, `setDate` attempts to update the date information in the `Date` object accordingly. For example, if you use `0` for `dayValue`, the date will be set to the last day of the previous month.

This is documented and consistent and it also allows us to showcase this behavior for beginners who don't know about it. Will leave as-is.

PR #1202, opened 2020-09-02
update: comments

yinsang:


Chalarangelo (commented):
This is more of an issue with JS itself than the snippet. I'll add a related blog to explain this in detail, instead of updating individual snippets and confusing readers.

PR #1203, opened 2020-09-02
del: useless function

yinsang:
Function delay is same to setTimeout, it's the same as 
```js
const delay = setTimeout;
```
I think it's useless, even costing a lot understanding.


Chalarangelo:
This snippet is an alias and will continue existing as such.

PR #1206, opened 2020-09-04
update: comment for easy understanding

yinsang:


Chalarangelo (commented):
The comment wasn't necessary, however the example could be rid of jquery.

yinsang:
Without comment is not friendly for beginners of JS, Who don't know about Event's property of `this` context.

PR #1207, opened 2020-09-04
hotfix: unfrozenObject in frozenObject return wrong value

yinsang:


shihab45:
#

yinsang:
```js
var obj_a = {a1:1}
var obj_b = {a:obj_a}
Object.freeze(obj_b)
var obj_c = {b:obj_b}
deepFreeze(obj_c)
obj_c.b.a.a1 = 'changedValue'
console.log(obj_c.b.a.a1); // log 'changeValue'

```
While Object.isFreeze(obj_b) return true , but obj_a.a.a1 is unfreezed.

Chalarangelo (changes requested):
This introduces unnecessary steps in most cases. Surely, there should be an easier way to deal with this specific edge-case without affecting performance for all other cases.

Chalarangelo (commented):
On second thought, there is no better option to check this, therefore we can move on with this update. I'll change the description to match and merge.

PR #1209, opened 2020-09-04
deepGet fix:add case and fix Bug

yinsang:
original program: 
```js
deepGet({a:false}, ['a']); // require return false but return null
```
Besides,`null` is so different to `undefined`, not exist property require to return `undefined` instead of `null`.
null stands for exist but has nothing, the value is empty.
undefined stands for undeclared in the object as not exist in it.

### DIFF
--- a/snippets/deepGet.md
+++ b/snippets/deepGet.md
@@ -7,10 +7,10 @@ Returns the target value in a nested JSON object, based on the `keys` array.
 
 Compare the keys you want in the nested JSON object as an `Array`.
 Use `Array.prototype.reduce()` to get value from nested JSON object one by one. 
-If the key exists in object, return target value, otherwise, return `null`.
+If the key exists in object, return target value, otherwise, return `undefined`.
 
 ```js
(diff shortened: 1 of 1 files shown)


Chalarangelo (changes requested):
Never return `undefined` for a programmatically empty value, always return `null`.

The snippet indeed has an issue, we should probably check for emptiness more explicitly instead of implicit checks, which break `false`.

yinsang:
> Never return `undefined` for a programmatically empty value, always return `null`.
> 
> The snippet indeed has an issue, we should probably check for emptiness more explicitly instead of implicit checks, which break `false`.
I still stands on my own opinion!
Please add to 2020 TODO, and talk to other contributors please.

yinsang:
> Never return `undefined` for a programmatically empty value, always return `null`.
> 
> The snippet indeed has an issue, we should probably check for emptiness more explicitly instead of implicit checks, which break `false`.

I still stands on my own opinion!
Please add it to 2020 TODO, and talk to other contributor please.

PR #1210, opened 2020-09-06
Add removeUnwantedFields

bhavyabhut:
add a snippet for clear and remove unwanted fields that exist in an object
add a snippet to generate a random string


Chalarangelo (changes requested):
Thanks for contributing. However, your PR doesn't adhere to guidelines (multiple snippets, formatting issues, syntax/grammar etc). Closing on that basis.

Issue #1175, opened 2020-08-05
js - unflattenObject: fails to correctly restore nested properties


tbasse:
The `unflattenObject` function works fine with the provided example

```
unflattenObject({ 'a.b.c': 1, d: 1 }); // { a: { b: { c: 1 } }, d: 1 }
```

But it won't work as expected on `{ 'a.b': 1, 'a.c': 2, d: 3 }`, returns an object with `a.b.` missing.
It also throws on nested values that are not numbers like `{ 'a.b': 'foo', d: 3 }`, results in `Unexpected token in JSON`.

See https://runkit.com/tbasse/5f2a9929c9b5c9001a4d3a6e for some tests

Sparkadelic177:
I'll like to take a look at this issue.

Chalarangelo:
@Sparkadelic177 Sure, go ahead! :ok_hand:

pachbrasil:
Hum..

Sparkadelic177:
@pachbrasil you can go for it if you want.

shivichoudhary:
sure

Taimoor0217:
@Sparkadelic177 can I please take this?

geekylilgirl420:
Look I am totally lost on here. All I is my ex has access to all my accounts idk how but he is ruining my life.

Taimoor0217:
@Chalarangelo I have opened a [pull request](https://github.com/30-seconds/30-seconds-of-code/pull/1220) for this.
Thankyou !

PR #1220, opened 2020-09-22
Fix issue #1175 and cater additional scenarios

Taimoor0217:
**The Issue(s)**

As shared in [issue#1175](https://github.com/30-seconds/30-seconds-of-code/issues/1175). The current code fails to restore nested properties and doesnot cater for scenairos like
```js
unflattenObject({ 'a.b': 1, 'a.c': 2, d: 3 })
unflattenObject({ 'a.b': 'foo', d: 3 })
```
where either the same key is being refered more than once or the values are non numbers. More detials can be seen in [this](https://runkit.com/tbasse/5f2a9929c9b5c9001a4d3a6e) runkit shared in the issue as well.

An additional scenario which the code failed to handle was when the flat path reffered to particular indices in the arrays values against a key. For example.
```js
unflattenObject({ 'a.b.0': 5, d: 3 })
```
should return something like `{a:{b:[5]}, d: 3}`. However, the code fails to compile.

**Solution**

I fixed the snippet to handle these scenarios along with some improvements. The updated code which passes all these tests can be seen in [this](https://runkit.com/taimoor0217/unflattenobject) runkit.

**Modifications / Improvements**

The previous code uses `JSON.parse()` to parse the manually built string into JSON and then merges it with the accumulator using `Object.assign()`. This happened for ever iteration of `Array.prototype.reduce()`. My code takes a slightly different approach of making the reference to the values as the next accumulator. This ways, I can directly mutate and nested objects without having to merge on each iteration. This arguably improves the performance fo the code.

### DIFF
--- a/snippets/unflattenObject.md
+++ b/snippets/unflattenObject.md
@@ -5,29 +5,27 @@ tags: object,advanced
 
 Unflatten an object with the paths for keys.
 
-- Use `Object.keys(obj)` combined with `Array.prototype.reduce()` to convert flattened path node to a leaf node.
-- If the value of a key contains a dot delimiter (`.`), use `Array.prototype.split('.')`, string transformations and `JSON.parse()` to create an object, then `Object.assign()` to create the leaf node.
-- Otherwise, add the appropriate key-value pair to the accumulator object.
+- Iterate over the object keys using `Object.keys(obj)` and `Array.prototype.forEach()`.
+- Split each key with a dot delimiter (`.`) using `Array.prototype.split('.')`and use `Array.prototype.reduce()` to convert flattened paths to leaf node.
+- If the current accumulator already contains value against a particular key, return its value as the next accumulator.
+- Otherwise, add the appropriate key-value pair to the accumulator object and return value as the accumulator.
+
 
 ```js
-const unflattenObject = obj =>
-  Object.keys(obj).reduce((acc, k) => {
-    if (k.indexOf('.') !== -1) {
-      const keys = k.split('.');
-      Object.assign(
-        acc,
-        JSON.parse(
-          '{' +
-            keys.map((v, i) => (i !== keys.length - 1 ? `"${v}":{` : `"${v}":`)).join('') +
-            obj[k] +
-            '}'.repeat(keys.length)
-        )
-      );
-    } else acc[k] = obj[k];
-    return acc;
-  }, {});
+const unflattenObject = obj =>{
+    var result = {}
+    Object.keys(obj).forEach((k)=>{
+      var keys = k.split('.')
+      keys.reduce(function(acc, e, i) {
+        return acc[e] || (acc[e] = isNaN(Number(keys[i + 1])) ? (keys.length - 1 === i ? obj[k] : {}) : [])
+      }, result)
+    })
+    return result
+}
 ```
 
 ```js
 unflattenObject({ 'a.b.c': 1, d: 1 }); // { a: { b: { c: 1 } }, d: 1 }
+unflattenObject({ 'a.b': 1, 'a.c': 2, d: 3 }); // { a: { b: 1, c: 2 }, d: 3 }
+unflattenObject({ 'a.b.0': 8, d: 3 }) //{ a: { b: [ 8 ] }, d: 3 }
 ```
--- a/snippets/unflattenObject.md
+++ b/snippets/unflattenObject.md
@@ -6,18 +6,25 @@ tags: object,advanced
 Unflatten an object with the paths for keys.
 
 - Use nested `Array.prototype.reduce()` to convert the flat path to a leaf node.
-- Split each key with a dot delimiter (`.`) using `Array.prototype.split('.')`and use `Array.prototype.reduce()` to add or objects against the keys.
-- If the current accumulator already contains value against a particular key, return its value as the next accumulator.
-- Otherwise, add the appropriate key-value pair to the accumulator object and return value as the accumulator.
+- Use `String.prototype.split('.')` to split each key with a dot delimiter and `Array.prototype.reduce()` to add objects against the keys.
+- If the current accumulator already contains a value against a particular key, return its value as the next accumulator.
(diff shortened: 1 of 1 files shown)


snippets/unflattenObject.md
```diff
@@ -5,29 +5,27 @@ tags: object,advanced
...
-            obj[k] +
-            '}'.repeat(keys.length)
-        )
-      );
-    } else acc[k] = obj[k];
-    return acc;
-  }, {});
+const unflattenObject = obj =>{
+    var result = {}
+    Object.keys(obj).forEach((k)=>{
```
Chalarangelo, line 17: I think this could be a `reduce()` with initial value of `{}`.

snippets/unflattenObject.md
```diff
@@ -5,29 +5,27 @@ tags: object,advanced
...
+      var keys = k.split('.')
```
Chalarangelo, line 18: Please add semicolons wherever appropriate (we use semistandard as our styling template).

snippets/unflattenObject.md
```diff
@@ -5,29 +5,27 @@ tags: object,advanced
...
+      keys.reduce(function(acc, e, i) {
+        return acc[e] || (acc[e] = isNaN(Number(keys[i + 1])) ? (keys.length - 1 === i ? obj[k] : {}) : [])
```
Chalarangelo, line 20: Why the assignment to `acc[e]` in `acc[e] = isNaN(Number(keys[i + 1])`? What am I missing?

snippets/unflattenObject.md
```diff
@@ -5,29 +5,27 @@ tags: object,advanced
...
-        )
-      );
-    } else acc[k] = obj[k];
-    return acc;
-  }, {});
+const unflattenObject = obj =>{
+    var result = {}
+    Object.keys(obj).forEach((k)=>{
+      var keys = k.split('.')
+      keys.reduce(function(acc, e, i) {
```
Chalarangelo, line 19: I think you can move `k.split(.)` in this line as the caller and use the fourth argument in the callback to access it like so and remove the previous line:
```suggestion
      k.split('.').reduce(function(acc, e, i, keys) {
```

snippets/unflattenObject.md
```diff
@@ -5,29 +5,27 @@ tags: object,advanced
...
 Unflatten an object with the paths for keys.
 
-- Use `Object.keys(obj)` combined with `Array.prototype.reduce()` to convert flattened path node to a leaf node.
-- If the value of a key contains a dot delimiter (`.`), use `Array.prototype.split('.')`, string transformations and `JSON.parse()` to create an object, then `Object.assign()` to create the leaf node.
-- Otherwise, add the appropriate key-value pair to the accumulator object.
+- Iterate over the object keys using `Object.keys(obj)` and `Array.prototype.forEach()`.
+- Split each key with a dot delimiter (`.`) using `Array.prototype.split('.')`and use `Array.prototype.reduce()` to convert flattened paths to leaf node.
+- If the current accumulator already contains value against a particular key, return its value as the next accumulator.
+- Otherwise, add the appropriate key-value pair to the accumulator object and return value as the accumulator.
+
```
Chalarangelo, line 12: :scissors: the extra empty line.

Chalarangelo (changes requested):
Some issues here and there but overall it seems like a solid fix. Please make the suggested changes and we should be able to merge this soon.

snippets/unflattenObject.md
Taimoor0217: @Chalarangelo  do you mean a nested reduce like this?
```js
const unflattenObject = obj =>{
    return Object.keys(obj).reduce((acc1 , k)=>{
      k.split('.').reduce(function(acc, e, i , keys) {
        return acc[e] || (acc[e] = isNaN(Number(keys[i + 1])) ? (keys.length - 1 === i ? obj[k] : {}) : [])
      }, acc1)
      return acc1
    },{})
}
```
if yes, can you suggest a naming convention for the internal and external accumulators? I am a little confused on what to name these.

Chalarangelo: You could probably call the outer one `res` (as in result) and the inner one could be just `acc`. Does that make sense as far as naming goes?

snippets/unflattenObject.md
Taimoor0217: @Chalarangelo  the assignmet to acc[e] basically adds the necessary nested key value pairs to the accumulator (which is a referenece to a value against the parents key). 
for example, say while resolving `{ 'a.b': 2, a.d: 3 }`.

Initially `acc[e]` is `{ }` so `acc['a']` becomes `{ }` after the assignment. Now, after assignment, the reference to `acc['a']` which is `{}`, is passed to the next call as accumulator.
Now for the next call with `b`, similarly when an assignment is again made to `acc['b']` it goes directly inside `acc['a']` because in this case the accumulator was a reference to `acc['a']'s` values.

I hope this explains.

Chalarangelo: Yeah, it makes sense. Maybe we could unnest this assignment into its own line? It would probably be more readable. What do you think?

Taimoor0217: Yes, I think that would be possible. But we will have to add an extra if statement. Shall I do this?

snippets/unflattenObject.md
Taimoor0217: Yes, it does. Thankyou!

Taimoor0217: updated the name in [this commit.](https://github.com/30-seconds/30-seconds-of-code/pull/1220/commits/351ce48f645efe85108812c691f99e2f3b9800a1)

Taimoor0217:
@Chalarangelo updated code and tests are [here](https://runkit.com/taimoor0217/unflattenobject).

Chalarangelo (approved):
Looks good. Great job!

I'll make a few minor changes and merge this.

Taimoor0217:
Thanks @Chalarangelo !

PR #1222, opened 2020-09-24
Added a snippets to get a random RGB color

GuilhermeOliveiraCasagrande:


snippets/randomRGBColor.md
```diff
@@ -0,0 +1,23 @@
+---
+title: randomRgbColor
+tags: color, rgb, beginner
+---
+
+Returns a object containing random values for red, green and blue
```
omriattiya, line 6: grammar:
> returns **an** object

Chalarangelo (commented):
Thanks for contributing. While the implementation is solid, the issue with this snippet is that there are very few (if any) realistic use-case scenarios where you would need an object with this structrue in JS code. What you usually need is a hex string or an rgb/rgba string, which we have random generators and conversion snippets for.

PR #1223, opened 2020-09-26
Added a snippet to get all unique objects in an array

omriattiya:


Trinityyi (commented):
Thanks for contributing. However, this snippet already exists [here](https://www.30secondsofcode.org/js/s/unique-elements) and `JSON.stringify()` is error-prone for comparing complex objects or class instances (eg. `new Set()` is serialized the same as `{}` ).

PR #1226, opened 2020-09-29
Add snippet execution measurement

omriattiya:
a snippet to measure the execution time of a function (also async snippet)


Trinityyi (changes requested):
Thanks for contributing. However, this snippet [already exists](https://www.30secondsofcode.org/js/s/time-taken) and the existing implementation is preferred due to it returning the function's result after timing it. Closing on that basis.

PR #1227, opened 2020-09-30
update median to be more general.

omriattiya:
Like presented in wikipedia: https://en.wikipedia.org/wiki/Median
[the furmola](https://wikimedia.org/api/rest_v1/media/math/render/svg/b5bfb7d81899f91c704e7931eb315affa86af0c6)

### DIFF
--- a/snippets/median.md
+++ b/snippets/median.md
@@ -10,12 +10,12 @@ Returns the median of an array of numbers.
 
 ```js
 const median = arr => {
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks for contributing. From what I can tell, both implementations have the exact same result for all the cases I could find. Also the existing one is more readable. I think we should leave as-is.

PR #1228, opened 2020-10-01
Added Snippet to get live Covid Data

saviomartin:
### Hi bro,

I have developed a cool snippet that can be installed with `npm` and **works to get the live data of our locality.**
The package to be installed is `novelcovid`

I have **clearly shown how to do so** and its **advanced things.**
I have also provided the **complete snippet code** to get the **live COVID data of the world**.

Thank you
**Savio Martin**

### DIFF
--- a/snippets/getCovidLiveData.md
+++ b/snippets/getCovidLiveData.md
@@ -1,0 +1,74 @@
+---
+title: Get Covid Data Using NovelCovid Module
+tags: array,node
+---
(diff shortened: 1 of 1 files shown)


Chalarangelo (changes requested):
Thanks for contributing. We do not allow 3rd-party packages, documentation or generally snippets that do not follow the exact format specified in the templates.

PR #1230, opened 2020-10-01
Get Random Item From an array

saviomartin:
Added Snippet that will get a random item from an array of items.
Nothing to install. No issues. Work perfectly.

**Hope It helps
Thankyou**


Chalarangelo (commented):
Thanks for contributing, but this already exists as [`sample`](https://www.30secondsofcode.org/js/s/sample).

PR #1233, opened 2020-10-02
Snippet for merging two arrays of objects with a common field

PrathamDogra:
* If we have two arrays of objects of different sizes and different fields with only one field in common and we need to combine them then this method is one of the simplest ways to achieve that.


Chalarangelo (commented):
Thanks for contributing. However, this snippet can be produced as a combination of other snippets (mainly using `merge` and a `.map()`), has a complexity of O(N^2) and ignores any elements in `a2` that are not present in `a1`. Additionally, there are formatting issues (missing description, missing result) and the snippet is not dynamic (using `field` instead of a dynamic field name for this), meaning its use cases are limited. Closing on that basis.

PR #1234, opened 2020-10-03
Use modern asynchronous clipboard API

Richienb:


Chalarangelo (commented):
Thanks for contributing. However, the change modifies the snippets' footprint (`async`) and is also an implementation detail mentioned as a notice. The existing snippet should remain as-is as it has better browser compatibility, therefore we are not going to make changes to it for a few more years probably.

PR #1235, opened 2020-10-03
Add formatTime snippet

BattleRattle:
This PR adds a snippet I recently needed myself in another project. So I'd like to contribute it here :)

### DIFF
--- a/snippets/formatTime.md
+++ b/snippets/formatTime.md
@@ -1,0 +1,31 @@
+---
+title: formatTime
+tags: date,math,string,beginner
+---
+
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks for contributing. However, the snippet's behavior is wrong if the duration is longer than 24h and we already have a very similar snippet [here](https://www.30secondsofcode.org/js/s/format-duration).

BattleRattle:
Hi, thanks for reviewing.

I wouldn't consider the behavior as wrong, as it was intentionally a 24-hour format. And I was aware of the similar, existing snipped, but I focussed more on the different format here, as padding with leading zeros can be challenging for JS beginners.

PR #1236, opened 2020-10-03
Added isAnyKeyInObjectNullOrUndefined snippet

domneedham:


Chalarangelo (commented):
Thanks for contributing. However, this snippet is highly error-prone when working with protypical inheritance. Consider the following example:

```js
let x = { a: 1};
let y = { b: null };
Object.setPrototypeOf(x, y);
console.log(x); // { a: 1 } - `b` is not part of `x`, but rather its prototype
isAnyKeyInObjectNullOrUndefined(x); // true, should be false
```

This fails due to the way properties in the prototype chain behave and shouldn't be trusted in real-world scenarios. Closing due to that.

PR #1237, opened 2020-10-03
Add orderByEnum snippet

JanMalch:
Based on the `orderBy` snippet this snippet shows how to sort an array objects with a explicitly defined order for known values.


Chalarangelo (commented):
Thanks for contributing. After tinkering a bit with this I found a more efficient way to implement it, I'll make the changes and merge it. Cool idea.

PR #1241, opened 2020-10-03
Add levenshtein snippet

BattleRattle:
This PR adds an advanced snipped to calculate the Levenshtein distance. I hope you find it as useful as I do :)


Trinityyi (commented):
Thanks, but we are not currently accepting algorithmic implementations due to them not being possible to understand in 30 seconds or less.

PR #1245, opened 2020-10-03
Adding snippet to sum using Clojure concept.

brenohq:
- Uses lexical closure technique to store function definitions.
- Uses recursion to call that definitions, calculating the result incrementally.


brenohq:
hi @Chalarangelo!

Can you help me to understand why this PR it was tagged as invalid?

I didn't find info about when one PR is marked as invalid.

PR #1248, opened 2020-10-03
feat(snippets): add characterCount

izexi:
Adds a function that accumulates the number of occurrences of a character in a string.

### DIFF
--- a/snippets/characterCount.md
+++ b/snippets/characterCount.md
@@ -1,0 +1,18 @@
+---
+title: characterCount
+tags: string,intermediate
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks for contributing. However, this case is already covered by [`frequencies`](https://www.30secondsofcode.org/js/s/frequencies) (although not documented) using `frequencies([...'foo'])`. I'm gonna close this PR, but feel free to open one adding this example to that snippet, if you like.

PR #1252, opened 2020-10-03
Add yardToMeter snippet

justDMNK:


Chalarangelo (commented):
I think we have a few too many of those already. Thanks for the contribution anyways. :smile:

PR #1253, opened 2020-10-03
Add getScrollDirection snippet

BraydenTW:
Returns `true` if scrolled up and `false` if scrolled down.

- ON the call of the function, it sets a scroll event listener to the `window` object 
- Returns `true` if scrolled up and `false` if scrolled down.

```js
function findScrollDirection() {
  let scrollBefore = window.scrollY;
  window.addEventListener("scroll", () => {
    return (scrollBefore > window.scrollY ? true : false);
    scrollBefore = window.scrollY;
  });
}
```

```js
findScrollDirection(); // Returns 'true' if scrolled up. Returns 'false' if scrolled down.
```

### DIFF
--- a/snippets/getScrollDirection.md
+++ b/snippets/getScrollDirection.md
@@ -1,0 +1,23 @@
+---
+title: getScrollDirection
+tags: function,type,intermediate
+---
+
+Returns `true` if scrolled up and `false` if scrolled down.
+
+- ON the call of the function, it sets a scroll event listener to the `window` object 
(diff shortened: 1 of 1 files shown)


Chalarangelo (changes requested):
Thanks for contributing. However, there are a few guideline issues (wrong tags, code formatting) and also the snippet doesn't account for left and right scroll as-is.

Chalarangelo:
Closing due to inactivity. Feel free to open a new PR if you want with the issues fixed.

PR #1254, opened 2020-10-03
Create countWords snippet

k3nsei:
Snippet describes how Set can be useful. In this example we counting words.

### DIFF
--- a/snippets/countWords.md
+++ b/snippets/countWords.md
@@ -1,0 +1,22 @@
+---
+title: countWords
+tags: string,beginner
+---
+
+Return number of words in given string
+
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks for contributing. However, the implementation is far too complex for the desired result, the code doesn't follow all guidelines and the real-world value of the snippet is a little limited. Closing due to that.

k3nsei:
I cannot agree with you. There is nothing complex there and real-world value is for example to count used words in document. If you have some objections just write them down and I would adjust my snippet. It's kinda rude behaviour to close PR just like that.

PR #1258, opened 2020-10-03
Add monthName snippet

teeteejo:
Add monthName to retrieve the name of the month from the given date


Chalarangelo (commented):
Thanks for contributing. However, this snippet provides little to no value in real-life scenarios, so we'll close this PR due to that.

PR #1259, opened 2020-10-03
Add addDays snippet

teeteejo:
Add given number of days to a date

### DIFF
--- a/addDays.md
+++ b/addDays.md
@@ -1,0 +1,24 @@
+---
+title: addDays
+tags: date,beginner
+---
+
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks for contributing. However, this snippet provides little to no value in real-life scenarios, so we'll close this PR due to that.

teeteejo:
Thank you for reviewing this PR @Chalarangelo. Even though it looks simple, it solves some special cases. Please check this - https://stackoverflow.com/a/19691491/295697

Thanks

PR #1260, opened 2020-10-03
Add removeAccents

danielesreis:


Chalarangelo (commented):
Thanks for contributing. However, there is a significantly easier way to write this, using `str.normalize("NFD").replace(/[\u0300-\u036f]/g, "")`. Could you please update your PR to use that instead?

Chalarangelo (approved):

danielesreis:
> Thanks for contributing. However, there is a significantly easier way to write this, using `str.normalize("NFD").replace(/[\u0300-\u036f]/g, "")`. Could you please update your PR to use that instead?

Done. Thank you for the review.

PR #1261, opened 2020-10-04
Add isInternetExplorer snippet

piraces:
Returns `true` if the function is executed in an Internet Explorer browser and `false` if it is any other browser.

Useful, to check and avoid compatibility issues in webpages.


Chalarangelo (commented):
Thanks, but IE is going away very soon. This is pretty much already redundant.

PR #1262, opened 2020-10-04
Add isTouchScreen snippet

piraces:
Returns `true` if the function is executed in a touch screen based device and `false` otherwise.

Useful, to check for mobile devices or tablets.

### DIFF
--- a/snippets/isTouchScreen.md
+++ b/snippets/isTouchScreen.md
@@ -1,0 +1,16 @@
+---
+title: isTouchScreen
+tags: browser,intermediate
+---
+
+Checks whether the website is being opened in a device with touch screen.
+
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks for the contribution. However, there is no failsafe way to detect a touchsreen, nor does it also help that different devices support different APIs. This might return `false` for older touchscreen devices, while it might also return `true` in other environments where `touchstart` event is not supported for example. The latter is rather problematic as you should theoretically be able to handle touch events, but can't. I'm gonna close due to that, as it's a tricky problem to handle correctly and dealing with browser discrepancies is out of the scope of regular snippets for the most part.

PR #1271, opened 2020-10-04
Add mixTwoColors

YordanIliev2002:
This snippet is used to "mix" two hex colors. It is done by taking the average of the red/green/blue for each of the two colors.
The example that is given is red+blue=purple. Hope that you like it :smile:


Chalarangelo (commented):

YordanIliev2002:
I will fix the commit so that it matches the guidelines. If the formatting is right, will it be approved because of the "and/or it doesn't provide enough value for the average reader" thing?

PR #1275, opened 2020-10-04
Add isDuplicates

jun-low:
A simple but useful check for duplicate values snippet.


Chalarangelo (approved):
Thanks for contributing. The snippet is good, but there are some guideline issues with it. I'll fix them, but take a closer look at the changes if you are going to submit any more snippets to the repo. Cheers.

PR #1280, opened 2020-10-04
Add new snippet

mailysantos:
Return the maximum elements from an array


Chalarangelo (commented):

PR #1282, opened 2020-10-04
Add new snippet

mailysantos:
Returns an array with the max value in the principal array


Chalarangelo (commented):

PR #1283, opened 2020-10-04
Add new snippet

mailysantos:
Filtering an array of objects based on a condition


Chalarangelo (commented):

PR #1284, opened 2020-10-04
Add template snippet

rubiin:
A simple but useful snippet on how to create a template function


Chalarangelo (commented):

PR #1295, opened 2020-10-04
Update CSVToArray.md

jikuja:
Added notes about the snippet


Chalarangelo (commented):
Thanks for the contribution. However, it would be far better to improve the code, as the note is pretty much reflected in the use cases already.

PR #1297, opened 2020-10-04
added randomString

rubiin:
A oneliner for generating random string


Chalarangelo (approved):
Thanks for contributing. The idea is solid, but the snippet has some small issues with formatting and usability. I'll make the necessary changes and merge it.

PR #1298, opened 2020-10-04
added convert-array-of-objects.md

jikuja:


Trinityyi (commented):
Thanks for contributing. However, your PR doesn't follow our [contribution guidelines](../blob/master/CONTRIBUTING.md) correctly (eg. title doesn't match file name,indentation is incorrect) and/or it doesn't provide enough value for the average reader. Closing due to that.

jikuja:
Oh god. This is a really hostile repository. Just sen everything into the trash without any feedback loop. Would not recommend it to anyone.

> it doesn't provide enough value for the average reader

Any guidelines for this? I'm running that code in production and it's greatly simplifying other parts of code.

PR #1300, opened 2020-10-05
formatDateShort.md

DanielVip3:
It basically is a browser way to format a date object to a DD/MM/YYYY string (or MM/DD).
Also supports time and 12-hours time system.

Uses the modern window.Intl object, but even supported from IE11.

I tried to follow all guidelines, tell me if something is wrong.


DanielVip3:
I don't know why it pushed older commits too. I'm gonna close this and re-open a new PR.

PR #1302, opened 2020-10-05
added binaryToDecimal.md

ranjitmnair:
added a snippet which converts a binary number to a decimal number


Trinityyi (commented):

PR #1305, opened 2020-10-05
Program that checks substrings

rootnarayan:
A simple program to check if the second string given is a substring of the first using recursion.


Trinityyi (commented):
This is the same as [`String.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes), just very poorly implemented. Closing due to that.

PR #1309, opened 2020-10-05
Added toFixed.md snippet

pritam047:


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [round](https://www.30secondsofcode.org/js/s/round).

PR #1310, opened 2020-10-05
add mergeSort.md

Taimoor0217:
**Is this a new snippet ?**
Yes, it is.

**About the snippet***
This snippet implements mergeSort using the divide and conquer algorithm

**Running Example**
The live example can be found in the runkit below.
https://runkit.com/taimoor0217/mergesort


Trinityyi (commented):

PR #1311, opened 2020-10-05
added parseYoutubeUrl

vishal-sancheti:
Is this a new snippet?
Yes, it is.

About the snippet*
This snippet implements parseYoutubeUrl using regex to get youtube video id from url.

### DIFF
--- a/snippets/parseYoutubeUrl.md
+++ b/snippets/parseYoutubeUrl.md
@@ -1,0 +1,28 @@
+---
+title: parseYoutubeUrl
+tags: youtube,URL parsing,intermediate
+---
+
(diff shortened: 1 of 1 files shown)


Trinityyi (commented):
Thanks for contributing. However, we are not currently any snippets that depend on 3rd-party services, libraries or APIs.

vishal-sancheti:
This doesn't use 3rd party API, Library, or Service. It just parses the URL String to fetch Video Id

Chalarangelo:
@vishal-sancheti But it depends on a 3rd-party service, so it counts as such.

PR #1314, opened 2020-10-05
Added nTimesString

jp9573:


Trinityyi (commented):
Thanks for contributing. However, I don't think this PR provides a lot of value for most of our readers. It also doesn't work with the provided example. Closing on that basis.

PR #1316, opened 2020-10-05
Added emsoChecker.md

febo99:


Trinityyi (commented):
Thanks for contributing. However, I don't think these provide a lot of value for most of our readers, except for the utility they do indeed provide. Closing on that basis.

PR #1319, opened 2020-10-05
Added formatToThousands

josno:


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [formatNumber](https://www.30secondsofcode.org/js/s/format-number).

josno:
Should have checked, thanks!

PR #1321, opened 2020-10-05
Added snippets/convertArrayOfObjectsToObjectOfObjects.md

jikuja:
New revision of the snippet:
 * Fixed code to follow contribution guidelines as suggested

Please let me know snippet still has some technical issues. The previous close message looked like a canned message and therefore I cannot be sure about **all** intentions of the close message.


Trinityyi (commented):
Stop spamming the same PR over and over. This provides no value to our users whatsoever. Thanks.

PR #1325, opened 2020-10-05
Add litersToGallons and gallonsToLiters snippets

pataar:
These might come in handy when comparing gas prices for example.


pataar:
@

PR #1326, opened 2020-10-05
Add detectLanguage.md

pataar:
Comes in handy to detect the language of the user to set is as default preference.


pataar:
Hey @Chalarangelo, do you think that this would fit in the snippet collection?

Chalarangelo (approved):
Thanks for contributing. This seems interesting, but it has some issues with guidelines. I'll go ahead and fix them before merging, but be more careful if you contribute any more snippets.

pataar:
Will do 👍

PR #1332, opened 2020-10-05
Create strongPasswordGenerator.md

syamilAbdillah:


Trinityyi (commented):
Thanks for the contribution. However, I don't think we should be adding snippets that handle sensitive information pertaining to user/account security and providing guidelines on how to implement them as each application/website/software is different and should have its own rules and make its own decisions based on the kind of security measures it needs/wants to take. I'm closing on that basis, as the maintenance cost for such an implementation would be coming back and revising it often based on guidelines, which I doubt could ever happen as our resources are quite limited.

PR #1333, opened 2020-10-05
Add substring.md

vsmjr:
How to find a substring in a string using Python.


Chalarangelo (commented):

PR #1334, opened 2020-10-05
Added snippet to make 3rd party API calls with vanilla JS

ahsankhan26:


Trinityyi (commented):

PR #1335, opened 2020-10-05
Add trimWhiteSpaces snippet

chilupa:
Added `trimWhiteSpaces` snippet which removes excess white spaces that are appended to a string from both ends.


Chalarangelo (commented):
Thanks for contributing. However, I don't think this PR provides a lot of value for most of our readers, as this is part of the language API. Closing on that basis.

PR #1338, opened 2020-10-05
Add withTimedReject

sesam:
This snippet is something I use to cook up some simple async action flows.

Imagine having a debounced search query as the input, then wait a bit and perform a request, but use timeout to be able to retry the request. If you have an existing app that already used a promise (like waiting for a GET or POST request) then this is easy to just plug in and move on.

### DIFF
--- a/snippets/withTimedReject.md
+++ b/snippets/withTimedReject.md
@@ -1,0 +1,43 @@
+---
+title: withTimedReject
+tags: promise,intermediate
+---
+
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks for contributing. However, your PR doesn't follow our [contribution guidelines](../blob/master/CONTRIBUTING.md) correctly. Closing due to that.

PR #1342, opened 2020-10-06
Add CSVToYAML snippet

amirkarimi:
Insipired by `CSVToJSON` but not exactly the same.


Chalarangelo (commented):

amirkarimi:
How `CSVToJSON` adds value but `CSVToYAML` that clearly shows how you can leverage `map` and `join` functions to transform data in addition to the idea of using `JSON.stringify` to properly escape the strings for these purposes doesn't?

PR #1344, opened 2020-10-06
Add: Mean

widyakumara:
Add snippet to calculate mean (average) value on an array of numbers.


Chalarangelo (commented):
Thanks, but this is a duplicate of [median](https://www.30secondsofcode.org/js/s/median).

widyakumara:
AFAIK mean != median :D

PR #1345, opened 2020-10-06
Added snippet `RGBToCMYK`

Melvin-Abraham:
This PR adds a new snippet `RGBToCMYK` for converting RGB values to CMYK format.

### DIFF
--- a/snippets/RGBToCMYK.md
+++ b/snippets/RGBToCMYK.md
@@ -1,0 +1,32 @@
+---
+title: RGBToCMYK
+tags: math,intermediate
+---
+
+Converts RGB color tuple to CMYK (Cyan, Magenta, Yellow, Key _(black)_) format as percentage values.
+
+- Initially, the RGB values are normalized by dividing each of them by 255.
+- Then, value of `key` (black) is calculated as `1 - max(r_norm, g_norm, b_norm)`.
+- Finally, the values of Cyan, Magenta and Yellow is calculated accordingly.
+- This function accepts value of R, G, B each in the range of 255 and returns an object with C, Y, M, K values as percentages.
+
+```js
+const RGBToCMYK = (r, g, b) => {
+  const r_norm = r / 255;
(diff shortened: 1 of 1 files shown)


fejes713:
Hey @Melvin-Abraham, thanks for your effort and submission.

I am not entirely sure we accept these kinds of snippets considering colors in **CMYK** format are not used in the digital world at all. 

based on this [source](https://99designs.com/blog/tips/correct-file-formats-rgb-and-cmyk/):
> When to use CMYK?
> _Use CMYK for any project design that will be physically printed, not viewed on a screen._

I honestly do not see a use case for such a snippet. 

I'll let @Chalarangelo make the final decision here.

Chalarangelo:
I agree with @fejes713 on this one. CMYK is a print-only format which doesn't really apply to web technologies. I'll close this due to that. Thanks for contributing anyhow.

Melvin-Abraham:
@fejes713 @Chalarangelo Your point seems fair but there's one important point to consider. If you read through [this](https://www.printingforless.com/color.html) article, you might notice the following:

> Certain RGB colors that you can see on your monitor (in particular, bright blue, green and red) cannot be replicated with standard CMYK inks. These unachievable RGB colors are said to be "out of the CMYK color gamut." **When selecting colors for your print project, use CMYK color builds to avoid potential RGB conversion issues.**

So, the main application of the CMYK color space is in _image manipulation applications that gives more accurate color representations for printers_ and avoids the suspense of prints coming out with inaccurate colors (e.g. for billboards, posters, etc.). This is implemented by some of the reputed softwares like Photoshop, Illustration, etc. (just refer to the link that was sent by you: [source](https://99designs.com/blog/tips/correct-file-formats-rgb-and-cmyk/)).

As an example, I could think of an electron.js based image manipulation application making use of CMYK color space optimizing for prints which could make use of this snippet.

Do reconsider this point and if you think that's valid, I would be happy to reopen this PR...

PR #1348, opened 2020-10-06
Added linear interpolation snippet

RealPeha:


Chalarangelo (commented):
Thanks for contributing. Closing as a duplicate of [this snippet](https://www.30secondsofcode.org/js/s/map-num-range).

RealPeha:
No problem, but it seems to me that linear interpolation is a little different: https://en.wikipedia.org/wiki/Linear_interpolation

PR #1351, opened 2020-10-06
add(isDefined): one liner for isDefined

computnik:
Added one line implementation for isDefined

### DIFF
--- a/snippets/isDefined.md
+++ b/snippets/isDefined.md
@@ -1,0 +1,23 @@
+---
+title: isDefined
+tags: type,beginner
+---
+
+Returns `true` if the specified value is defined (not null or undefined), `false` otherwise.
+
+- Use the equality operator (non-strict) to check if the value of `val` is not equal to `null`.
+
+```js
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks, but this snippet's implemenation is wrong. It's also a duplicate of sort to [this snippet](https://www.30secondsofcode.org/js/s/is-null).

computnik:
@Chalarangelo thanks for pointing out the duplicate and closing the issue 👍. This snippet checks for both `undefined` and `null` though. On checking further, I feel its similar to [isNil](https://www.30secondsofcode.org/js/s/is-nil).

Also, just a quick note on correctness, Please check [ECMAScript 11.9.3 Abstract Equality Comparison Algorithm](https://www.ecma-international.org/ecma-262/5.1/#sec-11.9.3) (Point 2 & 3) 

Also, I am not sure which one is performant, the strict comparison for `null` & `undefined` or abstract equality comparison. Maybe a good idea to benchmark it. Will give it a shot and share.

PR #1352, opened 2020-10-06
New Snippet: isIdentityMatrix

anaghkanungo7:
- Added a new snippet - isIdentityMatrix - which checks if a given matrix is an identity matrix or not.
- Generated using create-new-snippet as mentioned in the guidelines
- All front matter sections filled

I tried to follow the Contributing guidelines as closely as possible! If I missed anything, or if there's something wrong, please let me know! 

Thanks! 😄


anaghkanungo7:
I just updated the file to enforce strict equality checking.

Chalarangelo (commented):

PR #1354, opened 2020-10-06
Added latinToMilitaryTime

vishal-sancheti:
Added a new snippet - latinToMilitaryTime which takes latin(12h) time string and returns military(24h) time.

> Uses create-new-snippet to generate the snippet
> Doesn't use an external library or API


Chalarangelo (commented):

PR #1355, opened 2020-10-06
Added a new snippet to select a Random Item from Array

JagjiwanChimkar:


Chalarangelo (commented):
Thanks, but this is a duplicate of [sample](https://www.30secondsofcode.org/js/s/sample).

PR #1358, opened 2020-10-06
Remove duplicates in the array snippet

chilupa:
Add JS snippet for removing duplicates in the array.


Chalarangelo (commented):
Thanks for contributing, but this is a duplicate of [uniqueElements](https://www.30secondsofcode.org/js/s/unique-elements).

PR #1359, opened 2020-10-06
Added detectUserAgent.md snippet

madarsbiss:
Added handy snippet to get the information about the name, version and platform of the browser.


Chalarangelo (commented):
Thanks, but I think using the native code is just as easy or even simpler than wrapping it with a function.

PR #1360, opened 2020-10-06
Added hyphenToCamelCase.md snippet

madarsbiss:
Added handy function that turns hyphen-styled text into camelcase text.


Chalarangelo (commented):
Thanks, but [toCamelCase](https://www.30secondsofcode.org/js/s/to-camel-case) already handles that.

PR #1362, opened 2020-10-06
add array to object snippet

arthur-mts:


Chalarangelo (commented):
Thanks, but the PR doesn't follow guidelines and is also a duplicate of [mapObject](https://www.30secondsofcode.org/js/s/map-object).

PR #1363, opened 2020-10-06
Add yieldFilter snippet

JanMalch:
The snippet shows how you can easily huge amount of data and shows possible drawbacks to methods like `Array#filter`.


Trinityyi (commented):
Thanks for contributing. However, we do not support generator function snippets at this time.

PR #1364, opened 2020-10-06
Add to dolar method using Intl lib

arthur-mts:


Chalarangelo (commented):
Thanks, but this is a duplicate of [toCurrency](https://www.30secondsofcode.org/js/s/to-currency).

PR #1367, opened 2020-10-07
feat: add executeMacroTask snippet

RanadeepPolavarapu:
Thought I would share some performance tricks I use from the professional world.


Chalarangelo (commented):
Thanks for contributing. However, I don't think this PR provides a lot of value for most of our readers, as this needs a lot more explaining (not a 30s explanation) to make sense to the average reader. Closing on that basis.

PR #1368, opened 2020-10-07
feat: add executeMicroTask snippet

RanadeepPolavarapu:
Once again, sharing from my snippets from the professional world we use for performance or strict control of execution order where necessary and not in your direct control.


Chalarangelo (commented):
Thanks for contributing. However, I don't think this PR provides a lot of value for most of our readers, as this needs a lot more explaining (not a 30s explanation) to make sense to the average reader. Closing on that basis.

PR #1371, opened 2020-10-07
Added displacementWithAcceleration snippet

SarahLizDettloff:
Calculates the displacement with acceleration

- Displacement equals the original velocity multiplied by time plus one half the acceleration multiplied by the square of time.
- Part of a set of 4 kinematic equations which relate to displacement, velocity, time, and acceleration.
- Velocity in  m/s, acceleration in  m/s/s, time in s


Chalarangelo (commented):

PR #1372, opened 2020-10-07
added a snippet that helps find an object from an array of objects

umang-sinha:
added a snippet that helps find an object from an array of objects using the `find` method to iterate over the array


Chalarangelo (commented):

PR #1374, opened 2020-10-07
Added a snippet that gets a random item from an array

DrBeta:
I am honestly surprised it wasn't there


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [sample](https://www.30secondsofcode.org/js/s/sample).

DrBeta:
Welp all right

PR #1378, opened 2020-10-07
update digitize.md

ujalak1812:


Chalarangelo (commented):
Already fixed in 95c8618

PR #1381, opened 2020-10-07
Add factoryWithSeed snippet

JanMalch:
The snippet shows how you can quickly generate a large amount of unique data, e.g. for mocking. It also shows a handy use case for `map`'s index argument and a JavaScript pitfall (that `Array(n)` has length `n`  but there's nothing to iterate over)


Chalarangelo (commented):
Thanks for the contribution. I'll make any changes necessary and merge this.

PR #1388, opened 2020-10-07
Added matrixMult.md

plooney81:
These functions allow the user to multiply two matrices, of different sizes. The MatrixMult function will return a matrix of the correct size that is the product of the two matrices inputted.


Trinityyi (commented):

plooney81:
My apologies. I thought it may have been a little too long. Thank you for
your reply. Hopefully, I can make a meaningful contribution to yalls
project in the future!

Respectfully,
Peter

PR #1389, opened 2020-10-07
Create fastPower.md

sachinshingade:
Added a fast & efficient way to find power.


Chalarangelo (commented):
This is neither fast or efficient. Also, `Math.pow()` exists. Cheers.

PR #1391, opened 2020-10-07
Create ExponentPower.md

adityakamath16:
Added exponent power of the number.


KLVN:
Come on... this is just a wrapper around an already defined JS-function. On top of that it doesn't even work, because you're not following ES6 notation.

This is like opening a PR for
```js
const log = content => {
  return console.log(content);
};
```
with the difference that this one is actually working.

#shitoberfest

Trinityyi (commented):
^ what he/she said!

PR #1392, opened 2020-10-07
Added timeDiff snippet

helloitsm3:


fejes713:
Hi, Sean 👋 Thanks for submitting this PR. 

I don't see a particular use case for this snippet considering it's a one-liner subtraction. Modern date libraries do support this functionality but with **far advanced formatting** that is not in milliseconds. Things like _2 days ago_, _about a year ago_, etc.

I'll let @Chalarangelo make a final decision here. Note that I think **we had** snippet like this one in the past but just didn't accept it since it's far too low value to be on the website and pollute it.

Chalarangelo:
Thanks for the contribution. Indeed what @fejes713 said is true, as this is a trivial implementation that most people working with dates will learn pretty much instantly. Closing due to lack of value for the average reader.

PR #1394, opened 2020-10-07
Javascript array remove

GrayedFox:
This pull request:

- renames the JavaScript array remove snippet to removeAll (this is what that snippet actually does)
- adds a type safe, ES6 compliant removePrimitive Array.prototype method that safely extends the native prototype

This method has been thoroughly tested - tests for this method can be found [here](https://github.com/GrayedFox/multidict/blob/master/test/globals.spec.js).

The method is named `remove` in the tests however the code being tested in that repository is [nearly identical](https://github.com/GrayedFox/multidict/blob/master/src/globals.js).

Thanks for the great repository and service! Neat idea 👍🏾


Trinityyi (commented):

GrayedFox:
Hi there Trinityyi, thanks for the fast feedback and response. I will take a closer look at the contribution guidelines tomorrow and try to see where I've gone astray.

Just out of interest (if you'll indulge me!), I searched for other examples that extend the native array prototype and couldn't find any (I know most consider it bad practice), so wasn't too sure how to approach that. If I refactor this function to be an arrow function and a standalone method, would you then consider it as perhaps having more value to the average snippet browser?

I understand the desire to keep snippets to a high standard and would like to try and reach that standard, if possible. It's bedtime where I'm at right now - I may also have misread or failed to adhere to too many points in the contributing guide - I'll look over it again.

Aside from that, keep up the good (i.e honest and to the point) work :)

PR #1396, opened 2020-10-07
Added isMultiple.md

Japoncio3k:
isMultiple is a snippet that checks if a number is multiple of another


Japoncio3k:
Also added isCapitalized.md
isCapitalized is a snippet that checks if a string is capitalized

Trinityyi (commented):

PR #1397, opened 2020-10-07
Add Phone Number Formatting Function

aladin002dz:
Added Phone Number Formatting using `Array.prototype.splice()`  and `Array.prototype.join()` , and throw exception if invalid input using `throw` statement.


Trinityyi (commented):

aladin002dz:
Thanks for your review, and for your time, Still wonder how to add a code that add value for reader

PR #1398, opened 2020-10-07
Add isIPv4 snippet

kudaliar032:


Chalarangelo (commented):
Thanks, but we are not accepting validator functions at the moment. See #1402

PR #1399, opened 2020-10-07
Create portValidation.md

PPInfy:
Port number checker


Chalarangelo (commented):
Thanks, but we are not accepting validator functions at the moment. See #1402

PR #1400, opened 2020-10-07
Added detectOrientationChange snippet

shawnlobo96:
Snippet to detect device orientation change.


Chalarangelo (commented):
Thanks but there are multiple issues with this, such as it not running and `Screen.orientation` being an experimental API. Will close due to that.

PR #1401, opened 2020-10-07
Add. Adds objectToArrayOfArrays.md

rushelmac:
This function is useful when a NoSQL document needs to be processed using array functions like 'forEach', 'map' etc. To be used typically when all the nested objects inside the main object are identical (Seen in case of mongoDB database objects with same schema). 
This is my second ever PR to any open source project, please provide guidance in case of any mistakes :p


Chalarangelo (commented):
Thanks, but this is a duplicate of [objectToPairs](https://www.30secondsofcode.org/js/s/object-to-pairs).

PR #1404, opened 2020-10-07
sleepSync: add synchronous sleep snippet

danxgh10:
Also, make async sleep snippet conform to the contributing guide.


Chalarangelo (commented):
Thanks, but the new snippet is a duplicate of [delay](https://www.30secondsofcode.org/js/s/delay).

PR #1408, opened 2020-10-07
New snippet - changeStringToArrayOfChars

bartomiak:
Created new snippet with function that changes given string to array of characters


Chalarangelo (commented):
Thanks for contributing. However there's an easier way to do this using spread. I'll update and merge this.

bartomiak:
Ooo didn't event know that spread is able to do this, thank you!

PR #1409, opened 2020-10-07
Support number as input in palindrome snippet

testica:


Chalarangelo (commented):
Thanks, but this is unnecessary. Numeric input should be transformed beforehand, as we try to make snippets short and handling narrower cases (i.e. no error handling or type conversions).

PR #1411, opened 2020-10-07
Generate alphanumeric string snippet for a given length

chilupa:


Chalarangelo (commented):
Thanks, but this is a duplicate of [randomAlphaNumeric](https://www.30secondsofcode.org/js/s/random-alpha-numeric).

PR #1412, opened 2020-10-07
New snippet hexadecimalToDecimalNumber

Japoncio3k:
Created a snippet that transforms hexadecimal numbers into decimal numbers


Trinityyi (commented):
Thanks for contributing. However this is the same as  `parseInt('35FE4',16)`

PR #1414, opened 2020-10-07
Added limitText.md snippet

GiselaMD:


Chalarangelo (commented):
Thanks for contributing, but this is a duplicate of [truncateString](https://www.30secondsofcode.org/js/s/truncate-string).

PR #1418, opened 2020-10-08
Update hasKey snippet

juanarbol:
It is recommended to not use the `hasOwnProperty` method from the
target, the prototype of the object target could be changed very easily.
eg:

```
const meow = {
  hasOwnProperty: () => true
}
```


Trinityyi (commented):
This makes no sense whatsoever, as the same can be said for the prototype of `Object` itself. If the user or dev is manipulating the methods of a prototype, the problem is theirs, not the snippet's.

PR #1423, opened 2020-10-08
added a function that converts roman numbers to integers

kashyap-sojitra:


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [toRomanNumeral](https://www.30secondsofcode.org/js/s/to-roman-numeral).

kashyap-sojitra:
Hi ,

This is the reverse of what the toRomanNumeral does.

PR #1424, opened 2020-10-08
add snippet for make unique array in js

dao-phuong:
add js snippet for make sure array only contain unique value


Trinityyi (commented):

PR #1419, opened 2020-10-08
Add Gauss sum or sum of k naturals.

juanarbol:

PR #1425, opened 2020-10-08
sum upto first n natural number

hemantsonu20:


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - #1419

PR #1426, opened 2020-10-08
Adds remove primitive snippet

GrayedFox:
Second contribution attempt, this time with Semi-Standard style enforced and without the fancy/unnecessary Array.prototype.extension.

This PR aims to

- add a simplified, accessible removePrimitive array method


Trinityyi (commented):
Thanks for contributing. However, this functionality is already covered by [remove](https://www.30secondsofcode.org/js/s/remove).

PR #1427, opened 2020-10-08
Adding URLToDoc snippet

Subhash3:


Trinityyi (commented):
Thanks for contributing. However, your PR doesn't follow our [contribution guidelines](../blob/master/CONTRIBUTING.md) correctly and it doesn't provide enough value for the average reader. It also uses 3rd-party services which is not allowed. Closing due to that.

PR #1430, opened 2020-10-08
Update getElementsBiggerThanViewport.md

itamar244:
There is no reason to use `.reduce()` when when you don't need to.
`.filter()` is much simpler to reason, easier to read and more verbose


Chalarangelo (approved):
Thanks for the update.

PR #1432, opened 2020-10-08
Create keyBy.md

itamar244:


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [mapObject](https://www.30secondsofcode.org/js/s/map-object).

PR #1434, opened 2020-10-08
Add RGBAToRGB

nhonx:
Add RGBAToRGB snippet - a snippet which is useful to convert RGB-Alpha into RGB color in some cases.


Trinityyi (commented):

PR #1435, opened 2020-10-08
Added tetranacciSequence.md

Japoncio3k:
Created an algorithm that calculates the nth element from the Tetranacci Sequence.


Trinityyi (commented):

PR #1439, opened 2020-10-08
add snippet for power

christangga:


Trinityyi (commented):
Thanks for contributing. However, I don't think this PR provides a lot of value for most of our readers. Also, Math.pow() exists. Closing on that basis.

PR #1440, opened 2020-10-08
Add snippet: purgeArray

DevTony101:
This PR adds a new snippet called purgeArray with the following description: given an array `arr`, returns a copy of `arr` with all the undefined and null values filtered out.


Chalarangelo (commented):
Thanks for the contribution, but this is a duplicate of [compact](https://www.30secondsofcode.org/js/s/compact).

PR #1441, opened 2020-10-08
Added abs function to get absolute value of a number

amnindersingh12:
**The snippet shows how you can find absolute value of a number.**


Chalarangelo (commented):
Thanks, but this is the same as `Math.abs()`.

PR #1442, opened 2020-10-08
Add snippet: domainName

DevTony101:
This PR adds a new snippet called domainName with the following description: given a `url`, returns the domain name.

### DIFF
--- a/snippets/domainName.md
+++ b/snippets/domainName.md
@@ -1,0 +1,19 @@
+---
+title: functionName
+tags: array,intermediate
+---
+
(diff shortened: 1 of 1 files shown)


Trinityyi (commented):
Thanks for contributing. However, there are a number of cases where this doesn't work and it's tricky to write a snippet that actually accomplishes what the native API does (`window.location.hostname`). Consider the following cases:

- `http://localhost:3000`
- `https://docs.google.com`
-  `http://192.168.1.1`
- `ftp://my-domain.com`

PR #1444, opened 2020-10-08
New snippet shortestWordOnArray

dawalberto:


Chalarangelo (commented):
Thanks for contributing. However, I don't think this PR provides a lot of value for most of our readers, while there are also other snippets that cover the same ground. Closing on that basis.

KQJ-Diya007:
It's alright.

PR #1447, opened 2020-10-08
Capitalised the titles of the markdown tables

AppDevMichael:


fejes713:
Hey @AppDevMichael, this PR isn't needed at all considering this is not the standard practice when writing frontmatter. Also, this would break the build and is a very risky change. 

I am closing this for the mentioned reasons.

PR #1448, opened 2020-10-08
added stringRotation.md

PeadarODonnell:


fejes713:
Hey, @PeadarODonnell 👋 Thanks for opening this PR. 

It does look interesting but what would be the use case of this snippet? 

At this stage, we are trying to add quality content to 30-seconds with real-life use cases. This one doesn't seem that useful to me 😕 Could you provide some real-life use cases? Thanks!

Chalarangelo:
Thanks for contributing, but I'm also pretty certain this has no real-life applications for the most part. Closing due to that.

PeadarODonnell:
That's fair enough, it had a really specific use case, ie. if you were looking for a string rotation between two strings of same length but yeah, you's are right when I look at it, not a lot of people will actually be searching for a string rotation in real life :disappointed_relieved: 
But thank you so much for looking at it and giving advice, I definitely want to add only quality content to 30-seconds in the future! :smile:

Bryan-Herrera-DEV:
:3

PR #1449, opened 2020-10-08
Create a cloneObject

ArvedE:
you must clone the Object like that. If you try to clone it like the following
object = {bla: bla}
objectClone = object
and then try to edit it like
objectClone.bla = "blub"
You would also edit the real Object

### DIFF
--- a/snippets/cloneObject
+++ b/snippets/cloneObject
@@ -1,0 +1,20 @@
+---
+title: cloneObject
+tags: object,function
+---
+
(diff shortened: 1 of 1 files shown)


fejes713:
Hey, @ArvedE I am closing this since it's a duplicate of functions we already have on the website. [deepClone](https://www.30secondsofcode.org/js/s/deep-clone) and [shallowClone](https://www.30secondsofcode.org/js/s/shallow-clone)

Thanks for the effort though!

PR #1450, opened 2020-10-08
Added dateFormat snippet

davizardb:
Tried my best to improve my previous PR, this formats the DMY string date format into ISO 8601, can be changed even to different date formats too.

### DIFF
--- a/snippets/dateFormat.md
+++ b/snippets/dateFormat.md
@@ -1,0 +1,21 @@
+---
+title: dateFormat
+tags: string,date,beginner
+---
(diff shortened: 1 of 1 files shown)


fejes713:
Hey, @davizardb 👋 Thanks for trying to contribute.

Unfortunately, this is not the snippet format we are looking for as it's not a function at all. This formatting example is also built-in inside JavaScript so there's no point in adding it manually. 

Due to the mentioned reasons, I am closing this PR.

PR #1451, opened 2020-10-08
added daysBetween.md

nonseodion:
Calculates the number of days between two Date objects


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [getDaysDiffBetweenDates](https://www.30secondsofcode.org/js/s/get-days-diff-between-dates).

PR #1452, opened 2020-10-09
isEmail.md added

devwhodevs:
Checks if the given argument is an Email using RegEx


Trinityyi (commented):
Thanks for contributing. However, this has been brought up before and we are not going to add such a snippet due to the difficulty of checking all cases and also the fact that the regexp will be really hard to understand (and also prone to regexp attacks if you are running on a server). Closing due to that.

PR #1453, opened 2020-10-09
added DECIMALtoHEX

Vipul1142:
a new snippet that changes a decimal number into a hex string


Trinityyi (commented):

Vipul1142:
it's nice that you gave a reason. I will this in mind. thank you.

PR #1455, opened 2020-10-09
Update factorial.md

NihalSuresh007:
Updated Factorial snippet by using iterative approach.


Chalarangelo (commented):
Thanks, but the entire purpose if this snippet is to showcase recursion for beginners. Closing due to that.

PR #1461, opened 2020-10-09
Added the diceRoll snippet

davizardb:
The snippet allows the user rolls a y sided dice x times.


Chalarangelo (commented):
Thanks for the contribution, but this provides very little value for the average reader and the behavior can be replicated by other existing snippets.

PR #1465, opened 2020-10-09
add getAge.md

nonseodion:
Takes the date of birth as a `Date` object and returns the age.


Trinityyi (commented):

PR #1466, opened 2020-10-09
add of replaceAllInQuery

laenNoCode:
replaceAllInQuery is a simple snippet that I have been using for years. Its purpose is to replace all text elements that are in a specific part of the web page. It is usefull when you do templating, to test with certain texts without having to manually replace all in page.


Trinityyi (commented):

PR #1467, opened 2020-10-09
Proposal of a new template way

helderburato:
In this template, I suggest breaking snippets into three sections like the following:
- **Purpose**:  Briefly explanation about the snippet;
- **Explanation**: Set steps with bullets about the snippet;
- **Code Snippet**: Describe the snippet with code.

### DIFF
--- a/snippet-template.md
+++ b/snippet-template.md
@@ -3,12 +3,14 @@ title: functionName
 tags: array,intermediate
 ---
 
-Explain briefly what the snippet does.
+## Purpose
+Explain briefly what the purpose of the snippet.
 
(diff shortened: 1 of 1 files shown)


Trinityyi (commented):
Our current infrastructure supports this exact template, we are not going to change it as we have worked very hard to get everything up and running in a format that is easy to parse with the current tooling. Thanks.

helderburato:
> Our current infrastructure supports this exact template, we are not going to change it as we have worked very hard to get everything up and running in a format that is easy to parse with the current tooling. Thanks.

No problem, It was just a suggestion. :)

PR #1468, opened 2020-10-09
added function to remove any element from an array at specified index

kashyap-sojitra:


Chalarangelo (commented):
Thanks, but this is a duplicate of [shank](https://www.30secondsofcode.org/js/s/shank), which also does not mutate the original array as a side effect.

PR #1469, opened 2020-10-09
Add mergeObject snippet

helderburato:
It merges objects causing mutation only in the first empty object.


Chalarangelo (commented):
Thanks for contributing, but this is a duplicate of [merge](https://www.30secondsofcode.org/js/s/merge).

PR #1470, opened 2020-10-09
Add an universal UUID/GUID generator snippet

Mozartuss:


Chalarangelo (commented):
Thanks, but this is a duplicate of [UUIDGeneratorBrowser](https://www.30secondsofcode.org/js/s/uuid-generator-browser).

Mozartuss:
But this is a universal methode, so you can run teh code with the crypto API (only Browser) or with out the API (Node)

PR #1474, opened 2020-10-09
Create concatNtimes.md

i-ze:
Concatenating string N times


Trinityyi (commented):
Thanks, but this is the exact same as `String.prototype.repeat()`, which is a built-in.

PR #1475, opened 2020-10-09
Added DollarsToCents.md

BKirev:


Chalarangelo (commented):
Thanks for contributing, but we are not currently accepting conversion snippets, as the current ones are already underperforming and under investigation.

PR #1476, opened 2020-10-09
Update flatten to use latest syntax

dmahely:
I updated the `flatten` snippet to use the latest `Array.prototype.flat` method, introduced in ES2019.


Chalarangelo (commented):
Thanks, but we currently support only ES6. Later versions will be supported when browser support is higher.

PR #1477, opened 2020-10-09
added geometric-mean snippet

shrinish123:
tried to add a geometric mean snippet for an array of  numbers


sevnup666:
@

Chalarangelo (commented):
Thanks for contributing but I do not think this snippet has a lot of value for our average reader.

PR #1479, opened 2020-10-10
Add unary plus snippet

Hyftar:
Adds the unary plus snippet, a snippet designed to teach the unary operator to beginners.


Chalarangelo (commented):
Thanks, but I do not think this snippet adds a lot of value for the average reader and we have other examples showcasing that.

PR #1481, opened 2020-10-10
Convert numbers got added

SMAKSS:
A new snippet for converting non-English numbers to English ones got included.


Chalarangelo (commented):
Thanks for contributing, but I do not think this adds a lot of value for the average reader. Closing.

PR #1482, opened 2020-10-10
Add Falsy validator snippet

ipiranhaa:
Create an `isFalsy` which returns a `boolean` response.


Chalarangelo (commented):
Thanks for contributing, but I do not think this adds a lot of value for the average reader. It also suffices to check with `Boolean()` for the most part. Closing.

PR #1483, opened 2020-10-10
📝 Added formatphone helper.

MuhammadHasham23:
The helper function formats any number to US based mobile number.


Chalarangelo (commented):
Thanks for contributing, but I do not think this adds a lot of value for the average reader. Closing due to that.

PR #1484, opened 2020-10-10
Added new snippet for decimal to hex conversion

Melvin-Abraham:
This is a PR for new snippet `toHex` for decimal to hex conversion.


fejes713:
Hey, @Melvin-Abraham 👋 We have had this submission quite a few times already and decided not to include it in the codebase as it's a pretty easy one-line solution built into JavaScript that is not adding any value to our readers.

Thanks for the effort but at this time we won't be going forward and merging this. 🙌

Melvin-Abraham:
@fejes713 I just hoped that this will be merged considering [this PR](https://github.com/30-seconds/30-seconds-of-python/pull/316/files) was merged even though it is a one-liner

Chalarangelo:
Snippet value is evaluated based on techniques showcased, usefulness (which also depends on each language's needs and usage) and potential interest (this is judged based on previous history of interest on sites like StackOverflow etc). This might sound slightly arbitrary, but we are trying to deliver quality content based on our standards. Some existing snippets might not align with those 100%, which is why we are constantly reevaluating our content. Cheers.

PR #1486, opened 2020-10-10
Add validateEmail snippet

Leandrigues:
Created a new snippet to validate if a given string is on email format

### DIFF
--- a/snippets/validateEmail.md
+++ b/snippets/validateEmail.md
@@ -1,0 +1,21 @@
+---
+title: validateEmail
+tags: array,intermediate
+---
+Returns `true` if the given value has a valid email format, `false` otherwise.
+Explain briefly what the snippet does.
+
+- Validate an email based on a regular expression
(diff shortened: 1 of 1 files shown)


fejes713:
Hey, @Leandrigues 👋 Thanks for taking the time to submit this! I indeed think this is a cool little snippet for email validation. 

It has a really good use case although it's implementation is a bit tricky (would be cool to describe the `RegEx` part at least). 

@Chalarangelo @Trinityyi what is your opinion on this one? As far as I remember we had a quite few of those _magic_ solutions that are considered useful and we merged them. Should we give a green light to this one? 🚦

Chalarangelo (commented):
Thanks for contributing, but we've already closed a very similar pr (#1452) due to the difficulty of checking an email, how error prone it is and the fact that it's really hard to explain the regexp used properly.

Chalarangelo:
Thanks for the contribution,

PR #1487, opened 2020-10-10
Add FilterCounter

Justinborzi:


fejes713:
Hey, @Justinborzi 👋 Thanks for the contribution. Unfortunately, I would have to close this one as it has no particular use case and its formatting is not up to date with any of our guidelines. 

Thanks for your time!

PR #1488, opened 2020-10-10
Add exponentiation

avaneeshk098:
I have added a snippet to exponentiate a base.


avaneeshk098:
I have added a snippet to exponentiate a base.

Chalarangelo (commented):
Thanks for contributing, but I do not think this adds a lot of value for the average reader, as it is the same as `Math.pow()`. Closing.

PR #1489, opened 2020-10-10
Add isSubsetByComparator JavaScript array function

GrayedFox:
This pull request adds a JavaScript array method called `isSubsetByComparator` which evaluates if an array is a subset of another array using a comparator function.

Happy hacktoberfest and keep up the good work 👍🏾

### DIFF
--- a/snippets/isSubsetByComparator.md
+++ b/snippets/isSubsetByComparator.md
@@ -1,0 +1,20 @@
+---
+title: isSubsetByComparator
+tags: array,intermediate
+---
+
+Determines whether the first given array is a subset of the second given array based on the output
+of a comparator function.
+
+- Uses `Array.protoype.every()` to iterate over every element of arrA
+- Uses `Array.prototype.some()` to greedily match elements inside arrA to those in arrB
+- Uses a `comparator()` function to determine what exactly a match is
+
+```js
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks for contributing, but this is a duplicate of [`isContainedIn`](https://www.30secondsofcode.org/js/s/is-contained-in).

GrayedFox:
Hmm, it is true that this method achieves the same result but only when the comparator function looks for an exact match.

This method has added functionality in that when calling the method the user gets to define the match (as opposed to it being a direct match). This method therefore can be used to define the matching behaviour (which isContainedIn does not). See the second example:

`isSubsetByComparator((x, y) => x * 2 === y, [1, 3, 5], [2, 4, 6, 8, 10]); // 'true'`

@Chalarangelo would you mind taking another look at this PR? I understand it's a busy time for all of you maintainers, considering it's Hacktoberfest, so I don't mean to be pushy or anything.

I might be mistaken in believing that this is not a true duplicate of isContainedIn. Cheers! 🍺

GrayedFox:
@Trinityyi any chance of getting a 2nd pair of eyes on this? Thank you! 🙇🏾‍♀️ 👓

PR #1490, opened 2020-10-10
Palindrome recursion snippet added

malihassan20:


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [palindrome](https://www.30secondsofcode.org/js/s/palindrome).

malihassan20:
@Trinityyi if you check the code it is being implemented using the Recursion method which can help students if they are looking for a recursive method of it.

PR #1491, opened 2020-10-10
Add removeDuplicatesFromArray snippet

caiangums:
Getting comfortable with `Map` and `Set` JS object is always good and I think this is a tricky and useful usage of `Set` 😄 

Thanks for the awesome repo!


Trinityyi (commented):

PR #1494, opened 2020-10-10
Add isPerfectSquare.md

akxt1022:


fejes713:
Hey, @akxt1022 thanks for taking your time to add this.

**Q:** What would be the use case for such a code snippet? 

Over at https://30secondsofcode.org/, we try to include content that has a wide use case and teaches others good and useful practices.

akxt1022:
Ummm... Yes I'm agree !

Chalarangelo (commented):
Thanks for the contribution, but I'm pretty certain this has no use-cases in the real-world except mostly mathematical ones, in which cases the developer should already be familiar with these concepts, as they are just mathematical operations. Closing due to that.

PR #1503, opened 2020-10-11
Create a toPascalCase snippet

ipiranhaa:


Chalarangelo (commented):
Thanks for contributing, but this is the exact same as [`toCamelCase`](https://www.30secondsofcode.org/js/s/to-camel-case) combined with [`capitalize`](https://www.30secondsofcode.org/js/s/capitalize). It's also not as commonly used in JS as camel, kebab and snake case. Closing due to that.

PR #1507, opened 2020-10-11
Detect if a word is an isogram

nishanth2143:


Chalarangelo (commented):
Thanks for contributing. However, I don't think this adds a lot of value for readers, as its real-life applications are very limited.

PR #1509, opened 2020-10-11
feature(snippets): add dynamic comparator sorting by key

Hiroshii8:
hello sir @Chalarangelo @fejes713 , i would like to add this snippets for dynamicly sorting object by given key.. this comparator helper will be usefull for sorting messy order from array of object base on given key by the user..

please kindly to review this PR.. i would like to receive any feedback.. thank you


Chalarangelo (commented):
Thanks for the contribution, but this is a duplicate of [`orderBy`](https://www.30secondsofcode.org/js/s/order-by).

Issue #1514, opened 2020-10-11
Random Greetings or Words based on the given set of words having similar meanings


Keyurkariya:
Define a set of words having similar meanings and they will be used randomly. 

For example, to greet someone we can use ['Hello', 'Hola', 'Namaste'], etc and it will be randomly used.

Keyurkariya:
Can you assign this to me working on this PR

Keyurkariya:
Here is a PR for the same : https://github.com/30-seconds/30-seconds-of-code/pull/1515

Chalarangelo:
This is probably not very valuable to most readers.

PR #1515, opened 2020-10-11
spintext based on options provided snippet

Keyurkariya:
Based on Issue #1514 

Here is a code to generate sample text based on options and spin your input text.

Happy Hacktoberfest 2020!


Chalarangelo (commented):
Thanks, but this has very little to no value in real-life scenarios. Closing the PR and issue due to that.

PR #1516, opened 2020-10-11
Add generateOTP

rajdeepkhandelwal:
I have added a new snippet called **generateOTP**. please review and let me know what you think.


Trinityyi (commented):
Thanks for contributing. However, we are not currently accepting any snippets that depend on 3rd-party services, libraries or APIs.

PR #1520, opened 2020-10-11
Num to words

tatticoder:
This snippet converts numbers from numeric to text format. 
This snippet would be useful in many settings like making a prefilled bank cheque or making the numbers more accessible to screen readers.


Trinityyi (commented):

PR #1522, opened 2020-10-11
added a function that generates strong random alphaNumericAndCharacte…

kashyap-sojitra:
…rs combination passwords


Chalarangelo (commented):
Thank you, but as mentioned before password checking is beyond the scope of this repo. We are neither security experts nor can we advise people about password and auth system. Closing due to that.

PR #1524, opened 2020-10-11
added function that generates random OTP of desired length

kashyap-sojitra:


Chalarangelo (commented):
Thanks, but OTP is a far more nuanced concept than just generating a n-digit string and calling it a day. The presented implementation is just [randomIntArrayInRange](https://www.30secondsofcode.org/js/s/random-int-array-in-range) with the correct arguments and an `Array.prototype.join('')` at the end. It's also not very useful to the average reader. Closing as such.

PR #1525, opened 2020-10-11
added isEtheriumAddress

rubiin:
Function to validate an etherium address


Chalarangelo (commented):
Thanks but this doesn't provide a lot of value for the average reader, even though the implementation seems solid.

PR #1526, opened 2020-10-11
Add isNode, randomIpAddress and isRegularExpression

dephraiim:
Function to check if the environment is node
Function to generate a random Ip Address.
Function to check for Regular Expressions


Chalarangelo (approved):
Thanks for contributing. I only think `isNode` is a valuable addition, so I'll go ahead and merge just that one!

dephraiim:
LGTM

PR #1527, opened 2020-10-11
Create checkPassword.md

okezieuc:
I created a snippet with regular expressions that checks if a password is long enough and uses letters and numbers.


Chalarangelo (commented):
Thank you, but this has been brought before. Password strength and verification is beyond the scope of this project as we are neither security experts nor can we advise people on how to implement auth systems.

PR #1530, opened 2020-10-12
Add Sieve of Eratosthenes

tumile:
Add a snippet for finding all prime numbers less than `n` using the Sieve of Eratosthenes.


tumile:
Oops there is already an implementation. My bad.

PR #1531, opened 2020-10-12
Count words in string

ZombiMigz:
I find that this snippet is used surprisingly often depending on the project.


Chalarangelo (commented):
Thanks, but this doesn't really add all that much value for the average reader. It's also just a `.length` on top of [words](https://www.30secondsofcode.org/js/s/words).

PR #1532, opened 2020-10-12
Add function to calculate distance between two coordinates

tumile:
Add `coordinateDistanceKm`, which helps calculate Haversine distance between two coordinates.


Chalarangelo (commented):
Thanks, but this doesn't add a lot of value for the average reader. It also doesn't correctly follow our guidelines.

PR #1533, opened 2020-10-12
Omitted the third argument step

datthaprasad:


Chalarangelo (commented):
This doesn't add any value whatsoever, as it alters the core behavior of the snippet. Closing as such.

PR #1534, opened 2020-10-12
Create pageTransition.md

rahuldahal:


Chalarangelo (commented):
Thanks, but this doesn't seem to add a lot of value for the average reader. It also requires a lot of extra steps to actually make it worthwhile in real-life scenarios.

rahuldahal:
Okay, It's fine.
Thanks for the feedback.

PR #1535, opened 2020-10-12
Add toString

georgiosgiatsidis:
Converts a given input to string.

### DIFF
--- a/snippets/toString.md
+++ b/snippets/toString.md
@@ -1,0 +1,18 @@
+---
+title: toString
+tags: string,beginner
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks, but implicit conversion (i.e. ``` `${var}` ```) and explicit conversions (i.e. `.toString()`, `JSON.stringify()` etc.) already exist for pretty much everything as built-ins. Therefore, this snippet doesn't really add a lot of value for the average reader.

PR #1536, opened 2020-10-12
add triangle area snippet

honorezemagho:
Function that calculate triangle Area


Chalarangelo (commented):
Thanks, but this doesn't provide much value for the average reader.

PR #1541, opened 2020-10-12
fix: allow for shorter arrays in isSorted

yoptar:
Also add some style changes like using `Math.sign(direction)` instead of `direction / Math.abs(direction)`


Chalarangelo (approved):
Thanks for the update. Seems like this has some performance improvements as well!

PR #1542, opened 2020-10-12
Kelvin Conversions

lryanle:
Adds 4 conversions between Kelvin and Celsius/Fahrenheit. Should be especially useful for those working on a chemistry or thermodynamic project

Sorry about the extra commits, I made a few typos!


Trinityyi (commented):

PR #1543, opened 2020-10-13
Create removeAllWhitespaces.md

rahuldahal:
Uses `String.prototype.replace()` with a regular expression to replace any and all occurrences of whitespace characters with a empty string.


Chalarangelo (approved):
Thanks for contributing. There are a few minor issues that I'll take care of and merge this.

rahuldahal:
Could you tell me more about those "issues" in a bit detail ?
I will take those mistakes into consideration next time.

PR #1547, opened 2020-10-13
Added islandsCount.md

plooney81:


Chalarangelo (commented):
Thanks, but this doesn't add a lot of value for the average reader.

PR #1548, opened 2020-10-13
added convertTimestampToDDMMYYYYFormat

samishakhurana:
Hello,
I have added a new js snippet that converts timestamp to DD/MM/YYYY format and returns it as a string.


Trinityyi (commented):

khalidalsdiani:
Ok

PR #1553, opened 2020-10-13
Add http OPTIONS snippets

GadsDev:


Chalarangelo (commented):
Thanks for contributing. However, I don't think `OPTIONS` requests are very popular with the average developer and/or necessary in most cases. It's also super simple to tweak the `GET` request snippet for this case, as I'm pretty sure that a developer who knows when to use `OPTIONS` will know how to do so without much effort.

PR #1554, opened 2020-10-14
add initializeArray

Joshua-rose:


Chalarangelo (commented):
Thanks, but this is pretty much `Array.from()` reimplemented.

PR #1555, opened 2020-10-14
Add RGBDivider

CakeCrusher:
Converts an RGB string to an object of the color letters paired with their number.


fejes713:
Hey, @CakeCrusher 👋 Thanks for the PR! 

To me, this looks really interesting and opens a new question to @Chalarangelo and @Trinityyi  such as _do we include color helper methods in 30-seconds-of-code_. 

Maybe the name of your snippet is misleading but **this snippet** looks like it came from _(polished)[https://polished.js.org/docs/#parsetorgb]_ library. It has a use case definitely but not that common compared to others. 

What is your opinion on this one? Snippet like these have a very narrow use case and don't look like some material we currently have on the website 🤷‍♂️ 

**on the other hand** by merging we open quite a huge pool of snippets to be added such as all the ones from polished for example... 

Is this content we are looking forward to having? cc: @Trinityyi @Chalarangelo

Chalarangelo:
@fejes713 I was pondering the same thing tbh. 

The snippet looks and feels like material we should have on the website. However, having far too many of those might be overkill and actually clutter the listing pages with unnecessary stuff. I think a more careful, curated approach as to what we will be adding (and how we will be naming it) might be the best choice here. 

I think this snippet in specific is worth adding, as it has a reasonable amount of use-cases and is easy to understand and use. Others from the same or similar libraries will be evaluated per-case depending on usefulness, historic data (e.g. SO searches and interest in other programming communities), ease of use and actual implementation details. This is probably the best approach, as we can figure it out as we go along.

TL;DR: The snippet is worth adding, I'll review it shortly. Similar material is welcome as a PR, but some snippets might not be merged depending on the usual factors we evaluate.

Chalarangelo (approved):

PR #1556, opened 2020-10-14
Add RGBToCMYK

leviArista:
Adding conversion from RGB to CMYK


Chalarangelo (commented):
Thanks for contributing. This has been brought up before. In general, CMYK conversions are not all that popular and/or useful for the average web developer, so we're not going to add them to the repo at this time.

PR #1557, opened 2020-10-14
Added simple matrix operation snippet

anooppoommen:
Added a snippet for simple matrix multiplication, dot product and transpose operations.


Chalarangelo (commented):
Thanks for contributing. However, your PR doesn't follow our [contribution guidelines](../blob/master/CONTRIBUTING.md) correctly. Also matrices require a lot of explaining for the average user to understand their core concepts, so matrix snippets are not currently something we'll be adding.

PR #1558, opened 2020-10-14
Added celsiusToKelvin method - Converts degrees to kelvin

ayudhkrgupta14:


Chalarangelo (commented):
As mentioned in previous PRs, conversion snippets are not currently accepted and existing ones are being investigated so we can decide if we should keep them or not. Closing.

PR #1559, opened 2020-10-14
Adding isFibonacciNumber snippet

beingageek:


Chalarangelo (commented):
Thanks, but this doesn't really add a lot of value for the average reader.

PR #1562, opened 2020-10-14
Add randomChoiceFromArray

cursorweb:
This is the pull request for `randomChoiceFromArray`. It allows selection from array.

Also, there may be a typo in CONTRIBUTING.md:
> Only modify snippet files, never modify the generated files in the `snippet_data` directory.

Where is `snippet_data`?


Chalarangelo (commented):
Thanks, but this already exists as [sample](https://www.30secondsofcode.org/js/s/sample).

PR #1563, opened 2020-10-14
Add truncateAtSpace

aihowes:
Add's a truncate string utility that is careful about where it is truncated.

Helpful for excerpts that you don't want to be cut off mid-word.


Chalarangelo (approved):
Thanks for contributing. I'll make some minor changes and merge this.

PR #1565, opened 2020-10-14
add getVisibleChildNodes

andy-blum:


Chalarangelo (commented):
Thanks, but this is only useful for some very niche cases and [elementIsVisibleInViewport](https://www.30secondsofcode.org/js/s/element-is-visible-in-viewport) would probably cover most of them by applying for each element. Closing due to that.

andy-blum:
I don't know if this is too niche, but I've used this in my past projects while iterating over page elements to find elements that contain text & images or images and other elements. Any elements that only have images then get a `style="line-height:0;"` applied to them

andy-blum:
Additionally, I don't think this overlaps much with elementIsVisibleInViewport as this isn't checking if the child node is currently visible _in the browser_ but rather returns all the child nodes (text, elements) that the browser _can render_.

PR #1567, opened 2020-10-14
Add formatDate

ruandev:


Chalarangelo (commented):

PR #1568, opened 2020-10-14
Add productOfArray

mccarthykp:


Chalarangelo (commented):
Thanks, but this snippet doesn't add a lot of value and could be written using a very simple `Array.prototype.reduce()` chained to the array itself.

PR #1569, opened 2020-10-14
Add axios snippets

GadsDev:


GadsDev:
I added 2 snippets for http get and post requests using axes with asynchronous functions

Chalarangelo (commented):
Thanks, but 3rd-party libraries in snippets are not supported.

PR #1570, opened 2020-10-14
Adds another example to frequencies snippet

danielesreis:
Since the `frequencies` snippet can be used to count char occurrences in strings as well, I added a new example to it


Chalarangelo (approved):
Thanks for contributing, that's a useful example.

PR #1571, opened 2020-10-14
add focusTrap object

andy-blum:


Chalarangelo (commented):
Thanks, but this snippet is quite long and really hard to explain within 30 seconds or less. Closing due to that.

PR #1572, opened 2020-10-14
Add rafCycle.md

loruta:
Hi,
since I mainly develop web games I've always found myself writing this utility function.
I guess it could also be useful to others!


Chalarangelo (commented):
Thank you for contributing. However, this has some very specific use-cases and is probably not very useful to the vast majority of our audience.

PR #1573, opened 2020-10-14
clearAllTimers snippet

Aynn:
Clears all timer events from a page.

### DIFF
--- a/snippets/clearAllTimers.md
+++ b/snippets/clearAllTimers.md
@@ -1,0 +1,24 @@
+---
+title: clearAllTimers
+tags: setInterval, setTimeout, beginner
+---
+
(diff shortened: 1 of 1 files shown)


Aynn:
Taken from a "Handy JS" file I made a while back - also would 
```js
// Get even index elements from array
Arr.filter((el, i)=>i%2)
 ```
Make for a good snippet?

Chalarangelo (commented):
Thank you, but I think that this has very limited use-cases and, for the most part, if you end up having to clear all intervals, you've probably messed up big time. Let's not endorse bad practices, then.

Aynn:
I disagree on it being a bad practice but fair enough on it being niche cases!

PR #1575, opened 2020-10-15
created insertAtIf snippet

Kevnien:
Inserts an item into an array at index based on a condition.


Chalarangelo (commented):
Thanks, but this doesn't add a lot of value for the average reader.

PR #1576, opened 2020-10-15
created caesarCipher snippet

cypher-adi:


Chalarangelo (commented):
Thanks for contributing. However, your PR doesn't follow our [contribution guidelines](../blob/master/CONTRIBUTING.md) correctly and it doesn't provide value for the average reader. Closing due to that.

PR #1577, opened 2020-10-15
snippet added for Api-handling 

ayush015:
Here is a snippet which explains handling Api over on to the front-end part mostly . It explains promises as well !


Chalarangelo (commented):

PR #1578, opened 2020-10-15
Add SearchingInArray snippet

MrJim-06:


Chalarangelo (commented):
Thanks for contributing. However, your PR doesn't follow our [contribution guidelines](../blob/master/CONTRIBUTING.md) correctly and it doesn't provide any value for the average reader, as it's just a simple `Array.prototype.filter()` for a very specific use-case. Closing due to that.

PR #1580, opened 2020-10-15
Added new snippet distanceFromLatitudeLongitude

adityarajsinghvi:
It helps in calculating the arial distance between the two places provided the latitude longitude of both the places.


Trinityyi (commented):

PR #1582, opened 2020-10-15
Javascript array isSubsetByComparator

GrayedFox:
This function compares to arrays and outputs a boolean value based on a matching (comparator) function.

It's vastly extends the usefulness of [isContainedIn](https://www.30secondsofcode.org/js/s/is-contained-in) by allowing a developer to specify the match behaviour in the form of a comparator function (and achieves this using ES6 features in a readable one-liner).

### DIFF
--- a/snippets/isSubsetByComparator.md
+++ b/snippets/isSubsetByComparator.md
@@ -1,0 +1,20 @@
+---
+title: isSubsetByComparator
+tags: array,intermediate
+---
+
+Determines whether the first given array is a subset of the second given array based on the output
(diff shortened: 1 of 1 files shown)


GrayedFox:
Please note this is *not* a duplicate of isContainedIn (however it does achieve the same result if the matching comparator function takes an exact match form, i.e. `(x, y) => x === y`

GrayedFox:
Thanks for reconsidering 🙇🏾

30secondsofcode:
Closing due to inactivity.

PR #1587, opened 2020-10-15
Add snippet for converting binary to decimal number (JS)

barbaraszott:


Trinityyi (commented):

PR #1589, opened 2020-10-15
Add snippet for checking if at least one element in the array return true for provided callaback (JS)

barbaraszott:


Trinityyi (commented):

PR #1590, opened 2020-10-15
Add factorial

cursorweb:
Add `arbitrarilyLargeFactorial`, which does *not* use recursion, instead, using a `while` loop with bigint so you can go as large as you want.


cursorweb:
also as a side note, can I add multiple snippets in one pull request?
What would be the procedure if one was invalidated?

Chalarangelo (commented):
Thanks, but this is a duplicate of [factorial](https://www.30secondsofcode.org/js/s/factorial). Also, we generally do not accept multiple snippets in PRs, as it takes longer to handle and if one snippet is invalid, we'd probably just close the PR altogether. Cheers.

cursorweb:
ah ok, thanks

PR #1591, opened 2020-10-15
Add getShortMonthName snippet

rheasanjan:
This PR adds getShortMonthName snippet


Trinityyi (commented):

PR #1594, opened 2020-10-15
adds convertToBase()

KovaCro:


Trinityyi (commented):
Thanks for contributing. However, I don't think this PR provides a lot of value for most of our readers. Also this is a buit-in method! Closing on that basis.

PR #1596, opened 2020-10-15
Add email validator

KaustubhDamania:
Added script to determine if the email address entered is valid or not using regular expressions.


Trinityyi (commented):
Thanks for contributing, but we've already closed a very similar pr (#1452) due to the difficulty of checking an email, how error prone it is and the fact that it's really hard to explain the regexp used properly.

PR #1597, opened 2020-10-15
Create replaceAllStringOccurrences.md

ignsteamworks:


Chalarangelo (commented):
Thanks, but [`String.prototype.replaceAll()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) already accomplishes this and is a built-in.

PR #1599, opened 2020-10-15
Add a snippet for nand operator

Leandrigues:
Hello! 

I saw that the common logical operators are in the repo, so I decided to include the nand operator.


Chalarangelo (commented):
Thanks, but the same can be achieved by chaining `not` and `and`, so it's not necessary to add a dedicated snippet for that.

PR #1600, opened 2020-10-15
Add Copy to Clipboard Snippet

chilupa:


Chalarangelo (commented):
Thanks, but this has been brought up before. The Clipboard API is not fully supported yet, so until it is we're sticking to the existing implementation.

PR #1602, opened 2020-10-16
Add RGBAToHex Snippet

chonla:
This PR adds RGBAToHex snippet.


Chalarangelo (commented):
Thanks, but [RGBToHex](https://www.30secondsofcode.org/js/s/rgb-to-hex) covers this for the most part and adding the alpha channel is trivial enough that we don't need an extra snippet just for that.

PR #1603, opened 2020-10-16
Finding slope/gradient of a line

BasilVictor:


Chalarangelo (commented):
Thanks, but this is only useful in very specific cases (drawing graphics), so it's probably not worth adding right now.

PR #1604, opened 2020-10-16
Create CSharpUrlToFile

khoadev94:
Apply new format to snippets and template


Chalarangelo (commented):
Thanks for contributing. However, your PR doesn't follow our [contribution guidelines](../blob/master/CONTRIBUTING.md) correctly and it doesn't provide enough value for the average reader. Also, this is the wrong repo. Closing due to that.

PR #1605, opened 2020-10-16
Add factorQuadratic snippet

gerwinjonathan:


Chalarangelo (commented):
Thanks, but this provides little to no value for the average reader.

PR #1609, opened 2020-10-16
Improve mapObject snippet

lukasbals:
Improve `mapObject` snippet to be able to map arrays containing objects to an object.

Example: 

```js
mapObject(
  [
    { name: "key1", someBoolean: true },
    { name: "key2", someBoolean: false },
  ],
  (el) => el.someBoolean,
  (el) => el.name
); // { key1: true, key2: false }
```

### DIFF
--- a/snippets/mapObject.md
+++ b/snippets/mapObject.md
@@ -3,19 +3,29 @@ title: mapObject
 tags: array,object,intermediate
 ---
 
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
This changes the way the snippet functions entirely and is a completely different thing. I'm quite sure this can be achieved by chaining some existing snippets or would be far simpler to do ad hoc as it's a very narrow use-case and doesn't really serve as a general-purpose example.

PR #1610, opened 2020-10-16
calculate the arial distance between two coordinates

Aman-Maheshwari:
2nd commit code is as per the guidelines which were ask to be followed.


Trinityyi (commented):

PR #1611, opened 2020-10-16
Minmax

nandiya:
This is my snippet code when I need to implement some scaling to some data given from Rest Api.
I implemented math's method to convert a value between 0 and 1.


Trinityyi (commented):

PR #1612, opened 2020-10-16
Add new snippet ( generateUniqueString )

AbrahemAlhofe:
---
title: generateUniqueString
tags: string,random,unique,intermediate
---

Generate unique string

- Use `Math.prototype.random` to create a random number
- Use `Number.prototype.toString` and pass `16` as radix parameter to convert number to hexadecimal number 
- Use `String.prototype.slice` to skip first two characters because `Number.prototype.toString` will return `0.<number>`

```js
const generateUniqueString = () => Math.random().toString(16).slice(2)
```

```js
generateUniqueString(); // f3c9049d08d34
```

### DIFF
--- a/snippets/generateUniqueString.md
+++ b/snippets/generateUniqueString.md
@@ -1,0 +1,18 @@
+---
+title: generateUniqueString
+tags: string,random,unique,intermediate
+---
+
+Generate unique string
+
+- Use `Math.prototype.random` to create a random number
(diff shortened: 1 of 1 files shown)


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [randomAlphaNumeric](https://www.30secondsofcode.org/js/s/random-alpha-numeric).

PR #1613, opened 2020-10-16
Add a function to check the parity of an integer (IE: Even vs. Odd)

chexo3:
It returns `true` for even numbers, and `false` for odd numbers. 
The explanation is a bit long, would appreciate advice on shortening that.
Also, I'm not sure if the messing with bitwise operations provides any real-world benefit.
May also be a good idea to return `NaN` for inputs/array entries that aren't integers or subarrays.

&nbsp;

I'm very new to making pull requests, so any general advice would also be appreciated.

### DIFF
--- a/snippets/check-integer-parity.md
+++ b/snippets/check-integer-parity.md
@@ -1,0 +1,36 @@
+---
+title: checkParity
+tags: array,intermediate
+---
+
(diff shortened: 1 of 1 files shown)


Trinityyi (commented):

chexo3:
Thanks for your response! Will keep that in mind for the future.

PR #1617, opened 2020-10-16
add toHSLObject

fejes713:
In addition to [this PR](https://github.com/30-seconds/30-seconds-of-code/pull/1616) that adds `toHSLArray` I am adding a utility function to restructure it inside an object. 

I went with full `hsl` "component" names _(hue, saturation, lightness)_ due to code readability and use cases. For example, it would be much cleaner to say `lightness * 0.5` to make it darker than doing `l * 0.5` etc. PS I also vote we do this to [`toRGBObject`](https://www.30secondsofcode.org/js/s/to-rgb-object).

Anyways to celebrate 2000 followers on our Twitter bot we made a year ago, I got this Greek candy: 

IMG_7872

### DIFF
--- a/snippets/toHSLObject.md
+++ b/snippets/toHSLObject.md
@@ -10,12 +10,12 @@ Converts an `hsl()` color string to an object with the values of each color.
 - Use array destructuring to store the values into named variables and create an appropriate object from them.
 
 ```js
-const toHSLObject = (hslStr) => {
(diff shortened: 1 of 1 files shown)


Chalarangelo (approved):
:clap: :+1: :tada:

Chalarangelo:
@fejes713  I agree about the rgb object naming, you can PR it, too! :wink:

PR #1620, opened 2020-10-18
Add eventProgess snippet

AdamWier:
Adds a function to help calculate the progress of an event or time span


Trinityyi (commented):

AdamWier:
As you wish. I thought to contribute it because it didn't seem obvious to me (particularly for a beginner and the manipulation of date objects), but a relatively common thing that would need to be done for a progress bar, for example.

PR #1621, opened 2020-10-18
convert hexadecimal to rgb

maurxeugenio:


Trinityyi (commented):

PR #1622, opened 2020-10-18
Adds isWithinDistance

Kitania:
This method provides a way to check if two points in a 2-dimensional plane are within the given Euclidean distance. It uses basic mathematical functions and avoids using slow square root functions.


Trinityyi (commented):
Thanks for contributing. However, the submitted snippet does not seem to have many applications in the real world, except for mathematical ones, which we are not really covering in-depth as they require more explaining for the average user.

PR #1625, opened 2020-10-19
Added snippet for finding unique array of objects

shabab477:
A javascript snippet for finding unique array of objects.

### DIFF
--- a/snippets/uniqueArrayOfObjects.md
+++ b/snippets/uniqueArrayOfObjects.md
@@ -1,0 +1,55 @@
+---
+title: uniqueArrayOfObjects
+tags: array,intermediate
+---
+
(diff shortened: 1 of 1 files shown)


Trinityyi (commented):

shabab477:
Hi @Trinityyi [uniqueElements](https://www.30secondsofcode.org/js/s/unique-elements) snippet will work only for primitive types in Javascript, like numbers and strings. My snippet works for non-primitives as well as primitives. Therefore, I would debate that this is not a duplicate. 🤔

Chalarangelo:
Still, this is a duplicate of [uniqueElementsBy](https://www.30secondsofcode.org/js/s/unique-elements-by).

PR #1626, opened 2020-10-19
Added logical implication snippet

ElectronApplications:


Trinityyi (commented):

ElectronApplications:
There are "xor", "and" and other logical operations snippets. Why shouldn't there be an implication snippet?

PR #1628, opened 2020-10-19
add numberToArrayOfNumber

achigo16:


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [initializeArrayWithRange](https://www.30secondsofcode.org/js/s/initialize-array-with-range).

achigo16:
I didn't realize that, but thanks for your time

PR #1630, opened 2020-10-19
Update redirect.md to prevent XSS

niraj-codes:
If the code is used as is, it can cause Cross Site Scripting vulnerability. Added escape() function to prevent XSS.
Secure code :+1:


niraj-codes:
Opening PR with the account (nerrorsec) registered with hacktoberfest.

PR #1631, opened 2020-10-19
Update redirect.md to prevent XSS

nerrorsec:
If the code is used as is, it can cause Cross Site Scripting vulnerability. Added escape() function to prevent XSS.
Secure code :+1:

### DIFF
--- a/snippets/redirect.md
+++ b/snippets/redirect.md
@@ -14,5 +14,5 @@ const redirect = (url, asLink = true) =>
 ```
 
 ```js
-redirect('https://google.com');
+redirect(escape('https://google.com'));
 ```


Trinityyi (commented):
From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape):
>Warning: Although escape() is not strictly deprecated (as in "removed from the Web standards"), it is defined in Annex B of the ECMA-262 standard, whose introduction states:
… All of the language features and behaviours specified in this annex have one or more undesirable characteristics and in the absence of legacy usage would be removed from this specification. …
… Programmers should not use or assume the existence of these features and behaviours when writing new ECMAScript code. …

PR #1641, opened 2020-10-19
Add color Generator snippet

nawodyaonline:
Generate a hash value for a random color and it will change the color of HTML page's background color
You should configure the HTML file as follows.
Add a button tag with specific id to your HTML file.
By using the ID we write a JavaScript to change the color of Background.


Trinityyi (commented):

PR #1642, opened 2020-10-19
Remainder Snippet

sahuankit010:
This PR contains changes for the snippet which will accept two number. First one will be dividend and second one will be divisor. It will return the remainder.


Trinityyi (commented):

PR #1644, opened 2020-10-19
ADDING: New method to getCurrentTimestampInMilliseconds

BashCloud:


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [getTimestamp](https://www.30secondsofcode.org/js/s/get-timestamp).

PR #1645, opened 2020-10-19
ADDED: util method to detect, if device is offLine or onLine

BashCloud:
…point of time.

### DIFF
--- a/snippets/isDeviceOffline.md
+++ b/snippets/isDeviceOffline.md
@@ -1,0 +1,18 @@
+---
+title: isDeviceOffline
+tags: JavaScript,Browser,intermediate
+---
+
+A simpleset utility method, to check if device went offline.
+
+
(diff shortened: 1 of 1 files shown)


Trinityyi (commented):
From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine):
>In Chrome and Safari, if the browser is not able to connect to a local area network (LAN) or a router, it is offline; all other conditions return true. So while you can assume that the browser is offline when it returns a false value, you cannot assume that a true value necessarily means that the browser can access the internet. You could be getting false positives, such as in cases where the computer is running a virtualization software that has virtual ethernet adapters that are always "connected." Therefore, if you really want to determine the online status of the browser, you should develop additional means for checking. To learn more, see the HTML5 Rocks article, Working Off the Grid.

Therefore, using `navigator.onLine` is generally error-prone and insufficient to query the connection status. Closing due to that.

PR #1649, opened 2020-10-20
fix: refactor by a simple way & add a case

yinsang:


Trinityyi (commented):
Thanks, but this breaks for a bunch of cases. For example: `fromCamelCase('JSONToCSV', '.')` should produce `json.to.csv`, but in your version results in `.j.s.o.n.to.c.s.v` which is quite wrong.

yinsang:
Thanks, need add a case.

PR #1657, opened 2020-10-20
add query string to object function

RomikMakavana:
I've added a file **snippets\queryStringToObject.md**

with function queryStringToObject(url)
that will return object of key value pair of query params form url.

```
queryStringToObject('https://google.com?page=1&count=10'); // {page: "1", count: "10"}
```

Thanks

### DIFF
--- a/snippets/queryStringToObject.md
+++ b/snippets/queryStringToObject.md
@@ -1,0 +1,30 @@
+---
+title: queryStringToObject
+tags: object,intermediate
+---
+
+Returns a key-value pairs of object of query params of given url.
+
+- Use `String.prototype.split()` to split url .
+- Return the object of key value pair of query params of given url.
+
+```js
+
+function queryStringToObject(url){
+    var query = url.split('?');
+    if(query[1]){
+	    var queryParams = {};
+	    var val = '';
+	    query = query[1].split('&');
+	    query.forEach((e) => {
+	        val = e.split('=');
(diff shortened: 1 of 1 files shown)


snippets/queryStringToObject.md
```diff
@@ -0,0 +1,30 @@
...
+```js
+
+function queryStringToObject(url){
+    var query = url.split('?');
+    if(query[1]){
+	    var queryParams = {};
+	    var val = '';
+	    query = query[1].split('&');
+	    query.forEach((e) => {
+	        val = e.split('=');
```
lsmoura, line 20: You could improve this to consider the case where the variable itself contains `=`, like `http://foo.bar?one=one&two=two=two`.

lsmoura (changes requested):
Please use 2 spaces indentation, instead of mixing spaces and tabs, according to the [Snipped Guidelines](https://github.com/30-seconds/30-seconds-of-code/blob/master/CONTRIBUTING.md#snippet-guidelines)

snippets/queryStringToObject.md
Trinityyi: `=` in params are not a valid character and are usually serialized/escaped.

Chalarangelo (commented):
Thanks for contributing. While there are a lot of issues with this snippet, I'll go ahead and add it after fixing everything. Cheers.

PR #1659, opened 2020-10-20
Add snippet to sum N squares

Steffan153:
Sums all numbers from 1 to n, each one being squared first.


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [sumPower](https://www.30secondsofcode.org/js/s/sum-power).

Steffan153:
I actually noticed that, but the difference is that this is in `O(1)` but the other is `O(n)`, so this is more efficient if you're just doing powers of 2 ¯\_(ツ)_/¯

PR #1660, opened 2020-10-20
Added inlineJSON snippet

dasdachs:
Adds `inlineJSON` snippet to stringify and inline a javascript object/array of objects.


Trinityyi (commented):

dasdachs:
That is what I thought as well until I had to inline some json configs for Kubernetes :) But I get you point, thanks for the time!

PR #1662, opened 2020-10-21
Create randomItemInArray.md

sagar627:
This snippets select Random Item from an Array.


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [sample](https://www.30secondsofcode.org/js/s/sample).

PR #1663, opened 2020-10-21
Add new snippet for getURLQueryVariable.

Earle-Poole:


Trinityyi (commented):
Thanks for contributing. However, we consider this as a duplicate of an existing snippet - [getURLParameters](https://www.30secondsofcode.org/js/s/get-url-parameters).

Earle-Poole:
getURLParameters takes no arguments and returns an object from the queryString key/value pairs. 

getURLQueryVariable takes a string as an argument and returns the value associated to that key or undefined if that argument doesn't exist as a key.

Could you elaborate how these are duplicates, but [toTitleCase](https://github.com/30-seconds/30-seconds-of-code/blob/master/snippets/toTitleCase.md) and [capitalize](https://github.com/30-seconds/30-seconds-of-code/blob/master/snippets/capitalizeEveryWord.md) are not?

Chalarangelo:
getURLParameters is a more general version of the same thing. The returned object can be used to get any variable in the parameters.

Title case and camel case are different casings that have different uses, hence the two snippets. Also it's far more convenient when searching for one or the other.

PR #1669, opened 2020-10-21
Create kg to lbs and lbs to kg

ho1234c:
- kg to lbs
- lbs to kg

unification of unit is important.


53jk1 (approved):

53jk1 (approved):

53jk1 (approved):

Trinityyi (commented):

PR #1674, opened 2020-10-21
Fix date in example

g1tman:
calling with `new Date()` will give different result each time. It's better to use a fixed date in example.


Trinityyi (commented):
Closing as the example doesn't add anything useful to the snippet and is a low effort change. Thanks.

PR #1675, opened 2020-10-21
Simplify the code

g1tman:
- For cases n>r in nCr, we should return `NaN`
- Some special cases were not required


Trinityyi (commented):
This alters the snippet's behavior.

PR #1676, opened 2020-10-21
Update averageBy.md

g1tman:
Split line to prevent more than 80 characters in a line


Trinityyi (commented):
Closing as the change doesn't add anything useful to the snippet and is a low effort change. Thanks.

PR #1679, opened 2020-10-21
Update bind.md

ChaboCode:
The script now includes the use of Function.prototype.bind


Trinityyi (commented):
Prototype extension is strictly prohibited in snippets. Thanks.

PR #1682, opened 2020-10-22
removeFrom snippet

jehielmartinez:
removeFrom Snnipet

### DIFF
--- a/snippets/removeFrom.md
+++ b/snippets/removeFrom.md
@@ -1,0 +1,16 @@
+---
+title: removeFrom
+tags: array,beginner
+---
+
+Mutates array by removing a single element in provided index.  
+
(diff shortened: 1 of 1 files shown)


jehielmartinez:
@Chalarangelo can you review this one?
btw Thank you for opening this project for PR'S!

Chalarangelo (commented):
Thanks, but there are multiple issues with this snippet. First off, `Array.prototype.filter()` doesn't mutate the array. If you want to mutate the array, `Array.prototype.splice()` is a built-in that does exactly that, which also makes the snippet redundant. A non-mutating version of splice exists as [shank](https://www.30secondsofcode.org/js/s/shank). Closing.

Issue #1683, opened 2020-10-22
Suggestion : Add httpOptions feature for IT security 


eunbeek:
This repo has a really good example, and I looked for the HTTP method because I worked as an IT Risk analyst before. 
There are DELETE, GET, POST, PUT, REDIRECT method, except TRACE. I usually used TRACE to check XST(Cross-site Tracing) vulnerability.

Can I add HttpTrace method example for this repo? 

I will request pull soon.

----------------------------------------------------------------------------------------------------------------------------------------
// new
I am really sorry but XMLHttpRequest does not accept TRACE now. 
https://domstorm.skepticfx.com/modules/?id=53992f9bfd987e64ab000005

I do not want to break your format and style so I suggest adding OPTIONS method. This method also returns what HTTP methods are used in the WEB site. Hackers run the automation tool to check what method they used and when they figure out misconfiguration, they list up and attack it. 

So I suggest httpOptions feature for your repo. It will be similar to your other code such as 'httpGet' or 'httpDelete'.

Thank you

Trinityyi:
Thanks, but this is outside the scope of this repo. We provide minimal examples. Advanced use-cases and tweaks are up to the user.

PR #1684, opened 2020-10-22
Issue 1683 : Add 'httpOptions' code 

eunbeek:
I request pull about 'httpOptions'.  Whether this code will be merged or not is up to you.
I am just happy to work for your repo.

This code is about HTTP Options method, which will return what method will be allowed on the website. 
Some people need this code to check vulnerability maybe.

It is really similar to other HTTP method code. such as httpGet, httpDelete, or httpPut. 
#1683 

Thanks

### DIFF
--- a/snippets/httpOptions.md
+++ b/snippets/httpOptions.md
@@ -1,0 +1,33 @@
+---
+title: httpOptions
+tags: browser,intermediate
+---
+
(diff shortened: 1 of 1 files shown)


Trinityyi (commented):
Thanks, but this has been brought up before - #1553 . Closing as it's not necessary and not very useful.

PR #1686, opened 2020-10-22
Adding a snippet that returns current date

stefannypanameno93:
The snippet asks for the user to enter the name through a prompt, then shows the name and the current date in the console


Trinityyi (commented):

PR #1690, opened 2020-10-22
Bubble sort algorithm

claudiofus:
Bubble sort algorithm for arrays of numbers, chars and words


snippets/bubblesort.md
```diff
@@ -0,0 +1,33 @@
...
+            if (inputArr[i] > inputArr[i + 1]) {
+                const tmp = inputArr[i];
+                inputArr[i] = inputArr[i + 1];
+                inputArr[i + 1] = tmp;
+                swapped = true;
+            }
+        }
+    } while (swapped);
+    return inputArr;
+}
```
lsmoura, line 27: You usually don't want to change the inputs of a function...

```suggestion
bubbleSort = (inputArr) => {
    const outputArr = [...inputArr];
    const len = outputArr.length;
    let swapped;
    do {
        swapped = false;
        for (let i = 0; i < len; i++) {
            if (outputArr[i] > outputArr[i + 1]) {
                const tmp = outputArr[i];
                outputArr[i] = outputArr[i + 1];
                outputArr[i + 1] = tmp;
                swapped = true;
            }
        }
    } while (swapped);
    return outputArr;
};
```

lsmoura (changes requested):

Trinityyi (commented):

PR #1692, opened 2020-10-22
Added keyify function

TylerGauntlett:


lsmoura (changes requested):
Please use 2 spaces indentation, instead of 4, according to the [Snipped Guidelines](https://github.com/30-seconds/30-seconds-of-code/blob/master/CONTRIBUTING.md#snippet-guidelines)

Trinityyi (commented):
You can use [flattenObject](https://www.30secondsofcode.org/js/s/flatten-object) and `Object.keys()` for this with the added benefit of actually creating the object with the flattened keys.

PR #1693, opened 2020-10-22
Added a function to check if two numbers are coprime

joysamaddar:
Let me know if there are any issues. Glad to be able to contribute to the project.
#hacktober

### DIFF
--- a/snippets/isCoprime.md
+++ b/snippets/isCoprime.md
@@ -1,0 +1,22 @@
+---
+title: isCoprime
+tags: math,beginner
+---
+
+Checks if the provided integer is a prime number.
+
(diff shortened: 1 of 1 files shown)


lsmoura (approved):
It works. If you want to speed up things, check if the numbers are divisible by `2` before the for loop, and start the loop from `3`, going up by two with `i += 2`, like:

```js
for (var i = 3; i <= limit; i+= 2) ...
```

Kudos from going up to the square root!

Trinityyi (commented):
Thanks for contributing. However, the submitted snippet does not seem to have many applications in the real world, except for mathematical ones, which we are not really covering in-depth as they require more explaining for the average user.

PR #1696, opened 2020-10-22
Create optionalProperty.md

DeekshithRajBasa:
create a new object with an optional property


Trinityyi (commented):

PR #1698, opened 2020-10-25
add addListAria

andy-blum:


Chalarangelo (commented):
Thanks, but I don't think this is particularly useful as a snippet. Maybe a future blog with some more context would be a better fit. Closing due to that.

PR #1699, opened 2020-10-26
add changeLightness snippet

fejes713:
as previously discussed with @Chalarangelo I am adding a `changeLightness` snippet to provide more value to folks who are working with color. 

**Yes.** it's a simple modification of `hsl` values but too many this color theory is [_That's Greek to me._](https://en.wikipedia.org/wiki/Greek_to_me)

_**use case:**_
Modifying a single value could be done by hand, but to achieve things dynamically a user needs a method like `changeLightness`. Here is an example of dynamically generated user icons like we see on Figma or Google Docs etc.

image

### DIFF
--- a/snippets/changeLightness.md
+++ b/snippets/changeLightness.md
@@ -1,14 +1,14 @@
 ---
 title: changeLightness
-tags: string,browser,regexp,beginner
+tags: string,browser,regexp,intermediate
 ---
 
-Returns the string value of the color with changed lightness in `hsl` format.
+Changes the lightness value of an `hsl()` color string.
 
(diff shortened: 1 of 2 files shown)


fejes713:
Hey, @Chalarangelo and @Trinityyi do you have an idea if our repo by default passes these [rules](https://hacktoberfest.digitalocean.com/details#rules) for Hacktoberfest? _(asking for a friend)_

Chalarangelo (approved):
Looking good, I'll make a few tiny changes in the description for SEO and merge it asap! :tada:

PR #1701, opened 2020-11-22
Update takeWhile.md

cchandurkar:
The definition of take while seems inverted. When you use `takeWhile()`, what you really say is: out of these array elements `[1, 2, 3, 4]`, take (keep selecting) elements while an element (`n`) is less than 3.  Earlier example `takeWhile([1, 2, 3, 4], n => n >= 3);` was less intuitive as you are asking for elements greater than equal to 3 but actually expect `[1, 2]` in t=return.

### DIFF
--- a/snippets/takeWhile.md
+++ b/snippets/takeWhile.md
@@ -3,10 +3,10 @@ title: takeWhile
 tags: array,intermediate
 ---
 
-Removes elements in an array until the passed function returns `true`.
+Removes elements in an array until the passed function returns `false`.
 Returns the removed elements.
 
-- Loop through the array, using a `for...of` loop over `Array.prototype.entries()` until the returned value from the function is `true`.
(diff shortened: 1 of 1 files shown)


snippets/takeWhile.md
```diff
@@ -11,11 +11,11 @@ Returns the removed elements.
...
-  for (const [i, val] of arr.entries()) if (func(val)) return arr.slice(0, i);
+  for (const [i, val] of arr.entries()) if (!func(val)) return arr.slice(0, i);
   return arr;
 };
 ```
 
 ```js
-takeWhile([1, 2, 3, 4], n => n >= 3); // [1, 2]
+takeWhile([1, 2, 3, 4], n => n < 3); // [1, 2]
 ```
```
maggie-gif, line 21: nn

Chalarangelo (approved):
Yep, seems like it was inverted before.

uguryass:
CcC

PR #1702, opened 2020-11-23
created mode.md

mayur-samrutwar:
Function to find the most frequent element in a list.


Trinityyi (commented):

PR #1706, opened 2020-11-28
feat: add isISOString snippet

lukaskoeller:
`isISOString` is derived from `toISOString` ([see MDN Docs](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString)) and checks whether a given string is an ISO String / Date or not.

### DIFF
--- a/snippets/isISOString.md
+++ b/snippets/isISOString.md
@@ -1,0 +1,21 @@
+---
+title: isISOString
+tags: string,beginner
+---
+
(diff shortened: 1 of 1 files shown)


Chalarangelo (commented):
Thanks for contributing, but this will also evaluate to `true` for something like `2020-10-78T34:99:20.000Z`.

lukaskoeller:
Good point @Chalarangelo. I will make improvements on that point and add it to this PR

PR #1707, opened 2020-11-29
The product of the numbers of a list

FirasHachem:
Multiply numbers of a list with each other


Trinityyi (commented):

PR #1708, opened 2020-12-04
getMousePosition

Judikaelle:
I often use the events to retrieve the position of the mouse. Hoping that it can help


darshjain (approved):

Chalarangelo (commented):
Thanks, but there are multiple issues with the snippet (not following guidelines, whitespace issues, implementation problems when a value is not provided, example not working etc.). Closing.

PR #1710, opened 2020-12-12
Updated Pase Cookie Example check if values have double quotes

CreaTorAlexander:
Hey, 
I've added a check if the values contain double quotes. 
When this is the case then I slice the quotes away.

### DIFF
--- a/snippets/parseCookie.md
+++ b/snippets/parseCookie.md
@@ -15,12 +15,21 @@ const parseCookie = str =>
     .split(';')
     .map(v => v.split('='))
     .reduce((acc, v) => {
-      acc[decodeURIComponent(v[0].trim())] = decodeURIComponent(v[1].trim());
+      // check if values have double quotes
+      // when yes then slice these double quotes away
+    if(v[1].trim().startsWith("\"") && v[1].trim().endsWith("\"")){
+        acc[decodeURIComponent(v[0].trim())] = decodeURIComponent(v[1].slice(1, v[1].length - 2)); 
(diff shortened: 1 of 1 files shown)


dostiharise:
@CreaTorAleXander, 

Nice, I wish I could approve and merge. 

You may need to get some attention of the maintainers. Try tagging them.

snippets/parseCookie.md
```diff
@@ -15,12 +15,21 @@ const parseCookie = str =>
     .split(';')
     .map(v => v.split('='))
     .reduce((acc, v) => {
-      acc[decodeURIComponent(v[0].trim())] = decodeURIComponent(v[1].trim());
+      // check if values have double quotes
+      // when yes then slice these double quotes away
+    if(v[1].trim().startsWith("\"") && v[1].trim().endsWith("\"")){
+        acc[decodeURIComponent(v[0].trim())] = decodeURIComponent(v[1].slice(1, v[1].length - 2));
```
Li9x1, line 21: how about using regex? something like this.. 
`v[1].trim().replace(/^"|"$/g, '')`

CreaTorAlexander: Yes, that's a good solution.
But I have to say I like the way with startsWith and endsWith a bit more.
If you think .replace would make more sense then I can update it :+1:

DrSameerHealth: 👍

Chalarangelo:
Closing as the original issue was not valid.

PR #1711, opened 2020-12-17
Update differenceWith.md

Alex-Sokolov:


Alex-Sokolov:
@Chalarangelo but why? Now there is no examples with objects

PR #1720, opened 2021-01-08
Added storage function

akanshSirohi:


Chalarangelo (commented):

PR #1721, opened 2021-01-08
added a new snippet

MAG-BOSS:
Added a new snippet for storing cookies.


Chalarangelo (commented):

PR #1726, opened 2021-02-16
code for BODY MASS INDEX (BMI)

blaze83media:
Body Mass Index Code


Trinityyi (commented):

PR #1727, opened 2021-02-16
Get window size

blaze83media:
Added new javascript code getwindowSize();


Trinityyi (commented):

PR #1729, opened 2021-02-18
add randomSortArray

priyal27032001:
The function creates a randomly sorted array .


Trinityyi (commented):
Thanks for contributing. However this is a duplicate of an existing snippet - [shuffle](https://www.30secondsofcode.org/js/s/shuffle).

PR #1732, opened 2021-03-16
Add random color snippets

diegoinacio:
- Add randomColor_hex;
- Add randomColor_hsl;
- Add randomColor_rgb;

- Add randomColorN_hex;
- Add randomColorN_rgb;

### DIFF
--- a/snippets/randomColorN_hex.md
+++ b/snippets/randomColorN_hex.md
@@ -1,0 +1,35 @@
+---
+title: randomColorN_hex
+tags: math,random,intermediate
+---
+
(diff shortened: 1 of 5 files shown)


Trinityyi (commented):
Thanks for contributing. However this already exists as the snippet [randomHexColorCode](https://www.30secondsofcode.org/js/s/random-hex-color-code). After that someone can use the appropriate helper snippet (eg. [hexToRGB](https://www.30secondsofcode.org/js/s/hex-to-rgb))

PR #1733, opened 2021-03-18
New Snippet: Reversing an Array

albert3rd:
Hi! I wanted to create a simple snippet that shows how to reverse elements in an array. Let me know if it could use any changes.


Chalarangelo (commented):
Thanks for contributing, but this is already [natively implemented](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse).

PR #1741, opened 2021-04-07
Added an Email Validation Function

mohammedfarish:


Chalarangelo (commented):
Thanks, but email validation is a very complex subject outside the scope of this project. It's also prone to various attacks and can cause security issues if not implemented correctly. Closing on that basis, same as previous PRs for this snippet.

PR #1757, opened 2021-04-22
Add reverseArrayInPlace

sowmenappd:


Chalarangelo (commented):
Thanks for contributing. However, there's already `Array.prototype.reverse()` for this and the snippet mutates the provided array, which is not necessary. It's preferrable to just use `array = array.reverse()` in most cases than create a snippet for this. Closing due to these reasons.

PR #1759, opened 2021-04-22
Create HashFromString

filoucool:
generates a hash from a string and converts it to a 32 bit integer


Chalarangelo (commented):
Thanks for contributing. However, your PR doesn't follow our [contribution guidelines](../blob/master/CONTRIBUTING.md) correctly (naming, indentation, description, feature usage etc) and it doesn't provide enough value for the average reader. Closing due to that.

PR #1764, opened 2021-04-23
Added reverseWords snippet

mandy8055:


Trinityyi (commented):
Thanks for contributing. However, this is simply [words](https://www.30secondsofcode.org/js/s/words) with an `Array.protoype.reverse()` and `Array.prototype.join()` chained to it. We suppose most readers will be able to figure this out without an additionl snippet.

PR #1793, opened 2021-05-11
Add snippet

Yusfuu:


snippets/detectCapsLock.md
```diff
@@ -0,0 +1,25 @@
...
+---
+
+Check if the caps lock is on.
+
+- Suppose you have an input
+- The following shows a if the caps lock is on or is off.
+
+```js
+const detectCapsLock = element => {
+  element.addEventListener('keyup', e => {
```
lodenrogue, line 13: Does this work even if no keys are pressed?

lodenrogue (approved):

Chalarangelo (commented):
Thanks for contributing. However, this only works behind a listener which is not ideal in itself, while the snippet is not correctly formatted according to guidelines. Closing mainly due to low value and low reusability as-is.

PR #1813, opened 2021-06-18
Created a sortOddOnly snippet

jpotts7:
A function that takes in an array of numbers as an argument and sorts only the odd numbers in ascending order, without changing the positions of the even numbers.


Trinityyi (commented):

PR #1814, opened 2021-06-23
Added Radix sort to the snippets

aravindr22:
Request to add a new type of a sort to the snippets. It is one of the best way to sort array of positive integers. where the time complexity of the radix sort is far less than other sort. So it might be helpful for the people who is looking for sorting array of positive integers.


Chalarangelo (commented):
Thanks for contributing. While this is interesting, implementations are required to only have a single arrow function as their output. Feel free to resubmit a valid implementation if you want.

PR #1821, opened 2021-06-28
Fix luhnsCheck's miscalculation for some credit card numbers

### LINKED ISSUE #1817: Wrong result for some inputs
In case a doubled digit equals to 0, the sum is miscalculated and the result is wrong.
For example, the following valid number returns `false` instead of `true`:
```javascript
luhnCheck('4024007174941954');
```
Or a shorter version:
```javascript
luhnCheck('406'); // false, but should be true
```

A possible solution would be to replace the reducer function with this:
```javascript
(acc, val, i) => (i % 2 !== 0 ? acc + val : acc + ((val *= 2) > 9 ? val - 9 : val))
```

The fix has been tested against all [Paypal's credit card numbers for testing](https://developer.paypal.com/docs/payflow/payflow-pro/payflow-pro-testing/#credit-card-numbers-for-testing).

Nehimyeah:
This pull request fixes the miscalculation of the some credit card numbers and closes #1817 according to suggestion given by @tomer-gavish

### DIFF
--- a/snippets/luhnCheck.md
+++ b/snippets/luhnCheck.md
@@ -20,8 +20,7 @@ const luhnCheck = num => {
     .map(x => parseInt(x));
   let lastDigit = arr.splice(0, 1)[0];
   let sum = arr.reduce(
-    (acc, val, i) => (i % 2 !== 0 ? acc + val : acc + ((val * 2) % 9) || 9),
-    0
+    (acc, val, i) => (i % 2 !== 0 ? acc + val : acc + ((val *= 2) > 9 ? val - 9 : val))
   );
   sum += lastDigit;
   return sum % 10 === 0;


snippets/luhnCheck.md
```diff
@@ -20,8 +20,7 @@ const luhnCheck = num => {
     .map(x => parseInt(x));
   let lastDigit = arr.splice(0, 1)[0];
   let sum = arr.reduce(
-    (acc, val, i) => (i % 2 !== 0 ? acc + val : acc + ((val * 2) % 9) || 9),
```
halito22, line 23: ok i think i am learning how this work

snippets/luhnCheck.md
```diff
@@ -20,8 +20,7 @@ const luhnCheck = num => {
...
-    0
```
halito22, line 24: earase this line

halito22 (approved):

Trinityyi (commented):
Thanks but this change does not produce the correct results.

PR #1826, opened 2021-07-12
Add parse url

jiannanya:
- Use a trick of &#60;a&#62;  tag.
- Return an object with elements in the URL.
- Parse protocol, host, port, query, etc.


Trinityyi (commented):

PR #1835, opened 2021-07-17
Fix : using "func" instead of "fn" as the short term for function

Zawad-21:
Assuming that "**fn**" stands for **function**, I propose to use the word "**func**". I think most people, especially beginners are more familiar to the term "**func"** instead of "**fn**" as the short term for **function**.


halito22 (approved):
snippets/converge.md

Trinityyi (commented):
Closing as the PR doesn't add anything useful to the snippet and is a low effort change. Thanks.

PR #1837, opened 2021-07-20
Support for initializing data and passing parameters

xiangwenhu:
Support for initializing data for runPromisesInSeries and pass result to next Promise.


Chalarangelo (commented):
Thanks for contributing, but the implementation doesn't follow guidelines. Closing due to that.

PR #1841, opened 2021-07-24
Added reduceRightSingle.md and reverseArray.md files in the snippets folder

wytan99:
- reduceRightSingle.md
   - can be apply against an accumulator and each value of the array (from right-to-left) to reduce it to a single value.

- reverseArray.md
   - Make first array element becomes the last, and the last array element becomes the first.


Trinityyi (commented):

PR #1847, opened 2021-08-06
added functionExistance snippet

Saidamir-Sh:


Saidamir-Sh:
- Checks if function exists before calling it in order to avoid error

Chalarangelo (commented):
Duplicate of [isFunction](https://www.30secondsofcode.org/js/s/is-function).

Issue #1861, opened 2021-08-31
Make changes in File


code978:
Need to make changes in some of the major files: 
- like 'I' in but to make code more readable we need to make the 'leftIndex in the binary Search file.
- like 2d matrix calculation generally used i , j for row and col but can we make this to row for i and j for the column so that code makes it more readable.
- Instead of using 'el' in a file using an element in that file or any other file that 'el'. 
- like 'r' in but to make code more readable we need to make the 'rightIndex' in the binary Search file. 
- Instead of using the 'val' for storing value can be used to 'value'. 
- Instead of using 'c' for storing character can we used 'character'. 
- Need to add some more files for Node modules 
- Console.count() method is used to count the length of the specified label.
- path.dirname() method is used to get the string path for a specified file.

PR #1862, opened 2021-08-31
Make Changes in project to solve issues #1861

code978:
I make changes so that overall project files makes more readable and appealing to learn. 
- like 'I' in but to make code more readable we need to make the 'leftIndex in the binary Search file.
- like 2d matrix calculation generally used i , j for row and col but can we make this to row for i and j for the column so that code makes it more readable.
- Instead of using 'el' in a file using an element in that file or any other file that 'el'. 
- like 'r' in but to make code more readable we need to make the 'rightIndex' in the binary Search file. 
- Instead of using the 'val' for storing value can be used to 'value'. 
- Instead of using 'c' for storing character can we used 'character'. 
- Need to add some more files for Node modules 
- Console.count() method is used to count the length of the specified label 
- Need to make changes in some of the major files.
- Console.count() method is used to count the length of the specified label 
- path.dirname() method is used to get the string path for a specified file.
- Binary Search is used to make the searching from linear time complexity to logarithm. 
- path.dirname() method provides utility for working with file and directory paths. 
- default operation of the path module varies from operating system on which node application is running. like when node application on windows operating system it assume that windows based path are used. 
- path.dirname() is similar to `` dirname ``  in UNIX-based operating system.

### DIFF
--- a/snippets/EqualSubsetSum.md
+++ b/snippets/EqualSubsetSum.md
@@ -1,0 +1,36 @@
+---
+title: Partition Equal Subset Sum
+tags: algorithm,array,medium
+firstSeen: 2021-08-31T11:07:26+02:00
+lastUpdated: 2021-08-31T11:07:26+02:00
+---
+
+Given a non-empty array nums containing only positive integers, find if the array can be partitioned into two subsets such that the sum of elements in both subsets is equal.
+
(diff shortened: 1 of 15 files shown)


Chalarangelo (commented):
Thanks for the effort, but these don't add any value to the project whatsoever and break established conventions. Closing alongside the issue.

PR #1866, opened 2021-09-05
Added JStoJSON.md

Clerijr:
Added a snippet to convert a JS script into a JSON file


Chalarangelo (commented):

PR #1876, opened 2021-09-15
Used const instead of let on some files

OussamaAlouat:
Used const instead of let on vectorAngle.md file


OussamaAlouat:
Sorry, but why are invalid?

PR #1878, opened 2021-09-17
Added haversineDistance.md

vikrant4u2:
Calculate the great-circle distance between two points using the Haversine formula. It is the shortest distance between two points on the surface of a sphere, measured along the surface of the sphere.


Trinityyi (commented):

Annamisa:
T7

PR #1886, opened 2021-09-27
Add arrayNegativeIndex

fltenwall:
Creates an array that supports negative indexes.


Chalarangelo (commented):
Thanks for contributing. However, your snippet doesn't provide enough value for the average reader, as `Array.prototype.slice()` is already a valid alternative. Closing due to that.

PR #1888, opened 2021-09-28
Add oncontextmenu 

Min276:
Added oncontextmenu event, a snippet which enables or disables contextmenu to appear  when press right click.


Trinityyi (commented):

PR #1893, opened 2021-10-02
contribute todoList creation

Tharunkumar001:
Contribute a todoList creation in js.


Chalarangelo (commented):
Thanks for contributing. However, your PR doesn't follow our [contribution guidelines](../blob/master/CONTRIBUTING.md) correctly and it doesn't provide any value for the average reader. Closing due to that.

PR #1936, opened 2023-05-23
Added is-perm

MatthewAndreTaylor:
Added is-perm, a quick efficient way to check that two iterables are permutations of each other.


Chalarangelo (approved):
Thanks for the contribution. Seems to work pretty well as-is, but I'll make a couple of minor changes mainly in the metadata and merge. 👍

PR #1940, opened 2023-05-24
added a snippet for prime factors in python

hamidayubdev:


Chalarangelo (approved):
Thanks for contributing. I need to make some metadata and formatting changes, but other than that, good to go.

PR #1948, opened 2023-06-04
I created a new snippet named binary

CodingHusk3y:
It seems to be a standard binary search algorithm but it's not. I created another easy-to-understand version using loop and conditional statements.


Chalarangelo:
Thanks for contributing. However, this snippet doesn't match guidelines and isn't particularly useful or is very simple (documentation), so it's not a good candidate to be added to the collection at this time.

PR #1955, opened 2023-06-30
custom-toast-component-react-snippet added

VivekYadav105:


VivekYadav105:
Added a react snippet named custom toast component.md 
added image blue notifiication.jpg for cover

Chalarangelo:
Thanks for contributing. However, this snippet has numerous formatting issues and is quite complicated in its current state. Perhaps we can revisit it later as a full-fledged article with more context.

PR #1991, opened 2023-08-22
Added a HSLToHexadecimal code snippet

stephaniestelzer:


Chalarangelo:
Thanks for contributing. However, this can be easily produced by combining [HSL to RGB](https://www.30secondsofcode.org/js/s/hsl-to-rgb/) and [RGB to hex](https://www.30secondsofcode.org/js/s/rgb-to-hex/), so there's no need to add an additional snippet.

PR #1995, opened 2023-08-28
[BUG] Fix Cookie Parsing Bug for Complex Values

### LINKED ISSUE #1992: cookie parsing containing a `=` in value
https://github.com/30-seconds/30-seconds-of-code/blob/master/snippets/js/s/parse-cookie.md

if a cookie value contains an `=` would this break

Grenish:
## Description

This pull request fixes a bug in the `parseCookie` function that prevented it from correctly handling complex cookie values containing semicolons and equal signs.

## Changes Made

- Updated the `parseCookie` function to use a more robust approach for splitting and parsing cookies.
- Used a regular expression to split cookies based on semicolons followed by equal signs.
- Adjusted the logic to correctly handle complex cookie values containing special characters.

## Example

Before this fix:
```javascript
parseCookie('foo=bar; equation=E%3Dmc%5E2');
// Output: { foo: 'bar', equation: 'E=mc^2' }
```

After this fix:
```javascript
parseCookie('foo=bar; equation=E%3Dmc%5E2');
// Output: { foo: 'bar', equation: 'E=mc^2' }
```

## Related Issue
Fixes #1992

### DIFF
--- a/snippets/js/s/parse-cookie.md
+++ b/snippets/js/s/parse-cookie.md
@@ -16,8 +16,12 @@ Parses an HTTP Cookie header string, returning an object of all cookie name-valu
 ```js
 const parseCookie = str =>
   str
-    .split(';')
-    .map(v => v.split('='))
+    .split(/;\s?(?=[^=]*=)/)
+    .map((v) => {
+      const parts = v.split('=');
+      const name = parts.shift().trim();
+      const value = parts.join('=').trim();
+    })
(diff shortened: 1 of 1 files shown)


Chalarangelo:
Thanks for putting in the effort, but cookies are meant to be stored without special characters. It's a bad practice to store unencoded equal signs in cookies and it shouldn't be endorsed or shown as an example for beginners. I've explained further under #1992.

PR #2000, opened 2023-09-05
Added CSS snippet for inline-block elements

Noelzak03:


Chalarangelo:
Thanks for contributing. This seems like a pretty beginner, low complexity piece which is better suited to documentation websites like MDN.

PR #2001, opened 2023-09-05
Added a python snippet to reverse vowels in a string

khuzaima-a:


Chalarangelo:
Thanks for contributing, but this snippet doesn't solve a real-world problem.

PR #2003, opened 2023-09-12
Add target-sum-in-list

ritin0204:


Chalarangelo (changes requested):
Thanks for contributing. However, there are some formatting issues (no examples provided, code blocks starts with an empty line, 4 spaces instead of 2) and the explanation of the code could use some work to be more beginner-friendly.

PR #2004, opened 2023-09-12
Added a python snippet to check if a number is an Armstrong number

Pranav-AJ:


snippets/python/s/is-armstrong.md
```diff
@@ -0,0 +1,38 @@
...
+- Initialize Sum=0. 
+- Find the total number of digits in the given number using len(str(copy)). 
+- Iterate through each digit using modulo(%) operator, raise it to the power of order and add the result to Sum.
+- Divide the number by 10 at each iteration and exit the loop when num becomes 0.
+- Return True if Sum is equal to copy of num, False if it is not equal to copy of num.
+
+```py
+def is_armstrong(num):
+    copy = num 
+    Sum = 0
```
ArifShariar, line 25: Not mandatory, but I think it is better if you follow the same naming convention all over the Python snippet. All the other variables in the snippet start with smaller case letters, without "Sum". So, it will be better if you rename this variable "sum".

Pranav-AJ: Hey, thanks for your feedback. But I used 'Sum' because 'sum' is an inbuilt function in python so it would be a bad practice even though it supports it. Maybe I should consider changing the variable name to total/result.

Chalarangelo:
Thanks for contributing, but Armstrong number check has been brought up in the past and isn't really relevant to a lot of users, while also being trivial to compute.

PR #2005, opened 2023-09-14
new python snippet: adding to middle of stack

NithinJakrebet:


Chalarangelo:
Thanks for contributing, but this snippet isn't particularly useful and doesn't really showcase any interesting or best practices.

NithinJakrebet:
Ah ok. My apologies.

PR #2007, opened 2023-09-21
Created python snippets that allow for conversion from Inches to Centimeters and Vice-Versa

AnirudhRevanur:


Chalarangelo:
Thank you for contributing. Unit conversion snippets are generally considered low value, thus we're not currently adding new ones. If we do in the future, I'll reopen this PR. Cheers! 🍻

PR #2018, opened 2023-09-25
Added js snippet to generate unique id

Tusharknwl:
Added Unique id generator javascript snippet.

### DIFF
--- a/content/snippets/js/s/generate-unique-id.md
+++ b/content/snippets/js/s/generate-unique-id.md
@@ -1,0 +1,25 @@
+---
+title: Generate Unique ID in JavaScript
+type: snippet
+language: javascript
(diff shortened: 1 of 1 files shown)


mohanadft:
Hi @Tusharknwl 
Please make sure that `uniqueId` is a function, and I recommend to change the name to `generateId`.

Here is what I mean:
```ts
const generateId = () => Date.now().toString(36) + Math.random().toString(36).substr(2);
```

Chalarangelo:
Thanks but unique id (UUID) generation is already covered in [this](https://www.30secondsofcode.org/js/s/uuid-generator-browser/) and [this](https://www.30secondsofcode.org/js/s/uuid-generator-node/) snippet.

PR #2045, opened 2023-10-13
Added method that returns value from nested and untested object

Shubhamjatara:
Method that returns value from nested and untested object by using recursion technique.


Shubhamjatara (commented):
Tested working fine

PR #2121, opened 2025-07-22
Add timestampDifference utility snippet

tanvisandbhor:
This PR adds a new JavaScript snippet named `timestampDifference` that returns the difference in seconds between two timestamps. It uses `Math.abs()` to ensure a positive value regardless of the order of inputs.

Example:
timestampDifference(1650000000000, 1650000003000); // 3


Chalarangelo:
Thanks but I'm not accepting content contributions at this time.

PR #2122, opened 2025-07-23
Add isEven snippet to check if a number is a valid even integer

Varnika060306:
## What does this snippet do?

Checks if a given input is both:
- An integer
- Divisible by 2 (i.e., even)

This is useful to avoid edge cases with floats (like 4.0) or strings ("6").

---

## ✅ Preview

```js
const isEven = num => Number.isInteger(num) && num % 2 === 0;

isEven(4);      // true (valid even integer)
isEven(7);      // false (odd integer)
isEven(2.5);    // false (not an integer)
isEven("6");    // false (string, not number)
```

---

## 📌 Why is this useful?

- Safer and more robust check for even numbers.
- Works well with user inputs, form validation, APIs, etc.
- Avoids bugs from non-integer inputs like strings or floats.

---

## 📅 Date Modified
2025-07-23

### DIFF
--- a/content/snippets/js/s/is-even.md
+++ b/content/snippets/js/s/is-even.md
@@ -1,20 +1,44 @@
 ---
-title: isEven
+title: Check if a number is even
 shortTitle: isEven
 language: javascript
 tags: [math,number,basic]
 cover: image
-excerpt: Checks if a given number is even using the modulus operator.
+excerpt: Checks if a given number is an even integer using the modulus operator.
 listed: true
 dateModified: 2025-07-23
 ---
 
 ### isEven
 
-Checks if a given number is even.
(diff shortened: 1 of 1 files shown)


Varnika060306:
Hi maintainers 👋 I’ve submitted this snippet aligned with your guidelines. Happy to make any requested changes. Thanks for taking the time to review it!

Chalarangelo:
Please read **THE ENTIRETY** of the guidelines, i.e.:

- No duplicates - [this clearly exists already](https://www.30secondsofcode.org/js/s/number-is-even-odd/)
- [No new content contributions are accepted at this time](https://github.com/Chalarangelo/30-seconds-of-code#:~:text=New%20content%20contributions%20are%20not%20accepted%20at%20this%20time.)

